Issue #23, opened 2022-02-03
Deal with dependabot alerts


cmroberts:
Maintenance makes the world go 'round

starscouter1973:
hey guy,, allow me some question please im a newbie in every case . just scripting a bit depends on passion.   Deal with dependabot alert  is the thing that mekes me nervous. 
just looking for a point i can go to . 
im a little afraid that this warning means more then i hope..

can you give me an advice to which point i have to send my issues--

greets from cologne

cmroberts:
Relevant pieces captured in #45

PR #32, opened 2022-02-08
Upgrading node-sass, chromedriver, and eslint

cmroberts:
Upgrading node-sass, chromedriver, and eslint. Now compatible with node 16! 

Part of #23


Rapol:
In my windows computer I couldn't get the UI running. I get the following error:

> Syntax Error: Error: Node Sass version 7.0.1 is incompatible with ^4.0.0.

I tried deleting node_modules and same issue. Something is caching the node-sass package and build. Ill try tomorrow on my work mac.

cmroberts:
I see the `node-sass@^4.12.0` version referenced in `packages/client/yarn.lock`, but think that file is not used: only yarn.lock files at the project root are used when building. I'm adding removing the yarn.lock file in the client package to this PR, but don't think that'll fix the build issue you're seeing. 

Longer term I think we'll want to do a more thorough upgrade of dependencies here, including vue, but for now I was hoping to do a more targeted set of upgrades so that this project can run on the same version of node as the arpa-reporter project. Just adding that for color / motivation here; thanks for taking a look!

cmroberts:
Bah, even though the app runs/ builds for me with these changes there is an error with node-sass at runtime: 
```
client:  ERROR  Failed to compile with 1 error12:11:14 PM
client:  error  in ./src/App.vue?vue&type=style&index=0&lang=scss&
client: Syntax Error: Error: Node Sass version 7.0.1 is incompatible with ^4.0.0.
```

Closing this out, there's some work to do here (perhaps upgrade vue and retire node-sass and lib-sass as they've been deprecated).

Issue #6, opened 2022-01-12
Add tenant data model


cmroberts:
Add a tenants object to the database, and allow for simple editing of the tenant name field in the UI. 

Also probably want to add a foreign key to the agency table

PR #36, opened 2022-02-16
Adding tenant model and ability to edit display name

cmroberts:
This PR adds the tenant model to the database and the ability to edit the tenant's display name in the app.

Next steps: 
* Using tenant id instead of root agency in access helpers and as the organization id in the client 
* Restricting list of agencies to agencies in the user's tenant

Addresses #6


smattingly (approved):
I'm a bit rusty on USDR grants, and my vue(x) knowledge is limited. I hope Raf or others can review too but this gets my approval.

PR #39, opened 2022-02-21
Update Node and dependencies to work on Apple M1

Mr0grog:
I tried to set up this project (in order to add a simple CI workflow) and I couldn’t since the version of Node and some dependencies wasn’t compatible with my machine. These are the minimum updates to fix the issue (they’re all at least a major version behind what’s current).

I also cleaned up a few things related to basic setup:
- Removed the `package-lock.json` files since this project uses Yarn.
- Added a `.node-version` file for nodenv users, which is just a symlink to the `.nvmrc` file. **There’s a risk this could cause issues on Render, or not work correctly for Windows (but WSL should be fine) users with nodenv.**
- Updated the docs about Node versions and removed an instruction about the `.env.example` file in the `client` package, which does not actually exist.

I did not actually wind up setting up a CI workflow here, since I still can’t get any of the tests to work. That doesn’t appear to be a problem with dependencies or Node, though — just broken tests or config.


Mr0grog:
Also worth noting here: I had to update `chromedriver`, but it also looks like it’s not actually used in the first place, so we could also potentially delete it. I’m guessing this was originally meant to the end-to-end tests, but there aren’t actualy any end-to-end tests (there aren’t *any* tests in the `client` package, for that matter).

README.md
```diff
@@ -43,8 +45,6 @@ psql -h localhost -p 5432
 
 Create .env file in server workspace based on the .env.example. See Deployment section for more information on the .env file.
 
-Also create .env file in client workspace based on the .env.example.
```
cmroberts, line 46: Actually this client .env file does exist and is required for the app to run! The example file is named confusingly (https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/client/.env.development), I'll adjust that in a follow-on PR or feel free to rename it to env.example as part of this changeset.

.nvmrc
```diff
@@ -1 +1 @@
-12.11.1
\ No newline at end of file
+14.19.0
```
cmroberts, line 1: Could we make this v15.14.0 (the most recent version that is compatible with Vue.js version 2)?

cmroberts:
You are a hero for going through and updating a few of these archaic packages. 🙇‍♀️  Thank you for inching us to working with 2022 compatible artifacts. I think the next meaningful step along these lines will be tackling #45 so we can upgrade Vue + Node and then everything else. 

I have been running this app on an M1 for the past month or so using node 15 but the chromedriver workarounds were painful and I'm not going to bother walking you through them as they are unnecessary after #44 is merged. 

I'm going to go ahead and merge the chromedriver removal now. If you don't mind doing three small changes I think this should be ready to merge:
* re-instating the client .env file readme direction
* fixing chromedriver related conflicts after I merge #44
* moving to a slightly newer version of node (reasoning: let's be as modern as possible given our constraints)

README.md
Mr0grog: So FWIW I don't think the naming was confusing at all, but may require more familiarity with dotenv than some folks have (?). Dotenv will pick up any `.env` or `.env.$NODE_ENV` file, so it will pick up the existing `.env.development` file just fine without you needing to create a new `.env` (at least it did so quite reliably for me). Since you don't actually need to do anything here, I thought removing the instruction (which inaccurately references `.env.example` to boot) made the most sense.

.nvmrc
Mr0grog: NO PLEASE DO NOT! Odd-numbered major releases of Node should *never* be used in production and are ONLY meant for testing upcoming features in Node and ABI changes (for packages that use ABI, and this does not).

Mr0grog: (If you're curious, this versioning scheme was based on Debian Linux, which does the same thing. In my experience, a LOT of people make this same mistake as a result.)

README.md
Mr0grog: (If curious, this is a design choice inherited from the original Ruby dotenv gem, described here: https://github.com/bkeepers/dotenv#can-i-use-dotenv-in-production and you can see how NPM dotenv’s docs, which are newer, discourage this usage, described here: https://github.com/motdotla/dotenv#should-i-commit-my-env-file)

.nvmrc
cmroberts: Makes sense, I'll downgrade locally as well!

README.md
cmroberts: @Mr0grog just tested building out the client without a .env file in my local build and the app doesn't load even when in development mode. It works when a .env file is present. 

I'm obviously not deeply familiar with dotenv and the mechanics here but wanted to flag that since I needed to add this file manually when doing my initial set up; perhaps there's a way to make this seamless so the next person to walk through this README can build successfully

Mr0grog: > Does it actually work for you locally? …just tested building out the client without a .env file in my local build and the app doesn't load even when in development mode.

Weird! It worked great for me. I never created a `.env` file.

Mr0grog: Maybe the best solution here is to rename the file to `.env.example` and retain the instruction. (As a side note, I generally don’t think it’s good to commit `.env` files with values that are intended to be used, so I think this would be better overall anyway!)

Mr0grog:
Re: Node.js versions, there’s probably no reason not to upgrade to 16.x if 14.x is problematically old! I was going for *minimum* update in this PR, which is the only reason I chose 14 instead of 16 here. 16 is probably better over the long run anyway since it is the current LTS release.

cmroberts:
@Mr0grog I recently tried to upgrade the project to node 16 but found a dependency related to Vue.js version 2 that was gated on node 15 or lower. Feel free to give it a try, but I know you've sunk a lot of time into this already, so I think it's fine if that waits for #45. 

There is *SO MUCH* to upgrade in this repo and getting on a more modern version of Vue should unblock all of it once the app has been updated to be compatible with the new version.

README.md
cmroberts: 🎉 That seems great. I wonder what is happening here for me locally 😬

Mr0grog:
> I recently tried to upgrade the project to node 16 but found a dependency related to Vue.js version 2 that was gated on node 15 or lower.

Ahhhh, I think you mentioned this before and I forgot. So it seems like we should just leave this part of the PR as-is. 👍

cmroberts:
Yeah I think it's ready to go :+1:

Mr0grog:
OK, so what’s left here is:
- [ ] Rename `.env.development` → `.env.example` and reinstate the README instruction about it.
- [ ] Rebase and clean up merge conflicts from #44.

I won’t get back to this until next week, so feel free to finish these out yourself if you’d like to get it done sooner!

PR #40, opened 2022-02-21
Fix some tests and add a workflow configuration to run them on PRs

Mr0grog:
This is a first cut at fixing some of the tests, since none of them ran successfully when I cloned this repo. It’s also a follow-on to #39 and needs the commits from it to work. **This is also incomplete; I’m not sure I’ll have time to keep working on it, but hopefully it’s a good start for someone else.**

What I did so far:
- Fixed the API tests in the server so that they run.
    - Changed a few things in the scripts so that errors are handled correctly instead of swallowed.
    - Added some error-checking/logging to the test utilities so I could see what was going wrong.
    - Ensure the database is disconnected correctly so tests don’t hang.
    - Use the correct users. The tests rely on a lot of users listed in the seeds, and several of those users were changed or removed in b465e776be3b2d0d884c3fed20bd630fdaa3610e. I’ve added them back, but hopefully with more obvious names that mark them as needed for tests. **It doesn’t look like the seeds are meant to be used in production, but if they are being used in production, this change is not OK, and should not be merged as-is!**
- Fixed a lint warning that I think was also causing some test issues.
- Added a workflow file to run tests and linters on GitHub actions.

Still to do:
- [ ] Fix DB tests so they can run in a non-interactive environment (it always displays a password prompt right now, but otherwise the tests run successfully).
- [ ] Fix DB tests to use the same database as the API tests. (There is a different database name for two groups of tests; one is hardcoded and does not obey the `.env` file; I’m not sure what the history or reasoning was here, but it seems bad at first glance. Please LMK if this is for a good reason and shouldn’t change!)
- [ ] Add credentials and configuration so the e-mail tests can run on GitHub Actions. I’m not sure I’m the right person to do this and don’t have whatever credentials are *supposed* to be used right now. Who does? *(These tests pass if you have good credentials! I made my own accounts and verified everything works.)*

Other bits worth noting:
- The `client` package has two test scripts (`test:unit` and `test:e2e`), but neither has actual tests behind them, so they just fail if you try and run them.
- The server DB tests have their own set of seeds, which is a little complicated.
- I added test users to the server seeds. I think this is OK since they don’t look like they’re meant to be used in production, but if we are using them in production, that’s not good.
- ~Where possible, I like to use https://github.com/wearerequired/lint-action for linting, since it surfaces lint warnings and errors as check objects on the actual line they occur on in the files tab on PRs (it’s really nice!). That doesn’t work with all the extra magic `vue-cli-service lint` adds in the client package, though, so I didn’t use it here. Lint errors will still fail the build, but warnings won’t, so they’ll be easier to miss.~ Make sure you click the “details” link for the lint job, since there are already several warnings! https://github.com/usdigitalresponse/usdr-gost/runs/5281030868?check_suite_focus=true *(Update: it looks like GitHub added some magic to make this work automatically now, so the special action isn’t needed.)*


.github/workflows/ci.yml
```diff
@@ -0,0 +1,167 @@
...
+        run: |
+          cd packages/client
+          yarn lint --no-fix
+      
+      - name: Lint Server Source
+        run: |
+          cd packages/server
+          yarn lint
+
+  test-server:
```
Mr0grog, line 44: This job is currently failing, which is expected. See the PR description for details about what needs to be done here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,167 @@
...
+          # credentials to be invalid.
+          AWS_ACCESS_KEY_ID: 'Fake AWS Key'
+          AWS_SECRET_ACCESS_KEY: 'Fake AWS Secret'
+        run: |
+          cd packages/server
+          # The .env file needs to be present; the example file is good enough.
+          cp .env.example .env
+          yarn run test:api
+
+  test-server-db:
```
Mr0grog, line 109: This job is currently failing, which is expected. See the PR description for details about what needs to be done here.

packages/server/__tests__/api/users.test.js
```diff
@@ -99,13 +99,13 @@ describe('`/api/users` endpoint', async () => {
                 const response = await fetchApi(`/users`, agencies.own, fetchOptions.admin);
                 expect(response.statusText).to.equal('OK');
                 const json = await response.json();
-                expect(json.length).to.equal(11);
+                expect(json).to.have.lengthOf(7);
```
Mr0grog, line 102: I changed the `expect()` style here to give a more verbose and helpful message while I was at it. There’s a *lot* of room to improve the expectations with things like this all over these test suites. The `chai-as-promised` plugin can also help clean things up and provide better messaging around async code as well.

packages/server/package.json
```diff
@@ -12,7 +12,7 @@
     "db:seed": "knex seed:run",
     "test": "mocha __tests__/*.test.js",
     "test:db": "__tests__/db/run-tests.sh",
-    "test:api": "NODE_ENV=test SUPPRESS_EMAIL=TRUE node src > test.log & NODE_ENV=test mocha --require __tests__/api/fixtures.js __tests__/api/*.test.js && pkill -fn 'node src'",
+    "test:api": "NODE_ENV=test SUPPRESS_EMAIL=TRUE node src > test.log & NODE_ENV=test mocha --require __tests__/api/fixtures.js __tests__/api/*.test.js; exit_code=$?; pkill -fn 'node src'; exit $exit_code",
```
Mr0grog, line 15: This is pretty messy! A better pattern might be to use before/after each/all hooks in the tests to start and stop the dev server on a random port.

We have some nice tooling in Univaf for this (although ours is designed for Jest instead of Mocha, so it might require some slight tweaks to work here):

https://github.com/usdigitalresponse/univaf/blob/2fca59d558de365d84057bf09892d5e73e891e71/server/test/lib.ts#L18-L47

Use it like so:

https://github.com/usdigitalresponse/univaf/blob/2fca59d558de365d84057bf09892d5e73e891e71/server/test/api.edge.test.ts#L20-L36

Mr0grog (commented):
`test-server` and `test-server-db` are failing, but that’s expected. We can either comment them out, or do the work to fix them before merging.

Mr0grog:
> Where possible, I like to use https://github.com/wearerequired/lint-action …that doesn’t work with all the extra magic vue-cli-service lint adds in the client package, though, so I didn’t use it here.

It looks like GitHub has added some extra magic to automatically detect these kind of warnings and add them, so this actually *just works* now! Nevermind this bullet point. :)

packages/server/__tests__/api/utils.js
```diff
@@ -9,11 +9,15 @@ const knex = require('knex')({
...
-    await fetch(`${process.env.API_DOMAIN}/api/sessions`, {
+    const newSession = await fetch(`${process.env.API_DOMAIN}/api/sessions`, {
         method: 'POST',
         body: JSON.stringify({ email }),
         headers: { 'Content-Type': 'application/json' },
     });
+    const body = await newSession.json();
+    if (!newSession.ok || !body.success) {
+        throw new Error(body.message || `Error creating session: ${JSON.stringify(body)}`);
+    }
```
Mr0grog, line 20: Total side note, but it seems like this should have gotten a 4xx response (in practice, it was a 200, which is why I added the `|| !body.success` bit).

Rapol:
@mattbroussard can we close this one? We should be able to salvage the github ci workflow and readme changes.

Mr0grog:
This was originally meant as a bit of quick help for @cmroberts before she moved on. If it’s pretty hard to integrate at this point, definitely feel free to salvage what’s useful and close it!

as1729:
Closing for now. Will pull out parts of it in separate PRs as we continue to improve the test coverage.

Issue #43, opened 2022-03-09
[ID] Nevada's SPOC button appears even for non-Nevada tenants


mhuang-usdr:
From OH:

> Once the user indicates if they are interested or not, a button that says SPOC appears. If you click it, it is the form for the State of Nevada. Ohio does not follow the SPOC process nor has a form.

Raphomet:
Is the SPOC process rare or common? I'm wondering if this should be controlled a user setting, or if USDR should just set it on an account by account basis.

(Maybe the latter for MVP and Settings Page later?)

mhuang-usdr:
USDR should set it on a per-tenant basis. SPOC is a state program, ~11 states have it, each state has a different form for it. This feature doesn't need to be user configurable.

davidbkay:
@mhuang-usdr merged to staging

mhuang-usdr:
I'm unable to log into the staging environment to check

mhuang-usdr:
Just reviewed in the staging environment. Looks good to me.

as1729:
Closing this issue as it is complete 🥳

as1729:
Noticing that this PR was not merged. Will revisit this.

mhuang-usdr:
This button still shows up on the USDR tenant, though downloading it is broken

PR #58, opened 2022-04-19
43 add spoc button to tenants

davidbkay:
addresses #43 by adding `uses_spoc_process` column through migration and checks for the tenant `tenant.uses_spoc_process` to show/hide SPOC button in grants modal.

### DIFF
--- a/packages/client/src/components/Modals/GrantDetails.vue
+++ b/packages/client/src/components/Modals/GrantDetails.vue
@@ -56,7 +56,7 @@
               <b-button variant="outline-success" @click="markGrantAsInterested">Submit</b-button>
             </b-col>
           </b-row>
-          <b-row v-if="interested && !interested.interested_is_rejection">
+          <b-row v-if="interested && !interested.interested_is_rejection && shouldShowSpocButton">
             <b-col>
               <b-button variant="primary" @click="generateSpoc">Generate SPOC</b-button>
             </b-col>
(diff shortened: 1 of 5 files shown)


packages/server/src/db/index.js
```diff
@@ -440,11 +455,26 @@ function getInterestedCodes() {
...
+        'danger_threshold',
+    )
+        .from(TABLES.agencies)
+        .where({
+            id: agencyId,
+        })
+        .leftJoin('tenants', 'tenants.main_agency_id', '=', `${TABLES.agencies}.main_agency_id`);
+    // const query = `SELECT id, name, abbreviation, parent, warning_threshold, danger_threshold
+    // FROM agencies WHERE id = ?;`;
+    // const result = await knex.raw(query, agencyId);
```
Rapol, line 473: can we remove the comment?

packages/server/src/db/index.js
```diff
@@ -121,6 +132,10 @@ async function getUser(id) {
         }
         user.agency.subagencies = subagencies;
     }
+    delete user.tenant_id;
+    delete user.tenant_display_name;
+    delete user.tenant_main_agency_id;
+    delete user.tenant_uses_spoc_process;
```
Rapol, line 138: nice catch, would be nice to keep this in mind to delete other temp fields in the future, for example agency_name. Maybe we start using DTOs to hide and transform fields.

packages/client/src/components/Modals/GrantDetails.vue
```diff
@@ -166,18 +166,27 @@ export default {
...
       if (!this.selectedGrant) {
         return false;
       }
-      return this.selectedGrant.viewed_by_agencies.find((viewed) => viewed.agency_id.toString() === this.selectedAgencyId);
+      return this.selectedGrant.viewed_by_agencies.find(
+        (viewed) => viewed.agency_id.toString() === this.selectedAgencyId,
+      );
+    },
+    shouldShowSpocButton() {
+      if (!this.user.tenant.uses_spoc_process) return false;
```
Rapol, line 180: might be useful to create a tenant selector

Rapol (approved):
Looks good! Just some minor comments but nothing that blocks it.

packages/client/src/components/Modals/GrantDetails.vue
davidbkay: @Rapol can you elaborate? Do you mean having interested tenants as well as agencies?

mhuang-usdr: Not sure I understand the full context here, but users shouldn't be tracking interest in grants on the tenants level.

Rapol: @davidbkay Sorry David that was super vague. My fault.

Im thinking that this tenant property in the user will be reused in other places. If so it might be better to create a new selector to retrieve the tenant.

We would define a new selector in the user vuex module called `getCurrentTenant()`. We can then import that as a computed field

```
computed: {
    ...mapGetters({
      currentTenant: 'users/getCurrentTenant',
    }),
```

Usage of it in the component

```
shouldShowSpocButton() {
      if (!this.currentTenant.uses_spoc_process) return false;
}
```

davidbkay: @Rapol yeah I'm totally down with that. I'll make associated updates.

Rapol: @mhuang-usdr This change is for hiding the generate spoc pdf button that was request by nevada. This pdf is for filling out the required document to start tenant grant application process in their internal workflow.

PR #62, opened 2022-04-21
Remove COVID-related seeded keywords (closes #15)

### LINKED ISSUE #15: [ID] Remove Covid-related default keywords from Keywords
Under https://gost-grants-tools.onrender.com/#/keywords, currently we have a bunch of pre-seeded keywords. Partners have given feedback that this is no longer as helpful since CARES is wrapping up, and they tend to delete it and replace it with other keywords instead.

mattbroussard:
Closes #15 by doing pretty much what the title says.

I had to make some changes to the keywords API test because it assumed the presence of the pre-seeded keywords. Now, the test creates its own keywords.

I also noticed that the test referred to a user that doesn't seem to exist in the users seed, `michael@nv.gov`. I replaced this with a new admin user under the same agency and added a better error log for when `getSessionCookie` is called with an invalid user. I noticed several other tests were failing locally for a similar reason, I've fixed those too, in a separate stacked PR: #63

### DIFF
--- a/packages/server/__tests__/api/keywords.test.js
+++ b/packages/server/__tests__/api/keywords.test.js
@@ -36,7 +36,7 @@ describe('`/api/keywords` endpoint', async () => {
 
     before(async function beforeHook() {
         this.timeout(9000); // Getting session cookies can exceed default timeout.
-        fetchOptions.admin.headers.cookie = await getSessionCookie('michael@nv.gov');
+        fetchOptions.admin.headers.cookie = await getSessionCookie('admin1@nv.gov');
         fetchOptions.staff.headers.cookie = await getSessionCookie('user1@nv.gov');
     });
 
--- a/packages/server/__tests__/api/keywords.test.js
+++ b/packages/server/__tests__/api/keywords.test.js
@@ -3,6 +3,7 @@ require('dotenv').config();
 
 const _ = require('lodash-checkit');
 const { getSessionCookie, fetchApi, knex } = require('./utils');
+const { TABLES } = require('../../src/db/constants');
 
 describe('`/api/keywords` endpoint', async () => {
     const agencies = {
@@ -52,7 +53,7 @@ describe('`/api/keywords` endpoint', async () => {
             notes: '',
             agency_id,
         }));
-        const createdKeywords = await knex('keywords')
+        const createdKeywords = await knex(TABLES.keywords)
             .insert(testKeywords)
             .onConflict('id')
             .merge()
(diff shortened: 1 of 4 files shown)


packages/server/__tests__/api/keywords.test.js
```diff
@@ -34,10 +33,31 @@ describe('`/api/keywords` endpoint', async () => {
...
+        // Note: this does not dedupe agency IDs and that's by design; some of the tests need to
+        // delete multiple keywords from the same agency.
+        const allAgencyIds = Object.values(agencies.admin).concat(Object.values(agencies.staff));
+        const testKeywords = allAgencyIds.map((agency_id) => ({
+            mode: 'autoinsert ALL keywords matches',
+            search_term: 'test_keyword',
+            notes: '',
+            agency_id,
+        }));
+        const createdKeywords = await knex('keywords')
```
mattbroussard, line 55: note to self: perhaps this should reference `TABLES.keywords`

packages/server/__tests__/api/keywords.test.js
```diff
@@ -34,10 +33,31 @@ describe('`/api/keywords` endpoint', async () => {
...
+            .insert(testKeywords)
```
mattbroussard, line 56: Hmm, might need to clean this up after the test since other tests call e.g. `getGrants` which queries filtered based on the keywords table.

packages/server/__tests__/api/keywords.test.js
```diff
@@ -34,10 +34,43 @@ describe('`/api/keywords` endpoint', async () => {
...
+        const createdKeywords = await knex(TABLES.keywords)
+            .insert(testKeywords)
+            .onConflict('id')
+            .merge()
+            .returning(['id', 'agency_id']);
+        testKeywordsByAgency = _.groupBy(createdKeywords, 'agency_id');
+    });
+
+    after(async () => {
+        // Delete keywords created by this test to avoid impacting other tests
```
Rapol, line 65: hmm in the future we might want to look at the seeding and maybe clear and seed the db before every test

Rapol (approved):

packages/server/__tests__/api/keywords.test.js
mattbroussard: I like that conceptually, I'm not sure how much it would slow down running the tests overall. If it does have an impact, it could be annoying for local development since right now there doesn't seem to be a good way to run a single test/suite, we just run everything every time. I'm not going to try to do this now but I agree it's worth looking at in the future.

PR #63, opened 2022-04-21
Fix other tests affected by invalid seed user

mattbroussard:
Depends on #62

As mentioned in #62's description, in the course of working on the PR I discovered some other tests failing locally because they referenced users that do not exist in the users seed. This fixes that for the other affected tests that were not directly related to what I was doing in #62.

I also found that two tests depended on a specific number of users that did not match what the seed creates and fixed that here too.

More broadly, I'm wondering if I'm misunderstanding the use case of these DB seeds -- I assumed they were intended just for local and test environments, but the existence of #15 implies they are used for creating agencies, keywords, etc. in production as well? It also seems that for the issues above to arise, the users seed would have needed to be modified after the tests were written (indeed #36 seems to be the culprit). Perhaps there should be a separate set of (seldom changed) seeds for API tests specifically?

### DIFF
--- a/packages/server/__tests__/api/agencies.test.js
+++ b/packages/server/__tests__/api/agencies.test.js
@@ -34,7 +34,7 @@ describe('`/api/organizations/:organizationId/agencies` endpoint', async () => {
 
     before(async function beforeHook() {
         this.timeout(9000); // Getting session cookies can exceed default timeout.
-        fetchOptions.admin.headers.cookie = await getSessionCookie('michael@stanford.notreal');
+        fetchOptions.admin.headers.cookie = await getSessionCookie('mindy@usdigitalresponse.org');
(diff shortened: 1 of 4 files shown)


Rapol:
After double checking I think you are right.

There are two integration tests
test:db - have their own fixtures at packages/server/__tests__/db/seeds/fixtures.js
test:api - uses knex.js seeds at packages/server/seeds/dev/index.js

For a short term fix, we should create a new knex.js seed folder named test and copy the current seeds. We then can modify them without breaking our production seeds 😅

Rapol (approved):
I ran both db and api tests and they are passing. Thanks!

mattbroussard:
>For a short term fix, we should create a new knex.js seed folder named test and copy the current seeds. We then can modify them without breaking our production seeds 😅

I'll look into this, but I think it belongs in a separate PR so merging this one as-is (and also to unblock merging my other one #65).

Issue #55, opened 2022-04-07
(2) [ID Tool] Add, remove, and reorder grants table columns


Raphomet:
We recently got a walkthrough from Pari, a former state budget officer, of what she looks for when she's looking at grants. She gave us some new information about which data are most important and which aren't.

On both the My Grants and the All Grants view, let's make some changes to the table columns:

![grants table with new columns](https://user-images.githubusercontent.com/58006/162336581-5bab6055-bf0c-4960-b156-1db413336491.jpeg)

Here are the changes, and why:

1. **Remove the Grant ID column.** We don't need both Grant ID and Grant Number — both are unique identifiers, but grants.gov uses the Grant Number more (although they call it the "Opportunity Number").
2. **Rename Grant Number to Opportunity Number.** This matches what's on grants.gov and is more consistent for the user. Although I didn't wireframe it, this column can also be dark, just like Grant ID is now.
3. **Keep Title, Viewed By, and Interested Agencies the same.**
4. **Remove the Agency Code column.** We don't think it's important enough to show in this table. They can click through if they need to.
5. **Move Status to the right of Interested Agencies.**
6. **Move Opportunity Category to the right of Status.**
7. **Add Award Floor and Award Ceiling to the right of Cost Sharing.** We don't currently have these columns, but they are important to budget officers — they tell them how much they could expect to get if the grant is awarded! **These columns should also be sortable.**
8. **Remove Created At and Updated At.** These are debug columns that reflect the state of the database row, not the state of the grant — it's for developers, not the end user.

Raphomet:
Pari actually says to **get rid of the Posted Date column**, too.

Raphomet:
This applies to both the "My Grants" and "Grants / Browse Grants" tabs.

mhuang-usdr:
Current state:
<img width="557" alt="image" src="https://user-images.githubusercontent.com/91221265/174902894-de42040f-76a4-4fc5-82fc-0a59fee3929f.png">


Desired state:
![image](https://user-images.githubusercontent.com/91221265/174902783-c44a8ca1-60cd-4416-af81-d7bec0782256.png)

mhuang-usdr:
cavukovic is taking this on (waiting to add him to the repo before we can officially assign him to the bug)

PR #65, opened 2022-04-22
CSV export feature for grants views

### LINKED ISSUE #16: [ID] Export information from the ID tool into an excel spreadsheet for maximum compatibility with govt workflows
From Ohio

as a stopgap until new features are in, can we export the info to an Excel Sheet and we can track/modify on our own after?

mattbroussard:
Closes #16

image

Sample export output: [grants.csv](https://github.com/usdigitalresponse/usdr-gost/files/8558345/grants.20.csv)

This adds a CSV export button to the `GrantsTable` component, used in both the Grants and My Grants tabs. The endpoint for generating the CSV shares most logic and parameters with the existing `/api/organizations/:organizationId/grants` endpoint that populates the table UI, so the downloaded CSV should match the rows displayed on the page, including sorting, search query, filters (interested vs. assigned), etc. other than pagination.

The CSV column headers and data format are similar to what is shown on the page with some slight adjustments:

 - rearrange columns as suggested in #55 (note this change is only applied in the CSV; I haven't touched the UI)
 - remove created and updated timestamp columns
 - rename "Grant Number" to "Opportunity Number"
 - add Award Ceiling, Award Floor, and URL columns

There is a limit of 10,000 rows per export (originally I'd arbitrarily chosen 100k as a the limit but while writing API tests discovered that the queries in `getGrants` choke if there are this many rows, since they try to do a `WHERE agency_id IN (?, ?, ...)` in a second query to populate viewed & interested agencies. Lowering to 10k seemed reasonable; if this is a problem we can explore fixing that query to do some batching or something.

This PR adds a dependency on [`csv-stringify`](https://www.npmjs.com/package/csv-stringify), which has MIT license. Do we need to list open source dependencies somewhere?

Note: one of the new tests fails as-is because the query ends up depending on pre-seeded keywords that are not present in the test grants inserted, but I plan to address this by merging #62 first.


mattbroussard:
Moving discussion from Slack, @mhuang-usdr [said](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1650663378905009?thread_ts=1650660629.009109&cid=C0324KDQSCR):

> Thanks for raising this question! From what I understand, this PR would create an export button that exports exactly what the user sees, correct? If so, I think that's a great MVP.
>
> A heads up that this crosses a bit with https://github.com/usdigitalresponse/usdr-gost/issues/55 - essentially we dug into what users actually want to see on the table and made a recommendation to change the table itself. (They are in fact interested in award ceiling, etc.) So as long as this PR will reflect the changes to the table, we are solid.
>
> The one other product consideration I have - if we do get future asks from partners to export info that doesn't make sense to show on the table (ex. the URL), is this implementation extensible?

---

> exactly what the user sees

Yes, except for pagination (CSV will contain up to 100k rows, UI only shows 10 at a time)

>So as long as this PR will reflect the changes to the table, we are solid.

Not automatically but changing the columns in the CSV is a trivial code change.

If we were to make it possible for users to customize columns (i.e. turn them on and off or reorder them) in the UI, that's a little more complicated to propagate to the CSV. But thinking a bit more about it, I don't think we'd really want that anyway. Users can delete or reorder columns easily in Excel after downloading the CSV, so I think the CSV should just include as much information as possible and not be configurable past the search/sort filter.

What other fields do you think would be most valuable to add to the CSV? From poking at the current UI and API response, some ideas:
- Award ceiling
- Grants.gov permalink
- "Cfda list" -- I see this in the popup UI but I don't know what this means or if it's important
- Eligibility codes -- not in table or popup UI, but is in API response and appears to be part of the filtering
- Description -- this can include HTML though, which might be a little unwieldy to deal with in a spreadsheet?
- Award floor -- #55 mentions this but I don't see it in any existing UI or API response; in the DB it seems to be only within `raw_body` scraped from grants.gov?
- Contact emails for the interested agencies? -- not sure if this is desirable in CSV or not

>is this implementation extensible?

Yes, adding additional columns is easy, particularly if the data is already present in what [`getGrants`](https://github.com/usdigitalresponse/usdr-gost/blob/3a7275c0ed5aee74928640d6a2d54846e18371b4/packages/server/src/db/index.js#L263) returns.

mhuang-usdr:
> I think the CSV should just include as much information as possible and not be configurable past the search/sort filter.
Agreed. Also, we have no current plans to allow users to hide columns.

> What other fields do you think would be most valuable to add to the CSV? 
- Award Ceiling 
- Grants.gov permalink
- Ideally Award Floor but if it's too unwieldy we can deprioritize it until someone asks

No need for:
- Eligibility codes
- Description 
- Contact emails for the interested agencies
- Cfda list 


Thanks Matt :)

packages/server/__tests__/api/grants.test.js
```diff
@@ -378,4 +379,64 @@ describe('`/api/grants` endpoint', async () => {
...
+                description: 'fake grant inserted by test',
+                eligibility_codes: '25',
+                opportunity_status: 'posted',
+                raw_body: 'raw body',
+            }));
+
+            // batchInsert is needed for large quantity of rows; it splits the insert into multiple
+            // queries
+            // TODO: is there a better way to test this that doesn't require actually inserting this
+            // many rows? Having a test this slow is not ideal. On my machine, take 2500ms
```
Rapol, line 425: only way i can think of is mocking knexjs db call or mocking the getGrants in db module

packages/server/src/routes/grants.js
```diff
@@ -1,6 +1,9 @@
 const express = require('express');
 
 const router = express.Router({ mergeParams: true });
+// TODO: why is this necessary?
+/* eslint-disable import/no-unresolved */
+const { stringify: csvStringify } = require('csv-stringify/sync');
```
Rapol, line 6: I just tried it and my eslint is also showing the error. Might be because of csv-stringify being a esm module? `node_modules/csv-stringify/dist/esm/sync` No clue

packages/server/src/routes/grants.js
```diff
@@ -29,6 +32,24 @@ async function getAgencyForUser(selectedAgency, user, { filterByMainAgency } = {
     return agencies.map((s) => s.id);
 }
 
+// Award floor field was requested for CSV export but is not stored as a dedicated column,
+// so we have to extract it from raw_body
+function getAwardFloor(grant) {
```
Rapol, line 37: If we wanted to add it, this is where https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/server/src/lib/grantscraper/index.js#L29

packages/client/src/store/modules/grants.js
```diff
@@ -94,6 +94,16 @@ export default {
...
     },
+    exportCSV(context, queryParams) {
+      const query = Object.entries(queryParams)
+        // filter out undefined and nulls since api expects parameters not present as undefined
+        // eslint-disable-next-line no-unused-vars
+        .filter(([key, value]) => value || typeof value === 'number')
+        .map(([key, value]) => `${encodeURIComponent(key)}=${encodeURIComponent(value)}`)
+        .join('&');
+      const navUrl = fetchApi.addOrganizationId(`/api/organizations/:organizationId/grants/exportCSV?${query}`);
+      window.location = navUrl;
```
Rapol, line 105: this will open a new tab and close it or will the user need to click back to come back to the site?

Rapol (approved):
cant live without spreadsheets!

packages/server/__tests__/api/grants.test.js
mattbroussard: Do we have any other API tests that mock this way you can point me at?

As far as I can tell, these API tests make requests against an actual running server in a completely separate process. The server behaves basically identically to normal except that based on `NODE_ENV=test` it connects to a different DB url (and it seems not to send login emails; can't really see where in the code prevents this from happening though since the testutils function hits the same login endpoints a user would).

packages/server/src/routes/grants.js
mattbroussard: We have this eslint-disable comment in a few other places for imports too. ¯\_(ツ)_/¯

packages/server/src/routes/grants.js
mattbroussard: Cool. I prob won't bother with this for now since this is the only consumer and we don't have a need to query against it in the DB, but good to know. If I were to add a new column and fill it there, will existing rows get it backfilled from future grantscraper runs? Or will it only be filled on newly inserted rows?

packages/client/src/store/modules/grants.js
mattbroussard: I'll test in FF and Safari too, but at least in Chrome if you click a link or navigate in the same window like this to a URL that serves `Content-Disposition: attachment`, it just downloads the file without actually taking the user away from the page they're on. I suppose it's possible they'd navigate away to an error page if the export endpoint errors somehow, but I don't think clicking back is a big deal in that scenario.

packages/server/__tests__/api/grants.test.js
Rapol: ah I see, then mocking is not an option. Sorry not catching that. Maybe we can lower the number of rows for this test. Seems that this is more of a stress testing than functional, unless we want to do both. In grants.gov i see around 6k of active grants

packages/server/src/routes/grants.js
Rapol: hmm not sure but I believe so since it will update or insert the grant

packages/server/__tests__/api/grants.test.js
mattbroussard: The functional part of the test is that 1) the endpoint enforces a limit and doesn't try creating a massive export if there are more rows and 2) it correctly emits an error message at the bottom of the sheet if the exported data is incomplete.

But there is also a stress test component in that if someone changes the logic in `getGrants` in the future to do more complex queries, we would want to know that it doesn't break down at 10k the way it currently does at 100k (where I got errors in building a prepared statement with that many parameters).

If the current number is 6k, it seems believable to me that we could hit 10k at some point, though I think most users will probably have filtered that list down with eligibility codes and keywords.

mattbroussard: I decided 2500ms is not really acceptable in the test; decided to lower limit in test environment (via `NODE_ENV`) to 100 rows which makes the test run in under 100ms. Doesn't achieve the "stress test" goal above though.

packages/client/src/store/modules/grants.js
mattbroussard: Tested and confirmed FF and Safari on Mac have similar behavior.

Issue #67, opened 2022-04-26
Create separate knex seeds for dev vs. test environments


mattbroussard:
See comments on https://github.com/usdigitalresponse/usdr-gost/pull/63#issuecomment-1109270646 for more context, but basically we share the same knex DB seeds for local development and API tests right now, which makes it pretty easy to accidentally break tests by changing the seeds.

It would be better to have a separate set of seeds (to start with, can just copy-paste the existing ones) used for the tests, which are changed less often (& also could potentially be better tailored for the tests). @Rapol also suggested that it might be nice to reset DB + re-seed between test suites so that unrelated tests can't affect each other when they e.g. add or delete rows.

Issue #61, opened 2022-04-20
User should be able to assign a grant to any agency, not just agencies who are subagencies 


mhuang-usdr:

davidbkay:
i'll grab this one since it runs parallel with #66. We'll have to track `assigned_by` which is already in the `assigned_grants_agency` table

as1729:
Additional context here:
1. Query for populating the list will need a new endpoint that filters by tenantId and not SubAgency
2. The assignment of the grant to an agency will also need to be updated to allow to assign grants to agencies within the tenant.
```
router.put('/:grantId/assign/agencies', requireUser, async (req, res) => {
    const { user } = req.session;
    const { grantId } = req.params;
    const { agencyIds } = req.body;
    if (!agencyIds.every((agencyId) => isPartOfAgency(user.agency.subagencies, agencyId))) {
        res.sendStatus(403);
        return;
    }

    await db.assignGrantsToAgencies({ grantId, agencyIds, userId: user.id });
    res.json({});
});
```

mhuang-usdr:
@davidbkay our conversation about multitenancy yesterday bled over to this issue. Have you started on this issue yet? If not, do you mind if @as1729 takes over?

mhuang-usdr:
@slants3 I've verified that this bug still exists.

If you create a new sub-agency as a sub-agency, and switch yourself to view from that sub-agency, you are unable to assign the agency to anyone except yourself. You should still be able to assign a grant to any agency in the tenant.

<img width="960" alt="image" src="https://user-images.githubusercontent.com/91221265/181133230-76dfa2d6-b443-41f0-9526-681a9e226f60.png">

bspates:
@mhuang-usdr I'm having trouble understanding the use case from the context in this issue. Is it that a user that is part of an agency should be able to assign a grant to any sub-agency under their agency? Or something more of different?

mhuang-usdr:
@bspates good question!

One of the goals of our tools is to ease communications and break down information silos between different parts of an organization. Hence, any user that happens upon a grant should be able to assign that grant to any other agency in their TENANT.

Right now, the user is only able to assign grants to their own agency or children agencies.

PR #69, opened 2022-04-26
fix: added agency id check to remove agency button in agency table

### LINKED ISSUE #66: [ID Tool] Agency staff should not be able to see unassignment buttons to unassign other agencies
In the grant modal window, agency staff members are currently able to see the trash bin icons that imply that they are able to unassign the grant from other agencies.

In this example, the user is a staff member of the "Administration: Fleet Services Division" agency.

CleanShot 2022-04-26 at 13 20 48@2x

They should not be able to see the bottom 3 trash bin buttons.

(For what it's worth, when they are pressed, nothing happens, which is the correct behavior. I presume there is a backend authorization check.)

davidbkay:
easy fix added a check in the grants modal to only show delete button if the row's agency_id value matched the user's agency.id.

Closes #66

### DIFF
--- a/packages/client/src/components/Modals/GrantDetails.vue
+++ b/packages/client/src/components/Modals/GrantDetails.vue
@@ -92,7 +92,7 @@
         :fields="assignedAgenciesFields"
       >
       <template #cell(actions)="row">
-        <b-button variant="danger" class="mr-1" size="sm" @click="unassignAgenciesToGrant(row)">
+        <b-button v-if="row.item.agency_id === agency.id" variant="danger" class="mr-1" size="sm" @click="unassignAgenciesToGrant(row)">
           <b-icon icon="trash-fill" aria-hidden="true"></b-icon>
         </b-button>
       </template>
--- a/packages/server/seeds/dev/ref/users.js
+++ b/packages/server/seeds/dev/ref/users.js
@@ -25,7 +25,7 @@ module.exports = [
     },
     {
         id: 2,
-        email: 'mindy@usdigitalresponse.org', // fake email for testing
+        email: 'mhuang@usdigitalresponse.org', // fake email for testing
         name: 'Mindy Huant',
         agency_id: usdrAgency.id,
         role_id: roles[0].id,


packages/client/src/components/Modals/GrantDetails.vue
```diff
@@ -92,7 +92,7 @@
         :fields="assignedAgenciesFields"
       >
       <template #cell(actions)="row">
-        <b-button variant="danger" class="mr-1" size="sm" @click="unassignAgenciesToGrant(row)">
+        <b-button v-if="row.item.agency_id === agency.id" variant="danger" class="mr-1" size="sm" @click="unassignAgenciesToGrant(row)">
```
mattbroussard, line 95: I think this should check for membership in `agencies` -- any agency I can assign to I should be able to unassign from, I think (i.e. subagencies of my selected agency).

What if I accidentally assigned to an agency other than my own? As written I don't think this would let me unassign it even if I'm an admin who can change my `selectedAgency` to that agency.

mattbroussard (changes requested):

packages/client/src/components/Modals/GrantDetails.vue
Rapol: Correct. The /session returns user information which has the agency the user belongs. We can use user.agency.subAgencies to determine if the agency is one I have access to. 

But I think the list of agencies we show in the dropdown are already all agencies that i have access to so this wont do anything.

mhuang-usdr: Per #61 - only being able to assign to a subagency is a bug. Any user should be able to assign a grant to any other agency.

It sounds like the concern expressed here is "what happens if I assign something on accident"?

We have a WIP feature on the product/design side for a user to be able to track and manage all the grants they've assigned to someone else slated for the next major milestone. In the meantime, the workaround is for users to communicate to the other agency outside of the tool. It's not great, but it's MVP.

@Raphomet wdyt?

mhuang-usdr: From further discussion in the standup today:

Since we already track the user who assigned grant, let's go ahead and implement what @mattbroussard proposed - 

**any agency user has assigned, they should also be able to unassign**

Raphomet:
fox-jumping

packages/server/seeds/dev/ref/users.js
```diff
@@ -25,7 +25,7 @@ module.exports = [
     },
     {
         id: 2,
-        email: 'mindy@usdigitalresponse.org', // fake email for testing
+        email: 'mhuang@usdigitalresponse.org', // fake email for testing
```
mattbroussard, line 28: Note this will break some tests that refer to this email; be sure to grep for those and update the email there. See #63 and #67. #67 is to make this type of breakage less likely.

mattbroussard:
@davidbkay it seems like this isn't complete but Github shows it closed/merged since you merged into `staging`. But it's not merged into `main`.

It sounds like there isn't a way to "reopen", so for review on additional changes I think you'll have to open a separate PR off the same branch? https://stackoverflow.com/a/12838195

davidbkay:
Maybe...make the PR for main and just merge it into staging for verification?

PR #71, opened 2022-05-04
fix: added agency id check to remove agency button in agency table

### LINKED ISSUE #66: [ID Tool] Agency staff should not be able to see unassignment buttons to unassign other agencies

davidbkay:
Fixes #66


mhuang-usdr:
In staging, I verified these expected behaviors:

- I am able to remove assignment to my agency. 
- I am not able to remove assignment to an agency that I didn't assign to and do not belong to.

However, this expected behavior is not showing up
- I am able to remove an assignment that I assigned to another agency.

Rapol (approved):

packages/server/src/db/index.js
```diff
@@ -460,6 +486,27 @@ async function getAgencies(rootAgency) {
     return result.rows;
 }
 
+async function getAllAgencies() {
```
mattbroussard, line 489: This accessor should take a `tenantId` param and add a where clause to the query based on it. Then caller in `/allagencies/` endpoint (perhaps we should rename?) should filter based on `req.session.user.tenant_id`.

mattbroussard:
Note: will also need to update the permission check in `PUT /:grantId/assign/agencies` ([here](https://github.com/usdigitalresponse/usdr-gost/blob/e41ea733fee476d5ea95c32abc3f5e9162e83257/packages/server/src/routes/grants.js#L176)) to allow assignment to other agencies in same tenant that are not subagencies. @as1729 left a comment to this affect in draft PR #73 as well.

PR #83, opened 2022-05-25
chore: replace libsass with comparable dart-sass version

ajhyndman:
We also use Vue v2 in the arpa repo, but haven't had issues with Node 16.  I was curious what the bottleneck was in usdr-gost, so I poked around a little.

I think it might just be `node-sass` giving you headaches.

A couple years ago, `node-sass` (internally a distribution of the C implementation, libsass) [was deprecated](https://sass-lang.com/blog/libsass-is-deprecated) in favour of the more actively maintained `sass` (internally dart-sass) project.  Both are meant to have interchangeable node APIs, and sass-loader discovers whichever package you have installed.

I tried picking a version of `sass` from around the same time as the currently installed `node-sass`, and that seems to do the trick for me!

Now the repo runs in node 14 and 16, and I don't see any new sass warning or error feedback.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -43,7 +43,7 @@
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-vue": "^6.2.2",
     "geckodriver": "^3.0.2",
-    "node-sass": "^4.12.0",
+    "sass": "~1.17.4",
     "sass-loader": "^8.0.2",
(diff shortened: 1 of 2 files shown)


mattbroussard (approved):
Sorry this slipped and I never marked it approved. This seems pretty reasonable and should be safe I think -- `sass-loader`'s readme specifies it works with both versions.

PR #92, opened 2022-06-07
43 add spoc button to tenants

### LINKED ISSUE #43: [ID] Nevada's SPOC button appears even for non-Nevada tenants

davidbkay:
fixes #43 

note: this PR was originally opened as #58 which was accidentally closed by merging into `staging` -- there are a few comments there as well, that look like they were addressed before opening this one.


packages/client/src/components/Modals/GrantDetails.vue
```diff
@@ -164,20 +164,30 @@ export default {
...
+        (viewed) => viewed.agency_id.toString() === this.selectedAgencyId,
+      );
+    },
+    shouldShowSpocButton() {
+      if (!this.currentTenant.uses_spoc_process) return false;
+      return true;
     },
     interested() {
       if (!this.selectedGrant) {
         return false;
```
mattbroussard, line 186: nit: let's make this return `undefined` so the return type is consistent with what `find` could return below.

packages/server/src/db/index.js
```diff
@@ -440,11 +455,22 @@ function getInterestedCodes() {
...
+        'abbreviation',
+        'parent',
+        'warning_threshold',
+        'danger_threshold',
+    )
+        .from(TABLES.agencies)
+        .where({
+            id: agencyId,
+        })
+        .leftJoin('tenants', 'tenants.main_agency_id', '=', `${TABLES.agencies}.main_agency_id`);
```
mattbroussard, line 470: I think this should join `agencies.tenant_id = tenants.id` instead of using `main_agency_id` like this, in case you are querying about a subagency.

packages/server/migrations/20220417025208_add-spoc-setting-to-tenants.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable func-names */
+exports.up = function (knex) {
+    return knex.schema
+        .table('tenants', (table) => {
+            table.bool('uses_spoc_process').defaultTo(false);
```
mattbroussard, line 5: Can we set this column `notNullable`?

mattbroussard (approved):
One important change to the join in `getAgency`, then a few nits.

PR #94, opened 2022-06-07
chore: remove tenants tab

### LINKED ISSUE #60: Update the Tenants tab so it is only visible when tenant id = 1
The tenants tab was added for debugging purposes, but should be not be visible to every user.

Screenshot 2022-04-19 125248

davidbkay:
fixes #60

### DIFF
--- a/packages/client/src/components/Layout.vue
+++ b/packages/client/src/components/Layout.vue
@@ -1,43 +1,63 @@
 <template>
   <div>
     <b-navbar type="dark" variant="dark">
-      <b-navbar-brand href="/#/grants">Grants Identification Tool</b-navbar-brand>
+      <b-navbar-brand href="/#/grants"
+        >Grants Identification Tool</b-navbar-brand
+      >
 
       <b-collapse id="nav-collapse" is-nav>
-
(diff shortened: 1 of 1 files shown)


packages/client/src/components/Layout.vue
```diff
@@ -1,43 +1,63 @@
 <template>
   <div>
     <b-navbar type="dark" variant="dark">
-      <b-navbar-brand href="/#/grants">Grants Identification Tool</b-navbar-brand>
+      <b-navbar-brand href="/#/grants"
+        >Grants Identification Tool</b-navbar-brand
```
mattbroussard, line 5: There are some weird formatting changes in this diff, do you have some kind of editor integration doing this? Can you update the PR before merging to only contain the relevant changes (which looks like just one line right now)?

mattbroussard (approved):
Should we also in this PR delete the UI (`Tenants.vue`, `EditTenant.vue`, the reference to these in `router/index.js`, etc.) that was used in that tab now that it is dead code? Or do we foresee this being added back/reused at some point soon?

What about the VueX module `store/modules/tenants` that's only used from that UI?

PR #96, opened 2022-06-10
Add a clientside redirect to better handle email systems that may prefetch login link

### LINKED ISSUE #95: Login link limit hit by automated email processors
It seems that some email systems prefetch links from emails either on receipt or on click, causing us to count the link token as "used" before the actual login happens. We already [bumped](https://github.com/usdigitalresponse/usdr-gost/commit/781db143971a0f125c680eb2d4b8efbf7c520047) the limit from 1 to 2 link uses for this reason in the past, but it seems like that's not enough now.

There is a recent email thread from NV with subject line `NV Trouble Shooting_USDR Grants Tool Access Link` complaining about just this happening. On interactive debugging with them, it seems like the link is getting requested once on email receipt and 3 times (!) on click.

My suggested solution is to add some interstitial between the login link and actually counting the token as used that requires either a user click or a JS redirect (neither of which would be likely to be triggered by an automated email system). This is a best practice anyway since HTTP GET request are not supposed to cause state changes (and as such this is a [common problem](https://security.stackexchange.com/questions/211258/do-mail-servers-follow-links-in-emails-as-part-of-a-security-scan-before-inbox-d)).

We did find a workaround that works for NV but may not work for others: manually copy and pasting the link text from the email to browser location bar.

mattbroussard:
Fixes #95

### DIFF
--- a/packages/server/src/configure.js
+++ b/packages/server/src/configure.js
@@ -55,7 +55,7 @@ module.exports = (app) => {
         },
     });
     app.use(staticMiddleware);
-    app.use('/static', express.static(`${__dirname}/static`));
+    app.use('/static', express.static(resolve(__dirname, '../static')));
     app.use(
         history({
             disableDotRule: true,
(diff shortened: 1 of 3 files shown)


mhuang-usdr:
@Rapol are you able to review this PR?

Rapol (approved):

mhuang-usdr:
Matt is no longer around the project, so I'm merging this in on his behalf.

Issue #54, opened 2022-04-06
(3)[ID Tool] I should be able to unmark a grant as interesting and assigned.


mhuang-usdr:

Raphomet:
Just discussed this with @mhuang-usdr. There's a super low budget way of doing this that's not very pretty, but will get the job done for MVP.

The "Assigned" area already allows you to delete an assignment. It has a red trash can button. The super low budget solution is just to replicate that to the "Interested" area:

![CleanShot 2022-04-12 at 13 50 47@2x](https://user-images.githubusercontent.com/58006/163052640-12881e8f-cdf3-404d-a08c-cecae42b4a33.png)

The medium-effort way of doing this would be to improve the overall UX of this area. There are a number of minor usability issues that could be fixed with a redesign. Happy to talk through this if it's interesting, but for the sake of MVP-ability we may want to punt on that.

Raphomet:
(Note, of course, that the red trash button should only appear on YOUR row — the table may show other agencies that are "interested")

mhuang-usdr:
QA - LGTM

mhuang-usdr:
There was a regression, this is no longer fixed in `prod`.

mhuang-usdr:
Closing, superseded by #186

PR #107, opened 2022-06-24
can remove interested grants for loggedin user

anigonzales:
Issue #54

Issue solution: 
   Added actions tab in grant details view for "interested", added function to remove a grant from your list of interest grants, added shorter titles in some of the columns, only allows deletion for your interests, does not show trash can delete button for interests that aren't yours

Screenshots:

1.) What the interested agencies table looks like with delete functionality button of logged in user
Screen Shot 2022-06-23 at 6 54 18 PM

2.) What it looks like when it is not your own (this grant example was already past its date)
Screen Shot 2022-06-23 at 6 54 47 PM

Testing:
   Submitted different interested codes for multiple grants and compared with the grant with other user's interests (above photo). The trash can button did not appear for any other user's interests than my own.


install_nvm.sh
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/env bash
```
User1m, line 1: File should not be committed as part of project. Will need to remove this file from git

packages/client/src/components/Modals/GrantDetails.vue
```diff
@@ -67,7 +67,15 @@
       <b-table
         :items="selectedGrant.interested_agencies"
         :fields="interestedAgenciesFields"
-      />
+      >
+      <template #cell(actions)="row" >
+        <b-row v-if="loggedInUser.email === selectedGrant.interested_agencies[0]['user_email']">
```
User1m, line 72: `selectedGrant.interested_agencies[0]` looks like it's only checking the 1st agency - can we ensure the user won't be found in other agencies? eg. `selectedGrant.interested_agencies[1]` ? Should this comparison be more robust?

packages/client/src/components/Modals/GrantDetails.vue
```diff
@@ -166,6 +184,7 @@ export default {
       agencies: 'agencies/agencies',
       users: 'users/users',
       interestedCodes: 'grants/interestedCodes',
+      loggedInUser: 'users/loggedInUser',
```
User1m, line 187: 👍🏽

packages/client/src/store/modules/grants.js
```diff
@@ -53,6 +56,14 @@ export default {
         agencyIds,
       });
     },
+    unmarkGrantAsInterested(context, {
+      grantId, agencyIds, interestedCode, agencyId,
+    }) {
+      return fetchApi.deleteRequest(`/api/organizations/:organizationId/grants/${grantId}/interested/${agencyId}`, {
```
User1m, line 62: 👍🏽

install_nvm.sh
User1m: `git rm --cached install_nvm.sh`

User1m (changes requested):
Comments needing addressed

hadashirt (approved):

Issue #85, opened 2022-05-26
(3)Edit Agency is missing some functionality


mhuang-usdr:
Go to Agency tab > Edit icon. The modal is missing functionality to:

- [x] Change the name
- [x] Change the abbreviation
- [x] Change the parent agency
- [x] Delete the agency

<img width="508" alt="image" src="https://user-images.githubusercontent.com/91221265/170601362-0122d839-55d3-4377-9f2f-02f34c2b426b.png">

These fields (besides delete) already exist in the Add Agency modal.
<img width="337" alt="image" src="https://user-images.githubusercontent.com/91221265/170601738-232d1a10-cee1-40b9-b7db-f5847748e258.png">

User1m:
@mhuang-usdr is this still open?

mhuang-usdr:
It is in part. I'll pull out the remaining tasks into a separate issue

mhuang-usdr:
QA:  Parent Agency field shows the ID of the parent agency, but should show the Name of the Parent agency.

<img width="281" alt="image" src="https://user-images.githubusercontent.com/91221265/185464032-f641f29b-6992-414f-9377-b56dd5c737cc.png">

chrisxkeith:
Fixed by https://github.com/usdigitalresponse/usdr-gost/pull/313 .

PR #116, opened 2022-07-05
#85 Edit/Delete Agency Functionality

anigonzales:
Issue #85 

Issue Solution:
Added "name," "abbreviation" and "parent agency" input fields to allow the agency to be updated.
Added "admin delete agency" that only allows users to delete the agency. Confirmation shows up after clicking. If the user is not an admin, then it will be greyed out and show a tooltip on hover. 

Screenshots:
1.) The new fields and button for the edit agency modal
reg edit modal

2.) Input in one field will update the agency for that field only
change 1 input

3.) Agencies table reflects changes after form submission. (The column for the name is so long because I have all of the agencies on the table, so it adjusts the column size to the length of the longest name).
input changed table

4.) If the user is "admin," then the button will not be greyed out and will be clickable. After clicking it, a popup appears to confirm the deletion.
actual del popup

5.) If the user is not "admin," the button will grey out and not be clickable. A tooltip will appear on hover.
tooltip

Testing:
Added different inputs to each field and clicked submit, viewed those changes on the agencies table that the user can view.
The deletion button is disabled for any other user type. 
The tooltip only appears if the button is disabled. 
Deletion does not work if the agency has children.
Checked the usdr_grants db agency table in postgresql for the appropriate updates or deletions being reflected.

### DIFF
--- a/packages/client/src/components/Modals/EditAgency.vue
+++ b/packages/client/src/components/Modals/EditAgency.vue
@@ -207,7 +207,7 @@ export default {
         this.resetModal();
         this.$bvModal.hide();
       }
-      if (this.formData.parentAgency.id && (this.formData.parentAgency.id !== this.agency.id)) {
+      if (this.formData.parentAgency.id && (this.formData.parentAgency.id !== this.agency.id) && (this.formData.parentAgency.parent !== this.agency.id)) {
         this.updateAgencyParent({ agencyId: this.agency.id, parentId: this.formData.parentAgency.id });
         this.resetModal();
         this.$bvModal.hide();
--- a/packages/client/src/components/Modals/EditAgency.vue
+++ b/packages/client/src/components/Modals/EditAgency.vue
@@ -85,8 +85,8 @@
             required
           ></b-form-input>
         </b-form-group>
-        <form ref="form" @submit.stop.prevent="handleDelete">
-          <b-button v-if="userRole === 'admsn'" variant="danger">Admin Delete Agency</b-button>
+        <form ref="form" @click="handleDelete">
+          <b-button v-if="userRole === 'admin'" variant="danger" >Admin Delete Agency</b-button>
         </form>
       </form>
     </b-modal>
@@ -139,6 +139,7 @@ export default {
   },
   computed: {
     ...mapGetters({
+      agencies: 'agencies/agencies',
       userRole: 'users/userRole',
     }),
   },
@@ -150,6 +151,7 @@ export default {
       updateAgencyName: 'agencies/updateAgencyName',
       updateAgencyAbbr: 'agencies/updateAgencyAbbr',
       updateAgencyParent: 'agencies/updateAgencyParent',
+      deleteAgency: 'agencies/deleteAgency',
     }),
     resetModal() {
       this.$emit('update:agency', null);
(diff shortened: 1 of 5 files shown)


packages/client/src/components/Modals/EditAgency.vue
```diff
@@ -44,6 +84,16 @@
...
           ></b-form-input>
         </b-form-group>
+        <form ref="form" @click="handleDelete">
+          <span id="disabled-wrapper" class="d-inline-block" tabindex="0">
+            <b-button v-bind:disabled="userRole !== 'admin'" style="pointer-events: none;" variant="danger">
+              Admin Delete Agency
+            </b-button>
+          </span>
+          <b-tooltip v-if="userRole !== 'admin'" target="disabled-wrapper" triggers="hover">
+            You cannot delete an agency with children. Reassign child agencies to continue deletion.
```
User1m, line 94: Tooltip msg is kinda confusing. 
Can you not delete an agency b/c you are not an admin? or b/c the agency has children?

packages/server/src/routes/agencies.js
```diff
@@ -26,6 +27,39 @@ router.put('/:agency', requireAdminUser, async (req, res) => {
...
+    const result = await setAgencyAbbr(agency, abbreviation);
+    res.json(result);
+});
+
+router.put('/parent/:agency', requireAdminUser, async (req, res) => {
+    const { agency } = req.params;
+
+    // const { parentId } = req.body.parentId;
+    // console.log('reeeeeeqqqqq  ' + req.body.parentId);
+    const result = await setAgencyParent(agency, req.body.parentId);
```
User1m, line 59: Might want to cast `req.body.parentId` as a Number see [here](https://github.com/usdigitalresponse/usdr-gost/pull/116/files#diff-067094b50695ccd7d6188f26c341f3d1cd1213ffc6a762abcb069b5a7db841c4L38)

packages/client/src/components/Modals/EditAgency.vue
anigonzales: If the user is not an admin or if the agency has children, the button will be disabled and the tooltip will appear on hover

User1m (approved):

mhuang-usdr:
[QA]

A few things I noticed

**1. Upon clicking edit Agency - the fields are blank. They should be pre-populated with the existing values. **

So for this example, the name field should be prepopulated with "Administration: Administrative Services", the parent agency field should be pre-populated with "Administration, Department of", etc.

image


**2. I am still able to set the agency as its own parent - I shouldn't be able to do that.**

image

**3. A user should not be able to delete the agency they belong to, because that would orphan the user. Basically, since I am a member of the USDR agency, the "Delete Agency" button should be grayed out.** 

**4. Let's also rename "Admin Delete Agency" to simply "Delete Agency"**

image

**5. I am able to trigger the "Delete Agency" button even when I'm not clicking on the button itself. I am able to trigger it by clicking in the highlighted area as indicated below:**

image

**6. Is there any way to move the delete button to the bottom right corner of the modal?**
image

**7. The confirmation modal - let's rename the "OK" button to "Delete Agency" and make the delete button red.**

image

**8. For that same confirmation modal, I noticed that if click "Cancel" the entire "Edit Agency" modal disappears**. 
Let's change the behavior so that only the confirmation modal disappears when I click "Cancel" - the "Edit Agency" modal that is underneath should still remain. I recorded a quick Loom to show you what I mean.

https://www.loom.com/share/139ca4b065c24e98b1dc9bab04a1cb35


We're getting close - thank you!

Issue #111, opened 2022-06-27
[ID][3] Create frontend template for agency activity table on dashboard


hadashirt:

mhuang-usdr:
QA: LGTM

PR #134, opened 2022-07-18
111 template agency table

anigonzales:
Issue #111 
Given Mockup:
IMG_3329

Issue Solution:
Added empty columns next to the total grant columns to display the monetary value. Since this is just the frontend template, the functionality is not yet there.
Added spacing between columns  to make is more concise and readable to the user.
Added the characters "($...)" to the monetary value columns so they will automatically display in that form.
Added css styling to the appropriate characters, following the provided mockup.

Screenshot:
1.) The table fomatted according to the mockup. The values are null since the backend api has not been created or implemented yet. Other values can still be placed in there (refer to "Testing").
Screen Shot 2022-07-18 at 11 12 32 AM

Testing:
Added different values for the agency (such as agency_id and abbreviation) to test if they are displayeed with the correct formatting and css styling.


CGowan44:
@mhuang-usdr It looks like this follows what I wrote in the design doc saying the monetary value columns will be sortable instead of the number of grants. Did you mention we should not be able to sort by money for now but instead the number of grants in each category like it was originally? I will make sure to update the document as well.

mhuang-usdr:
@CGowan44 yes - let's make it sortable by number of grants, not by the sum of the money.

CGowan44 (approved):

Issue #133, opened 2022-07-18
(4) Create backend functionality / API calls to obtain data for upcoming grants column


hadashirt:

mhuang-usdr:
No QA needed for pure backend work.

PR #144, opened 2022-07-20
133 upcoming grants table backend

anigonzales:
Issue #133 

Issue Solution: 
Created getClosestGrants() query to get 3 grants that have the closest close dates in comparison to the current date, returns the title and date of the of the grant since the Upcoming Closing Dates sections will only display those pieces of grant information. The API for getting the abbreviations of all agencies interested (to be placed under the grant name) already exists (used in the Grants Table "interested Agencies" column).

Screenshots:
Only backend API work and no frontend implementation

Testing:
Checking the grants date in the Grants Table using the sorting feature and comparing them to the query return value.

### DIFF
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -341,14 +341,17 @@ async function getGrant({ grantId }) {
 }
 
 async function getClosestGrants() {
-    const query = await knex('grants')
+    const timestamp = new Date();
+    const query = await knex(TABLES.grants)
         .select('title')
-        .orderby('close_date', asc)
+        .where('close_date', '>=', timestamp)
+        .orderBy('close_date', 'asc')
         .limit(3)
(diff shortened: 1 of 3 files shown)


packages/server/src/db/index.js
```diff
@@ -340,6 +340,16 @@ async function getGrant({ grantId }) {
     return results[0];
 }
 
+async function getClosestGrants() {
+    const timestamp = new Date();
+    const query = await knex(TABLES.grants)
+        .select('title', 'close_date')
+        .where('close_date', '>=', timestamp)
+        .orderBy('close_date', 'asc')
+        .limit(3)
```
User1m, line 349: future iterations might take the limit as in API input from the client instead of "hard-coding" 3

User1m (approved):

PR #180, opened 2022-07-28
Fix 129: Interested and Rejected Grants views should be Agency-scoped, not User-scoped 

bspates:
### Description
The MyGrants view was being limited to grants marked as interested by the current user, when they should see all grants for an agency. 

I made it so that the showInterested flag did not filter by the user id on the grants_interested table, and made a new flag that could do that filtering.

One thing to bring up, which may be outside the scope of this fix, is that it looks like the grants_interested table has the user_id of the creator as a foreign key, that if the user got deleted would delete this linking record. 

https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/server/migrations/20201214213110_grants_interested.js#L13

If a user is deleted would we want the record of interest in a grant for the agency to also be deleted? Based on this bug my intuition says no, but I'm not sure. 

### Screenshots / Demo Video

https://user-images.githubusercontent.com/3039734/181634869-2ba157d2-5f34-4403-af82-f19b79b5ea71.mov

### Testing
I did some manual testing. I am unfamiliar with the project so if I should add some front end unit/integration tests pls point me in the right direction. 

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/GrantsTable.vue
+++ b/packages/client/src/components/GrantsTable.vue
@@ -92,6 +92,7 @@ import GrantDetails from '@/components/Modals/GrantDetails.vue';
 export default {
   components: { GrantDetails, Multiselect },
   props: {
+    showMyInterested: Boolean,
     showInterested: Boolean,
     showRejected: Boolean,
     showAging: Boolean,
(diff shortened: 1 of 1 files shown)


bspates:
Hi :wave: 
I'm new, let me know if I did anything wonky.

bspates:
@Rapol Thanks for the review. It is actually only a one file change. The only uncommitted file is the `yarn.lock` and the `packages/server/seeds/dev/index.js`. It seems like the issue was just how the front-end was filtering the request, not the structure of the data. Let me know if I'm missing something.

Rapol:
@bspates oh my so sorry 😅

Rapol (approved):

Issue #84, opened 2022-05-26
Set up Nevada as a tenant


mhuang-usdr:
- [ ] Add "State of Nevada" as a tenant
- [ ] Make the root agency "Administration: Grant Office"
- [ ] Add "grants.dev@usdigitalresponse.org" as an admin to that tenant

mhuang-usdr:
@slants3 - I remember now! We set up `Nevada Tenant` as a test development tenant, but we meant to blow it away and create a new one before launching with NV.

Can you rename `Nevada Tenant` to `Test Tenant` (there are people dev-ing in this tenant so I don't want to blow it away), and create a new tenant to the specs in the original description please?

If you can make this a script, that'll allow me to set up a new tenant with ease in the future as well :)

slants3:
@mhuang-usdr got it!

To clarify, the tenant creation process should not be available as an UI element on the site correct but should instead only be accessible via a script in the console or something along those lines?

Also do you know if we have existing scripts for anything else?

mhuang-usdr:
@slants3 - eventually it'd be great to have a UI element on the site, but the faster (and very acceptable) solution to build and maintain is to have a script.

I don't think we have any in this repo, but you can check out this sister repo (different tool, but still working in the Grants space) for an example - https://github.com/usdigitalresponse/arpa-reporter/tree/staging/scripts

chrisxkeith:
```
gost_grants_tools_db=> select * from users where name = 'USDR Dev NV';
 id |                email                |    name     | role_id |          created_at           | tags | agency_id | tenant_id
----+-------------------------------------+-------------+---------+-------------------------------+------+-----------+-----------
 56 | grants.dev+nv@usdigitalresponse.org | USDR Dev NV |       1 | 2022-08-10 21:32:03.591422+00 |      |       384 |         2
```

PR #189, opened 2022-08-01
Introduce Create Tenant API Endpoint and "USDR Super Admin" access helper

slants3:
### Ticket #
[84](https://github.com/usdigitalresponse/usdr-gost/issues/84)

### Description
To add new tenants, we need to create an API endpoint.

This replicates our existing API infrastructure currently used for adding users.


packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,28 @@ router.get('/', requireUser, async (req, res) => {
...
+router.post('/', requireAdminUser, async (req, res) => {
+    if (!req.body.name || !req.body.agency_id) {
+        res.status(400).send('Tenant name and agency ID is required');
+        return;
+    }
+
+    try {
+        const tenant = {
+            display_name: req.body.name,
+            main_agency_id: req.body.agency_id,
```
Rapol, line 29: i might be getting this confused, since this tenant entity is new to me. A tenant is the main entity and agencies are then created under that tenant, correct? So I think we will not have agency_id.

1. Create Nevada tenant
2. Create Grant Office tenant
3. Update Nevada tenant as the main agency

Im not entirely sure what is the usage of main_agency_id

mhuang-usdr: the agencies are in a tree hierarchy under the tenant. `main_agency_id` is the root of the tree - the top-level agency that belongs to the tenant. This agency is required when the tenant is first created.

packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,28 @@ router.get('/', requireUser, async (req, res) => {
...
+        };
+
+        const result = await createTenant(tenant);
```
bspates, line 32: Should we be doing any validation/sanitization of the request data before sending it  to the DB, or is that built in somewhere else?

packages/server/src/routes/tenants.js
slants3: @mhuang-usdr I'm a little confused - if the agency belongs to the tenant, couldn't it be possible for a tenant to be created before its agency is created?

packages/server/src/routes/tenants.js
slants3: That's a great idea - as far as I understand it there aren't any sort of validation built into our endpoints as is the case here with [users](https://github.com/usdigitalresponse/usdr-gost/blob/4389f12e9aa797b77d4102337d3480491c9efa27/packages/server/src/routes/users.js#L8) as well.

Since we don't have any validation or sanitation anywhere - what do you think about adding this endpoint for now to unblock https://github.com/usdigitalresponse/usdr-gost/pull/189 and then creating a separate issue where we add validation + sanitation to all of our endpoints?

bspates: Makes sense to me.

packages/server/src/routes/tenants.js
mhuang-usdr: @slants3 I apologize, I realize forgot to explain this earlier!

The requirements for the app to function as expected are:
- Every tenant must have at least one agency - the main agency 
- Every agency must have at least one user, and they must be an admin.

Thus the requirements for spinning up a new tenant are:
- A new tenant in the tenant database
- A new agency assigned to the new tenant ("Grants Office" by default)
- A new admin user ("grants.dev@usdigitalresponse.org" by default) assigned to that tenant

I'm not sure what the best order or best practices to create those entries are, that's up to the eng team!


(Also, a possible snag - iirc users _might_ still be keyed by their email. In which case having the same user (grants.dev@usdigitalresponse.org) across multiple tenants will cause issues. The desired experience is that we _are_ able to have the same user in multiple tenants, but in the interim you can use "grants.dev+nv@usdigitalresponse.org" for the Nevada tenant.

slants3: Makes total sense, thanks for explaining that!

mhuang-usdr: @slants3 we should create a new agency when we create a new tenant. Going forward, we should never need to reuse an agency for a new tenant. 

That is only happening right now due to a database cleanup (there are a lot of things in there that got really messed up)

slants3: I see - what is the workflow for the agency + tenant creation? I'm trying to understand if they should be happening synchronously or if the new agency can be created first and then just passed into the tenant creation as an `main_agency_id`.

mhuang-usdr: @slants3 no current existing workflow. The agency should be created as the tenant is created.  Here's an example user journey.

Mindy, the PM, wants to onboard a new tenant. She runs the script (or, when we have a UI down the road, clicks the button), and a new tenant will be automatically set up with the main agency and default admin user (grants.dev@usdigitalresponse.org)

packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,40 @@ router.get('/', requireUser, async (req, res) => {
...
+    try {
+        const agency = {
+            name: req.body.agency_name,
+            abbreviation: req.body.agency_abbreviation,
+            parent: 0,
+            warning_threshold: 30,
+            danger_threshhold: 15,
+        };
+
+        const createdAgency = await createAgency(agency, 15);
```
slants3, line 35: Hey @bspates, I had a quick question for you - I'm using the updated `createAgency` endpoint but I'm unable to successfully create an agency and the creation hangs. I was able to create before the change - but as far as I can tell I'm using this correctly. Would you have an idea on what I'm doing wrong here?

bspates: @slants3 Do you get any error? What is the hardcoded number 15? Shouldn't you be using the user id that is making the request?

bspates: @slants3 Also I see the tenant creation is happening after this. You'll want to swap those around. How the foreign key dependencies in the tables work now is that only main_agency_id in tenant can be left null to start. So you have to create the tenant, then the agency, then go back and update the main_agency_id in the tenant.

packages/server/src/db/index.js
```diff
@@ -569,6 +569,18 @@ async function getTenant(main_agency_id) {
...
+async function createTenant(tenant) {
+    const response = await knex
+        .insert(tenant)
+        .into('tenants')
+        .returning(['id', 'created_at']);
+    return {
+        ...tenant,
+        id: response[0].id,
+        created_at: response[0].created_at,
+    };
```
mattbroussard, line 581: Why not just get the whole row from Postgres and return that?

packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,40 @@ router.get('/', requireUser, async (req, res) => {
     res.json(response);
 });
 
+router.post('/', async (req, res) => {
```
mattbroussard, line 20: This route doesn't seem to check any kind of permission. I think we might want a new helper like `requireUSDRAdminUser` that calls `requireAdminUser` and additionally verifies the user is part of a hardcoded agency ID (`1`, I think).

packages/server/src/routes/tenants.js
mattbroussard: >Shouldn't you be using the user id that is making the request?

👍, should use `req.session.user.id` (available when using `requireAdminUser` as suggested in a separate comment)

>So you have to create the tenant, then the agency, then go back and update the main_agency_id in the tenant.

👍, this should also mean `setAgencyTenantId` is not necessary since agencies must be created with a `tenant_id` to start with and cannot move between tenants.

packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,40 @@ router.get('/', requireUser, async (req, res) => {
...
+
+        const createdAgency = await createAgency(agency, 15);
+
+        const tenant = {
+            display_name: req.body.tenant_name,
+        };
+
+        const createdTenant = await createTenant(tenant);
+
+        _ = await setAgencyTenantId(createdAgency.id, createdTenant.id);
```
mattbroussard, line 43: nit: remove `_ =`

packages/server/src/routes/tenants.js
```diff
@@ -17,6 +17,40 @@ router.get('/', requireUser, async (req, res) => {
...
+    } catch (e) {
+        if (e.message.match(/violates unique constraint "agencies_name_unique"/)) {
+            console.log(e.message);
+            res.status(400).send('Agency with that name already exists');
+        }
+    }
+});
+
 router.put('/:tenant', requireAdminUser, async (req, res) => {
     // Currently, tenants are seeded into db; only display name is mutable.
```
mattbroussard, line 55: This comment will no longer be true; revise.

mattbroussard (changes requested):
Is the idea that this will be called from some UI (e.g. added to currently-unused `Tenants.vue`?) that is gated to only USDR admins? Or that this endpoint will be called from some script we run? If the latter, why have an endpoint at all vs. having a script that just does this directly?

mattbroussard:
Just ran across #193 -- is this intended to supersede that, or will we have _both_ UI for this and a script?

packages/server/src/routes/tenants.js
slants3: > Just ran across #193 -- is this intended to supersede that, or will we have _both_ UI for this and a script?

Referencing this - I think in the immediate term we're going to use a script, but in the long term we will eventually want this to be an UI element

mattbroussard:
As [mentioned](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1662679750039039) on Slack, taking this over because I need this logic for ARPA Integration.

In update a bit ago, I've mostly rewritten much of the PR. Notably:
- Reuses most of the logic I wrote in #193's script in the endpoint
  - I also moved the file. Since there is both a move and modifications to it, Github confusingly shows the entire file `tenant_creation.js` as new, but the bulk of the logic in it was actually implemented in #193 and has just been slightly reorganized in this PR.
- Adds a new access check middleware, `requireUSDRSuperAdminUser` based on a hardcoded tenant ID, and uses it on these routes that view/modify across tenants.
  - However, it looks like in prod there are lots of non-USDR users as "admins" with tenantId=1. @mhuang-usdr is there a different condition we should be using for "super admin", or should we clean up this prod data so only actual USDR staff/volunteers are in the USDR tenant?
- Some small tweaks to the UI of the tenants tab (gating its display (superseding #227), showing all tenants for super-admins)

I might go ahead and hack up a simple "Create Tenant" modal since it seems silly to have this endpoint but nothing that calls it?

packages/server/src/lib/access-helpers.js
```diff
@@ -4,6 +4,16 @@ function isPartOfAgency(agencies, agencyId) {
...
 
+// TODO: in prod, there are a bunch of non-USDR looking users set as admin in USDR tenant?
+// some are CTG interns, but some are @cityoftulsa.org, @treasury.gov, etc.
+// probably need to move them out
+const USDR_TENANT_ID = 1;
+const USDR_AGENCY_ID = 0;
+function isUSDRSuperAdmin(user) {
+    // Note: this function assumes an augmented user object from db.getUser(), not just a raw DB row
+    return (user.tenant_id === USDR_TENANT_ID && user.agency_id === USDR_AGENCY_ID && user.role_name === 'admin');
+}
```
mattbroussard, line 15: Thinking more about this: maybe the right thing to do actually is to create a third `role` instead of hard-coding tenant/agency IDs like this. Would require at least:
1. Finding places that hardcode checks for admin role to also allow super-admin (including in soon-to-be-merged ARPA Reporter code)
2. Adding a filter to what roles someone can set a user to when creating or editing a user. Only super-admin can designate others as super-admin.

@bspates @mhuang-usdr thoughts?

packages/server/src/db/tenant_creation.js
```diff
@@ -0,0 +1,235 @@
+require('dotenv').config();
```
ajhyndman, line 1: Nit: I suspect this might be redundant with #323

mattbroussard: I don't think anything bad happens if you call it multiple times, but yeah, this will be unnecessary after #323 merges.

packages/server/src/routes/tenants.js
```diff
@@ -1,27 +1,37 @@
...
+
+    res.json(result);
+});
+
+router.put('/:tenantId', requireUSDRSuperAdminUser, async (req, res) => {
+    // Only display name is mutable after creation.
+    const { tenantId } = req.params;
     const { displayName } = req.body;
-    const result = await setTenantDisplayName(tenant, displayName);
+    const result = await setTenantDisplayName(tenantId, displayName);
```
ajhyndman, line 34: Are these routes used anywhere yet?  I'd just as soon not support creation of a tenant via API if not.

packages/server/src/db/tenant_creation.js
```diff
@@ -0,0 +1,235 @@
...
+            name: agencyName,
+            abbreviation: agencyAbbreviation,
+            code: agencyCode,
+            parent: null,
+            // main_agency_id is non-nullable, but trns.ref('id') doesn't seem to work. So we have
+            // to set to an existing value, then update it afterward.
+            //
+            // TODO(mbroussard): should we remove the NOT NULL constraint and have null mean root
+            // agency vs. having self-loops (similar to parent field)?
+            main_agency_id: await firstAgencyId(trns),
```
ajhyndman, line 179: This is definitely awkward.  And it should be redundant with the `main_agency_id` field in the `tenants` table, is that correct?

I would definitely prefer to land a migration that removes this redundancy and updates any queries rather than find a weird workaround for this self-reference.

ajhyndman (commented):
I did give this a read through, and was able to run the script and verify it works!

I'm not as familiar with all the context here, so I left a couple questions.

packages/server/src/routes/tenants.js
mattbroussard: The "list tenants" and "update display name" do have UI that calls them; the "create tenant" one does not yet.

It's not pressing since we have the script implemented in #193, but this PR was open already and I wanted to make sure that if/when it did merge it correctly handled the ARPA tables.

packages/server/src/db/tenant_creation.js
mattbroussard: I think you're right that these fields are redundant and we could get rid of the one on `agencies`, though a few callsites might need to do an extra db call to get the `tenants` row.

mhuang-usdr:
Sorry, I just saw these questions. >_<

>   * However, it looks like in prod there are lots of non-USDR users as "admins" with tenantId=1. @mhuang-usdr is there a different condition we should be using for "super admin", or should we clean up this prod data so only actual USDR staff/volunteers are in the USDR tenant?

We will be cleaning the USDR up so that only USDR users are admins with tenantID=1 and shifting everyone else out to a Demo tenant. We don't necessarily need a super-admin role in that case, but if the work is already done don't undo it!

packages/server/src/db/tenant_creation.js
mattbroussard: Filed this as a separate issue, #357. Will not be addressing in this PR.

packages/server/src/lib/access-helpers.js
mattbroussard: Punting on this for now because it adds some additional complexity (though if we ever want more than 2 roles we may end up adding such complexity anyhow). Instead, I'm restricting this "super admin" permission further by email (must be `@usdigitalresponse.org`) and filed a follow-up task #359 to clean up non-USDR users in USDR tenant and relax the email requirement.

mattbroussard:
I think this should be ready to merge unless you have other changes you'd like to see @ajhyndman. If not, can you mark the PR approved so I can merge?

I've filed the following follow-up tasks to be handled separately:
- #357
- #358
- #359

packages/server/src/routes/tenants.js
mattbroussard: Filed follow-up task #358 to implement such UI.

mattbroussard (approved):

PR #191, opened 2022-08-01
190 subagency can now delete their interested status

CGowan44:
### Description

Adds the ability for a subagency to delete their interested status. 

Note/Question: It looks like unmarking a grant as interested is tied to the user id, so if multiple subagencies and the parent agency all have a grant marked as interesting and one of the agencies deletes their interested status, all of the associated agencies will have their status deleted too. Is this the expected behavior?

### Screenshots / Demo Video

(logged in as usdrtest3)

Before: 
before 8-1

After
after 8-1

### Testing

Mark a grant as interesting with a subagency and check that the status can be deleted correctly. Also test that it works with parent agencies.

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/Modals/GrantDetails.vue
+++ b/packages/client/src/components/Modals/GrantDetails.vue
@@ -61,7 +61,7 @@
       <b-table :items="selectedGrant.interested_agencies" :fields="interestedAgenciesFields">
         <template #cell(actions)="row">
           <b-row
-            v-if="loggedInUser.email === row.item.user_email && (row.item.agency_name === loggedInUser.agency_name)">
+            v-if="loggedInUser.email === row.item.user_email && (String(row.item.agency_id) === selectedAgencyId)">
             <b-button variant="danger" class="mr-1" size="sm" @click="unmarkGrantAsInterested(row)">
               <b-icon icon="trash-fill" aria-hidden="true"></b-icon>
             </b-button>


cavukovic (approved):
UX team might want the trash can to be moved over a bit but you didn't create that functionality in this pr so I'll leave that up for another ticket if we want it

mhuang-usdr:
Agree with what @cavukovic said.

CGowan44:
@mhuang-usdr Forgot to tag you here yesterday about the note/question in the main PR comment above. Should unmarking a grant as interesting be linked to not only user id but also agency id so all of the user associated agency interests in a grant don't get deleted at once?

mhuang-usdr:
I'm not too sure I understand the question, but here's what I think the answer is (tell me if not)

- The Interested/Assign/Rejected lists of grants are owned by agency, NOT by user.
- This means that if a grant is unmarked by one user on behalf of an agency, it should show up as unmarked for everyone who is a member of or views that agency.

Let me know if that helps!

CGowan44:
@mhuang-usdr That definitely helps. I think what is happening is that when a grant is unmarked as interested, every entry in the grants_interested table with a matching pair for the given grant id and currently logged in user id is deleted without checking the agency id at all. So if in the picture above I were to hit the trash can, all 4 of those interested statuses would be deleted, not just the one (since they are all on the same grant with my user id). So should this deletion instead only apply to the agency/subagency status that is selected (the row with the trash can)? And should the user who created the interested status remain the only one who can delete the entry or could others from the agency as well?

mhuang-usdr:
Ah, thanks for clarifying!

>  So should this deletion instead only apply to the agency/subagency status that is selected (the row with the trash can)?
Yes. It should only delete the row that is clicked.

>  And should the user who created the interested status remain the only one who can delete the entry or could others from the agency as well?

Anyone from the agency should be able to delete an entry for that agency.

PR #193, opened 2022-08-01
84: Add Tenant Creation Script

slants3:
### Ticket #
[84](https://github.com/usdigitalresponse/usdr-gost/issues/84)

### Description
Introducing a script to add tenants with the following command:

```
yarn add_tenant tenantName={NAME} agencyId={AGENCY_ID}
```


package.json
```diff
@@ -10,6 +10,7 @@
     "packages/*"
   ],
   "scripts": {
+    "add_tenant": "cd packages/server && node src/lib/addTenantScript",
```
Rapol, line 13: This is great, but might be better to have a reusable migration or seed file using knexjs. That way we gain predictable db configuration and initialization setup and also adding tenants will be shown in the migrations table.

Rapol (approved):

package.json
slants3: If I'm understanding you correctly - wouldn't that only create `tenants` in a test database?

I'm imagining us running this in the Render terminal shell and directly adding `tenants` to our production DB that way

packages/server/src/lib/addTenantScript.js
```diff
@@ -0,0 +1,39 @@
...
+        const agency = await db.getAgency(tenantAgencyId);
+
+        if (!agency.length) {
+            throw new Error('No Agency with this ID exists');
+        }
+        const tenant = {
+            display_name: tenantName,
+            main_agency_id: tenantAgencyId,
+        };
+        const result = await db.createTenant(tenant);
```
Rapol, line 28: im not familiar with tenant terminology maybe tomorrow yall can ramp me up on this. I thought tenants were top level entities (Tenant > Agencies) so in this case how does tenantAgencyId exists?

package.json
Rapol: [Knex.js migration/seeds cli](http://knexjs.org/guide/migrations.html#migration-cli) can target any environment. It is standard to have seeds folder structure like
- seeds
  - development
  - staging
  - production
     - ...data

I was thinking that we would have the tenants and their agencies in csv files that a seed can then insert into a database. This will help us modify agencies easily and also gives us historical data of the agencies since it will be in git.

I think we should still add the script that you added since it will be useful for local dev and quick production changes if needed.

Rapol (approved):

mattbroussard:
As [mentioned](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1662679750039039) on Slack, taking this over because I need this logic for ARPA Integration.

In update a moment ago, I've mostly rewritten the script actually. Notably:
- Using `inquirer` instead of argv to make allow validation messages and make it easier for a non-eng to run
- Create a new root agency and admin user instead of moving an existing agency from one tenant to another

package.json
mattbroussard: I decided to keep this as a script, not a seed, because it requires user input and is intended to be invoked manually in one-off instances on prod. The tenants created there need not also be created in dev environment.

packages/server/src/lib/addTenantScript.js
mattbroussard: A _tenant_ can be thought of as an isolated namespace/world for a given state using GOST. Within one, users can only see and manage agencies etc. within the same tenant -- the only cross-tenant actions that exist are creation of new tenants, which only USDR super-admins will be allowed to do.

As such, I updated this script to create a new root agency in the new tenant, as opposed to moving an existing agency from one tenant to another as you commented on.

mattbroussard:
Since this has been accepted, I can technically merge now -- so I'm going to to unblock work in my other branch.

However, the script has been substantially written so @Rapol / @ajhyndman I'd appreciate a brief post-merge look. This script is standalone so this change shouldn't impact anything else.

PR #206, opened 2022-08-04
201 update unmark grant as interesting functionality

CGowan44:
### Description

* Parent agencies now have the option to unmark statuses for children
* Unmarking is available to other users in the same agency or a parent agency, it is no longer only available to the user who created the status
* Unmarking one agency/subagency as interested from a grant now only deletes that row or status and not every status for other agencies entered by the user for that grant
* Fixed a bug where the agencies list in GrantDetails was not updating when changing selected agency/subagency
* Note: This would be a good thing for multiple people to test together in staging to make sure it works with different users

### Screenshots / Demo Video

With usdr3, a USDR subagency, selected:

With USDR selected, all child statuses are available to be deleted:

### Testing

* Check that parent agencies can delete their own and their child agency statuses, but children cannot delete parent statuses
* Check that other users in the same agency or a parent agency are able to unmark a status
* Check the assignable agency dropdown in GrantDetails to make sure it is the correct subagency list for the selected agency/subagency (this list was used in checking ability to unmark)

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/Modals/GrantDetails.vue
+++ b/packages/client/src/components/Modals/GrantDetails.vue
@@ -61,7 +61,7 @@
       <b-table :items="selectedGrant.interested_agencies" :fields="interestedAgenciesFields">
         <template #cell(actions)="row">
           <b-row
-            v-if="loggedInUser.email === row.item.user_email && (String(row.item.agency_id) === selectedAgencyId)">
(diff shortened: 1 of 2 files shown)


cavukovic (approved):
Looks good and agree that this seems like something that is gonna be really fully tested when it gets to staging and multiple people can play around with it

Issue #72, opened 2022-05-04
Flesh out multitenancy


mhuang-usdr:
### **Tenant-scoped**

_Definition_: grab all agencies assigned to a tenant, for example children, grandchildren AND parents, grandparents, AND cousins, etc. Everything belonging to a tenant

_Places in the UI that should be tenant-scoped are:_

- [x] Users
- [x] Agencies (see #11 for more bg info)
- [x] Profile settings > change agency
- [x] Browse Grants > Viewed By
- [x] Browse Grants > Interested Agencies
- [x] Click into any grant modal > Interested Agencies
- [x] Click into any grant modal > Assigned Agencies
- [x] Click into any grant modal > Assigned Agencies dropdown (a la #61 )

### **Agency scoped** 

_Definition:_ the agency only. No children or grandchildren either, just the self.

Places in the UI that should be agency-scoped are:

- [x] Dashboard - All the grey boxes (Total Grants Matching Search Criteria, New Grants Matching, Updated Grants Matching Search Criteria, Total Viewed Grants, Total Interested Grants)
- [x] Keywords
- [x] Eligibility codes
- [x] My Grants > Interested
- [x] My Grants > Assigned

### **Agency + child scoped** 

_Definition:_ the agency + children and grandchildren.

Places in the UI that should be agency+child-scoped are:

- [x] Dashboard - Total Interested Grants by Agencies, Recent Activity, Upcoming Closing Dates
- [x] Agency View Switcher - Profile picture > Settings

bspates:
Copying over discussion from slack:
```
Brian:
Similar question to earlier, but slightly different context. For displaying users on the users tab, right now it is scoped such that an admin user can only see other users at their agency level or below. This should now be tenant wide as well?

Mindy: 
Tl;dr - same as the dropdowns - let's scope such that all users can see everyone in the tenant, but only admins can edit users that lie within their agencies or subagencies.My thought process below:

    The concerns of seeing all users in a tenant is information overload.

    Right now we have the ability to sort by agency, so while the situation is not ideal, the overload isn't as bad as just a straight list.

    The pros of seeing all users would help in breaking down information silos - allows users to help pinpoint who's working on what  

    I've asked govts before whether they want to keep their activity secret from other agencies, and the answer is a NO - their problem is always that they don't have enough visibility

    Thus, the ideal solution is probably for all users to be able to see everyone, and also introduce a better search/sort/filter to the tables down the road.
[...]

PR #227, opened 2022-08-08
60: Only Display Tenants Tab to USDR Tenant Users

slants3:
### Ticket #
[60](https://github.com/usdigitalresponse/usdr-gost/issues/60)

### Description
Tenants tab was introduced for debugging purposes but should only be visible to users that are members of the USDR Tenant.

To do this, we had to join the `Tenants` table with the `User` table on the `User` fetch request and do a lookup to see the `Tenant` that the `User` is a part of. Only display the tab if they are a part of the correct `Tenant`.


packages/client/src/components/Layout.vue
```diff
@@ -31,7 +31,7 @@
         <b-nav-item to="/keywords" exact exact-active-class="active">Keywords</b-nav-item>
         <b-nav-item to="/users" exact exact-active-class="active" v-if="userRole === 'admin'">Users</b-nav-item>
         <b-nav-item to="/Agencies" exact exact-active-class="active">Agencies</b-nav-item>
-        <b-nav-item to="/Tenants" exact exact-active-class="active">Tenants</b-nav-item>
+        <b-nav-item to="/Tenants" exact exact-active-class="active" v-if="tenant.id === 1">Tenants</b-nav-item>
```
bspates, line 34: I'll preface this with, I am not familiar how this all works together. 

Hard coding the tenant id in the front end seems bad. Is there another flag that can be added, "superuser" or the like to designate that the user can see across tenants? 

Is access to the tenant info being controlled here, or is this just whether to render the tab. Can you point me to where the tenant api determines whether a user can see other tenant info?

packages/server/src/db/index.js
```diff
@@ -82,19 +82,22 @@ async function getUser(id) {
...
             'agencies.main_agency_id as agency_main_agency_id',
             'agencies.warning_threshold as agency_warning_threshold',
             'agencies.danger_threshold as agency_danger_threshold',
-            'users.tags',
+            'tenants.id as tenant_id',
+            'tenants.display_name as tenant_name',
         )
         .leftJoin('roles', 'roles.id', 'users.role_id')
         .leftJoin('agencies', 'agencies.id', 'users.agency_id')
+        .leftJoin('tenants', 'tenants.main_agency_id', 'users.agency_id')
```
bspates, line 100: I remember @chrisxkeith talking about back filling tenant ids, does that need to happen before this change goes out? Does work on this https://github.com/usdigitalresponse/usdr-gost/issues/72 need to be finished before rolling out dependencies on tenant id?

bspates (commented):
A few comments/questions for my understanding. Looks like you have a lint issue.

packages/client/src/components/Layout.vue
mhuang-usdr: There is no place (yet) in the tenant API that determines whether a user can see other tenant info. Currently it's the frontend that determines what users can see. The only users who should be able to see other tenants are tenants that belong to USDR (`tenant.id === 1`). 

I like the idea of adding a flag for readability and in case this functionality is extended in the future.

packages/server/src/db/index.js
mhuang-usdr: The current state of things is that there already are dependencies on tenant id, and they were added in before we could get the database to match. Based on what @chrisxkeith has shown me all the null tenants have already been filled in, so this change should be alright now.

packages/client/src/components/Layout.vue
mattbroussard: yeah I would at minimum define a constant somewhere like `USDR_TENANT_ID` and reference that here. For server-side permissions, I suggested [over here](https://github.com/usdigitalresponse/usdr-gost/pull/189#discussion_r950583560) making a `requireUSDRAdminUser` middleware to enforce this for routes that read or write across tenant boundaries.

packages/server/src/db/index.js
```diff
@@ -82,20 +82,22 @@ async function getUser(id) {
...
             'agencies.warning_threshold as agency_warning_threshold',
             'agencies.danger_threshold as agency_danger_threshold',
-            'users.tags',
-            'users.tenant_id',
+            'tenants.id as tenant_id',
+            'tenants.display_name as tenant_name',
         )
         .leftJoin('roles', 'roles.id', 'users.role_id')
         .leftJoin('agencies', 'agencies.id', 'users.agency_id')
+        .leftJoin('tenants', 'tenants.main_agency_id', 'users.agency_id')
```
mattbroussard, line 100: This join is wrong. We should join based on `users.tenant_id = tenants.id` not the `main_agency_id`. Current user could be from a subagency that is not the `main_agency_id` for any tenant.

We will need to ensure that all users/agencies with null `tenant_id` are filled in before deploying this change. I believe that is already the case on staging, perhaps not on prod yet based on discussion in GOSTbusters meeting today?

mattbroussard (changes requested):

packages/client/src/components/Layout.vue
mattbroussard: Actually, I think it might be good for the display of this to also gate on `loggedInUser.role.name == 'admin'`.

Right now it looks like there are two routes in routes/tenants.js:
- `GET /api/organizations/:organizationId/tenants`: only ever returns a single tenant, the one you belong to; and then only if you are an admin. That should change to use e.g. `requireUSDRAdminUser` and then return all tenants.
- `PUT /api/organizations/:organizationId/tenants/:tenant`: right now has no permission check other than `requireAdminUser`; should use `requireUSDRAdminUser`.

mattbroussard:
Closing because this has been sitting around for a few weeks and today I ended up making changes in #189 that accomplish the same goal.

PR #226, opened 2022-08-05
153 integrate upcoming closing dates

anigonzales:
Issue #153 

Issue Solution:
Integrate the query for the 3 grants with the closest close dates into the frontend, displaying grant title, close date, and interested agencies
Format the close date in MM/DD/YY and styled in accordance with the logged in user's agency thresholds (red - danger, yellow - warning)
Interested Agencies styled underneath grant title

Screenshots:
1.) Upcoming Closing Dates card on Dashboard with 3 grants with the closest closing dates. The close date is red in accordance with the danger threshold. The interested agencies are underneath the title.
Screen Shot 2022-08-05 at 1 03 44 PM

Testing:
Changed agency threshold to check if the close date would change colors correctly.
Checked in sql db the 3 closest grants if the correct grants were being displayed
Put interest in grants as different agencies to see they were styled and displayed correctly

Issue #153, opened 2022-07-22
(1)Integrate frontend template and backend API for the Upcoming Closing Dates


anigonzales:

mhuang-usdr:
QA: It looks like the changes from #225 (feedback from the UXD) got reverted.

- The colors need to be changed
- The dates should be bolded
- Looks like the agency list under the grant isn't showing up

Currently:
<img width="386" alt="image" src="https://user-images.githubusercontent.com/91221265/183788866-56305931-f284-4b09-b70a-d7b69c5d1ff9.png">


Used to be:
<img width="326" alt="image" src="https://user-images.githubusercontent.com/91221265/183788505-57b84680-aaf5-4f71-bd22-c1e3fa4bb946.png">


@anigonzales

anigonzales:
@mhuang-usdr QA changes addressed in #247

mhuang-usdr:
@anigonzales should this issue be closed?

PR #247, opened 2022-08-10
153 updated

anigonzales:
Issue #153 
#226  PR updated

Solution:
Added UX/UI to fit Upcoming Closing Grants table
Interested Agencies will only list under a grant if the grant has any, so having no abbreviations underneath means there are no agencies interested in that grant.
Changes do not include full page view

Screenshot:
1.) With the bold dates and font size changes

Screen Shot 2022-08-10 at 11 48 08 AM

Testing:
Added interest in a given grant and saw it display underneath the correct grant name.

### DIFF
--- a/packages/client/src/views/Dashboard.vue
+++ b/packages/client/src/views/Dashboard.vue
@@ -187,10 +187,13 @@ export default {
           thStyle: { width: '20%' },
           tdStyle: {
             color: 'red',
+            fontWeight: 'bold',
           },
           trStyle: {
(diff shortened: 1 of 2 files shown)


mhuang-usdr (changes requested):
Ah, I see. Any grants without interest shouldn't show on the upcoming closing dates feed. I'll create a new issue for that.

But for this PR, please still fix the colors of the dates. #225 was using accessible colors, please match it.

mhuang-usdr (approved):

PR #255, opened 2022-08-11
fixed bold

cavukovic:
### Description

fixed bold so that only the word assigned is bold

### DIFF
--- a/packages/client/src/views/Dashboard.vue
+++ b/packages/client/src/views/Dashboard.vue
@@ -30,7 +30,7 @@
                           <strong> interested </strong>
                       </span> in
                     </span>
-                    <span v-if="agencies.item.interested === 2" > <strong> was assigned </strong> </span>{{ agencies.item.grant }}
+                    <span v-if="agencies.item.interested === 2" >  was <strong> assigned </strong> </span>{{ agencies.item.grant }}
                   </div>
                 </template>
                 <template #cell(date)="dates">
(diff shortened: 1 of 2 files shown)


packages/client/src/views/Dashboard.vue
```diff
@@ -30,7 +30,7 @@
                           <strong> interested </strong>
                       </span> in
                     </span>
-                    <span v-if="agencies.item.interested === 2" > <strong> was assigned </strong> </span>{{ agencies.item.grant }}
+                    <span v-if="agencies.item.interested === 2" >  was <strong> assigned </strong> </span>{{ agencies.item.grant }}
```
mhuang-usdr, line 33: Looks like there's a double space between was and assigned.

packages/client/src/views/RecentActivity.vue
```diff
@@ -29,7 +29,7 @@
               <strong> interested </strong>
               </span> in
             </span>
-          <span v-if="agencies.item.interested === 2" > <strong> was assigned </strong> </span>{{ agencies.item.grant }}
+          <span v-if="agencies.item.interested === 2" > was <strong> assigned </strong> </span>{{ agencies.item.grant }}
```
mhuang-usdr, line 32: Same - Looks like there's a double space between was and assigned.

mhuang-usdr (changes requested):

packages/client/src/views/Dashboard.vue
cavukovic: updated it

packages/client/src/views/RecentActivity.vue
cavukovic: updated it

mhuang-usdr (approved):

PR #258, opened 2022-08-12
257 bug fix for upcoming closing dates

CGowan44:
### Description

There seemed to be an issue where the upcoming closing dates feed would not display anything if there were less than three grants in the table. This should give a fix for it.

### Screenshots / Demo Video

After fix:

image

Before fix:

image

### Testing

Make a new subagency and select a grant or two that would appear in the upcoming closing dates feed -> verify that they show up in the feed.

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/views/Dashboard.vue
+++ b/packages/client/src/views/Dashboard.vue
@@ -49,7 +49,7 @@
               <div class="card-block text-left">
                 <h4 class="card-title gutter-title2 row">Upcoming Closing Dates</h4>
               </div>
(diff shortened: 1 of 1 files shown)


cavukovic (approved):
curious if you even need the v-if on line 52, all that v-if is checking is if the length is > 0 but how would the table handle a grantsAndIntAgens array of length 0 if that v-if wasn't there, I don't think it hurts anything having it there though.

CGowan44:
> curious if you even need the v-if on line 52, all that v-if is checking is if the length is > 0 but how would the table handle a grantsAndIntAgens array of length 0 if that v-if wasn't there, I don't think it hurts anything having it there though.

Yep it seems to work fine without it.

Issue #250, opened 2022-08-10
Close Date Threshold colors not changing according to agency data.


anigonzales:
In #153 , dates in Upcoming Closing Dates should change to red if the date is in user's agency danger threshold and yellow if in warning. It is no longer changing accordingly.

PR #277, opened 2022-08-16
250 close date colors

anigonzales:
Issue #250 

Issue Solution:
- Changed handling of close_date values
- Specified date color matching to correct grant
- Re-worked logic for agency thresholds
- Red if in danger thresh, yellow if in warning thresh, black if outside (using accesible colors)

Screenshots:
1.) Card on Dashboard; Warning Thresh = 30 days, Danger Thresh = 15 days.
Screen Shot 2022-08-16 at 4 11 21 PM

2.) Full Page view of table with same thresholds as previous picture
Screen Shot 2022-08-16 at 4 11 44 PM

Testing:
Changed agency thresholds to see color changes.


cavukovic (approved):

packages/client/src/views/UpcomingClosingDates.vue
```diff
@@ -176,25 +179,38 @@ export default {
...
+      };
+      const daysTillDanger = days(dangerDate, new Date());
+      const daysTillWarn = days(warnDate, new Date());
+      const daysTillClose = days(new Date(value), new Date());
+      //                      ---assigning correct colors---
+      for (let i = 0; i < this.grantsAndIntAgens.length; i += 1) {
+        if ((daysTillClose <= warn) && (daysTillWarn > danger) && ((daysTillClose > danger) || (daysTillDanger <= daysTillClose))) {
+          this.yellowDate = true;
+          this.redDate = false;
+          this.blackDate = false;
```
mattbroussard, line 204: It's odd that these fields live on the top level component data since they pertain to a single item in the list -- this feels fragile because correct rendering depends on the ordering of rendering each item vs. the formatter being called -- an implementation detail in BootstrapVue's Table component that could easily change.

I think it would make more sense to compute the color in `formatUpcoming` and store it as a field on each member of `grantsAndIntAgens` that can be referenced in the cell template (e.g. `grantsAndIntAgens[index].redDate`).

packages/client/src/views/UpcomingClosingDates.vue
```diff
@@ -137,8 +140,8 @@ export default {
       await this.setup();
     },
     upcomingItems() {
-      // https://lukashermann.dev/writing/how-to-use-async-await-with-vuejs-components/
       this.formatUpcoming();
+      this.formatDate();
```
mattbroussard, line 144: Is this needed? `formatDate` expects a parameter and appears intended to be called from BootstrapVue's Table component as a cell formatter.

mattbroussard (commented):
Sorry for the late comments; had opened this in a tab yesterday but didn't get a chance to look at it before it got merged today.

Issue #195, opened 2022-08-02
[Performance Reporter] Add the URL redirect into the login email 


mhuang-usdr:
Right now, the login email will take you to the Grants ID tool by default. But if they came to the authentication page via the Performance Reporter, they should get redirected back to [grants.usdigitalresponse.org/#/](https://grants.usdigitalresponse.org/#/arpa-annual-performance-reporter)

ThienBNguyen:
Hi, I would like to work in this issue.

mhuang-usdr:
Hi @ThienBNguyen! Thanks for reaching out with your interest. This repository is part of [U.S. Digital Response](https://www.usdigitalresponse.org/), a nonprofit that organizes pro-bono technologists to help with critical government needs.

If you'd like to volunteer, please fill out our [volunteer handraiser form](https://www.usdigitalresponse.org/volunteer), so we can put you through the onboarding process. Thank you!

ThienBNguyen:
Thank you for the opportunity with US Digital Response as a volunteer. I can not join the organization.

mattbroussard:
As noted in my comment on #288, I'm taking this over since haven't heard from @romangurovich and I need this functionality for ARPA integration project.

PR #288, opened 2022-08-19
Support redirecting to specific application page after login (#195)

romangurovich:
### Ticket #
#195

### Description
In conjunction with some changes already added in #138, this adds support for redirecting after login. This is useful both for the Annual Reporter (already in GOST) and the ARPA Reporter app (soon to be integrated into GOST).

The flow is as follows:
- User attempts to access Annual Reporter page, but isn't logged in yet
- User is redirected to login page
- User attempts to login with email
- Server generates security token for user, and sends magic link via email
- User clicks magic link from email
- User is redirected directly to Annual Reporter (rather than to ID tool dashboard, as before

### Testing
Added a test covering validation of the redirect URLs, and that `/api/sessions/init` correctly serves a redirect.

### Misc
There is one drive-by fix here: in 33e38f9ec78a67fbdf508fc475f4f898776aff1f some duplicate code was added (having previously been removed in #96); here, I undid that change and fixed the test it was fixing in a more elegant way.


mattbroussard:
Taking over this PR since this functionality is needed for ARPA Integration and this seems to be a little stale (I slacked @romangurovich on 8/29 and haven't heard back so assuming they haven't been active recently).

packages/server/src/routes/sessions.js
```diff
@@ -16,12 +16,18 @@ const {
 // links. Now, we send login links through a clientside redirect instead so this should not be necessary.
 const MAX_ACCESS_TOKEN_USES = 1;
 
+function validatePostLoginRedirectPath(url) {
+    // TODO implement
+    return url !== null;
```
ajhyndman: What cases is this validation trying to prevent?  Is the idea to replace this functionality?

```
// Prevent them from accessing an undefined route, causes ugly blank screen failure
```

ajhyndman (approved):

packages/server/src/routes/sessions.js
```diff
@@ -17,8 +17,22 @@ const {
 const MAX_ACCESS_TOKEN_USES = 1;
 
 function validatePostLoginRedirectPath(url) {
```
ajhyndman: I guess another way to go about this would be with an allowlist for known routes?  That's just not very straightforward right now, because server and client codebases are separated.

packages/server/src/routes/sessions.js
mattbroussard: You seem to be commenting on one of my intermediate commits before the function was actually implemented; if you look at the [latest version](https://github.com/usdigitalresponse/usdr-gost/pull/288/files#diff-bd5fd4836f4c77fa6cd87a50e817574b0f7f3166dc12c3f34107db964d918c02) on the main PR page, there are some comments explaining. But basically, now that we are actually serving an HTTP redirect, it's important to not expose an [open redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html) that could be used for e.g. phishing links.

>Is the idea to replace this functionality?
>```
>// Prevent them from accessing an undefined route, causes ugly blank screen failure
>```

No, I removed that because it was redundant -- we already have a VueRouter config that redirects unknown routes back to the dashboard:
```
{
  path: '*',
  redirect: '/dashboard',
}
```

ajhyndman: Wow, you responded before I could amend.

Sorry!  I realized my mistake right after submitting.

packages/server/src/routes/sessions.js
mattbroussard: Yeah, I thought about that and didn't do it for that reason.

PR #292, opened 2022-08-22
ARPA integration: copy DB table schema from ARPA reporter

mattbroussard:
This adds a script `generate_arpa_table_migrations.js` that uses the `pg_dump` and `knex migrate:make` utilities to generate migrations that recreate certain ARPA reporter tables in GOST. This PR also includes the output generated by that script.

TODO:
 - Decide if we should rename any tables to make ARPA-specific use clearer?

In #296, I updated the GOST's `agencies` table to add the `code` field that ARPA's had. The other shared tables `user`, `access_tokens`, and `roles` do not require schema changes to support ARPA code using them.

See overall integration plan: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82


mattbroussard:
@ajhyndman not urgent, but this is ready to look at

ajhyndman (approved):
This looks great! I confirmed that I can run the migration generation and the migrations locally. The resulting tables look correct to me.

PR #295, opened 2022-08-23
Disable func-names linter

mattbroussard:
This was causing warnings on several PRs for things that are usually unnamed by convention and doesn't feel that useful anyway.


bspates (approved):
Sounds GTM

PR #296, opened 2022-08-24
[ID] ARPA integration: add agency code field to GOST agencies model

mattbroussard:
This adds `code`, a field that existed on ARPA's `agencies` table but not GOST's, to GOST's `agencies` table and correspondingly updates the add/edit agency flows on the client side and the routes/db accessors used by those. The new field is specified `NOT NULL` -- even though the ARPA field is not set this way, application code appears to expect it to be nonnull, so I went ahead and added the constraint here.

To minimize confusion for existing GOST/ID tool users, I've added some explanatory copy in the add/edit agency flows describing the differences, and also configured the form validator to allow you to leave `code` blank and default it to `abbreviation` (still nonnullable in Postgres).

image
image

Note: `code` is similar but not exactly the same as GOST's existing `abbreviation` field (though it's close enough that I used `abbreviation` for the backfill). `code` is part of ARPA Reporter spreadsheet templates, while `abbreviation` is purely used for display purposes in the ID tool. Before landing this, it might be worth confirming one against with @mhuang-usdr or @jcomeau1 that there's not a clean way to unify these concepts.

This has brought up a few questions for me:
- GOST/ID tool and ARPA reporter both have UI and routes for managing users and agencies that look and behave slightly differently. In a merged world, should we:
  - keep both around to reduce confusion for users that only use one of the tools
  - consolidate to one of them (probably GOST's? though ARPA's code is a little cleaner)
- How should permissions work for managing users/agencies in a merged world? In ARPA, an admin (of any agency) can see and modify all agencies in the tenant. But in GOST, admins can only see and modify agencies designated as a _subagency_ of their (a concept that did not exist in ARPA). Similar story for users: admins in GOST can only manage users that belong to their subagencies.
  - in practice, I suspect most admin users will exist on a top-level agency designated as a tenant's `main_agency_id` so this difference may be somewhat moot?
  
This change was originally going to be part of #292 but I decided to split it out because it's a bit more involved than the other db migrations.


bspates:
@mattbroussard 

> How should permissions work for managing users/agencies in a merged world? In ARPA, an admin (of any agency) can see and modify all agencies in the tenant. But in GOST, admins can only see and modify agencies designated as a subagency of their (a concept that did not exist in ARPA). Similar story for users: admins in GOST can only manage users that belong to their subagencies. 

This is actually being changed by https://github.com/usdigitalresponse/usdr-gost/issues/72. Only a select few places are still controlled by the agency -> subagency relationship, almost everywhere else admins will have tenant-wide access (and non admins will have visibility across their tenant as well).

ajhyndman (approved):
👍

PR #297, opened 2022-08-25
ARPA integration: script for declarative file copying + import rewrites

mattbroussard:
The main parts of this PR are:
 - A script that automates copying files from the arpa-reporter repo to the GOST repo based on a JSON configuration, and also rewrites imports in those scripts and adds footer comments referring to where and when the files came from.
 - A config for that script that I think should be pretty close to what we end up using. I created it by using `find` to list all files and directories in the arpa-reporter repo, then going through that list and thinking about what should happen to each folder.

There are several benefits to having this process automated:
 - **Declarative/Self-documenting/Reviewable:** we have a written documentation of what all is being copied (and not) that can be reviewed
 - **Repeatable:** We can experiment with different directory layouts, re-run the script after changes in the ARPA repo, etc.
   - (though right now it has no concept of deltas; it will copy everything. So we still need a "cutover" to the new repo at some point, unless I add more logic to deal with partial copies).
 - **Less tedious:** there are a lot of files and they aren't arranged in exactly the same way in both repos. I didn't want to copy/move things by hand
 - **Atomic:** it's hard to end up in any partial state (e.g. copied some files but not others; or copied files from different commits)
 - **Safer:** no chance of missing an import change and only discovering it when something blows up in the app itself

For review, I think the most pertinent thing to look at is the config JSON. Feel free to review the script itself, but its implementation details are not really the point here. The main files are:
- `arpa-to-gost.json5`: config file.
- `copy_with_rewrites.ts`: top level script, invoked with `yarn copier`
- `copier.ts`: logic for expanding the glob paths and copying files
- `footer_comments.ts`: logic for appending a comment to all copied files
- `rewrite_imports.ts`: logic for rewriting imports. It uses Babel's parser to generate an AST -- perhaps a little overengineered but I was worried about missing or mangling imports if I tried to do this with a regex.
- `undo_copies.ts`: script for undoing the actions of the main script to allow an edit cycle of run, test, modify config, rerun. It basically just deletes every file that was copied from the old repo and any directories created. This was also useful for testing the copier script itself. Invoked with `yarn undo:copier`.

I wrote all this in a separate Lerna "package" with somewhat different conventions/config than our other code because I wanted to use TypeScript and Prettier for developing this & didn't anticipate that others would be needing to modify it much (in fact, I'm not sure if it'll be useful past this project, but it's possible it could be if e.g. there are other USDR grants-related projects prototyped separately then brought into GOST). I would eventually like to get GOST+ARPA using TypeScript and Prettier in general, but that is a different project.

There are a few drive-by changes here that aren't directly related but were useful in testing atop this branch:
 - Add deps for `moment` and `ssf`, which ARPA Reporter uses but didn't have explicitly specified (so I didn't catch them when I wrote #302; but did when eslint complained about them). I guess these just happened to work because transitive dependencies brought them in?
 - Add `ignorePatterns` to entirely disable eslint for the paths being copied into. This is temporary to ease development/testing atop this branch since ARPA Reporter has tons of errors against GOST's linters.

After this script is run and the output committed, there are a few things that will have to be manually edited. I'm keeping a running patch of them in a [gist](https://gist.github.com/mattbroussard/7f725a9090ee5a9ff9be77eaab555e76) (it should be possible to download and `git apply` that file)

Overall design doc for ARPA integration: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82


mattbroussard:
@ajhyndman this is ready to look at now. I think the script is basically done and the config for it most of the way to its final state. My intent is to merge this and make smaller tweaks to the config in separate PRs. Then at some point when we're ready there will be a PR with the result of actually running the script.

packages/arpa_integration_tools/simple_test.json5
```diff
@@ -0,0 +1,21 @@
+{
+  "srcPath": "/Users/mattb/Desktop/usdr/arpa-reporter",
+  "srcRepoName": "arpa-reporter",
+  "destPath": "/Users/mattb/Desktop/usdr/usdr-gost",
+  "copies": {
+    "src/server/use-request.js": "packages/server/src/arpa-reporter",
+    "src/server/environment.js": "packages/server/src/arpa-reporter",
```
ajhyndman, line 7: I noticed that this test file uses the directory `arpa-reporter`, while `arpa-to-gost.json5` uses `arpa_reporter`.

Obviously only one of those will survive the real migration, but it might be nice to make these consistent just to avoid confusion.

packages/arpa_integration_tools/util.ts
```diff
@@ -0,0 +1,14 @@
+import * as path from "path";
+import * as fs from "fs/promises";
+
+export function truncateExtension(filePath: string): string {
+  const { dir, name } = path.parse(filePath);
+  return path.join(dir, name);
+}
```
ajhyndman, line 7: very clean 👍

ajhyndman (commented):
I don't feel strongly about folder structures, but I'm curious what options you considered so far.

Do you think something like the following could work?  I'm imagining a world in which we add a third tool to this repository and thinking through how we might keep things decoupled enough that it makes sense to keep seperate dev teams for each tool.

```
/packages
  /arpa-reporter
    /client
    /server
  /gost
    /client
    /server
/src
  /client
    App.vue
    main.js
  /server
    configure.js
    index.js
```

I realize that requires shuffling things a bit more in the existing gost code though.

packages/arpa_integration_tools/simple_test.json5
mattbroussard: Ah yeah. I might just delete this file anyhow.

ajhyndman:
I have successfully been able to run the import script, update `packages/client/vue.config.js` and see ARPA tool running!  That's pretty exciting.

trumanc (approved):

PR #299, opened 2022-08-26
ARPA integration: Database shims for ARPA Reporter code

mattbroussard:
These two files are intended as drop-in replacements for the ARPA Reporter's equivalent database accessor modules for the tables that will be shared. They will be designated as an import rewrite in my copy_and_rewrite script (see draft PR #297) The intent is to eventually remove these by updating callsites in ARPA Reporter to use the GOST database accessors directly.

This PR is a sister to https://github.com/usdigitalresponse/arpa-reporter/pull/505 in the arpa-reporter repo which modifies the usage of these tables there to more closely match their shape in GOST (particularly with respect to the `role`/`role_id` fields and joins to `agencies` in the `users` accessors.

This covers tables `users`, `roles`, and `agencies`. The `access_tokens` table will not be used directly in any ARPA reporter code (instead, ARPA reporter code will use GOST's `routes/sessions` and `access-helper` modules directly, which already have nearly identical API).

Overall design doc for integration project: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82

### DIFF
--- a/packages/server/src/db/arpa_reporter_db_shims/agencies.js
+++ b/packages/server/src/db/arpa_reporter_db_shims/agencies.js
@@ -1,0 +1,59 @@
+// This file exists as a shim to replace ARPA Reporter's db/users.js
+// Eventually, the functions in this file should have callsites updated to use GOST's existing methods
+// and this file can be deleted.
+
+const knex = require('../connection');
+const gostDb = require('..');
+
+// TODO(mbroussard): after merge, replace with an import to ARPA's use-request module
+function useTenantId() {
+    throw new Error('import missing -- need to update after ARPA merge');
+}
+function useUser() {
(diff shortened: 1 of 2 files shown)


ajhyndman (commented):
This looks great!  I confirmed that I can run the migration generation and the migrations locally.  The resulting tables look correct to me.

ajhyndman (approved):

mattbroussard:
>This looks great! I confirmed that I can run the migration generation and the migrations locally. The resulting tables look correct to me.

Did you mean to leave this comment on #292? It kind of sounds more relevant to that PR, since this PR doesn't contain any migrations.

ajhyndman:
Oh.  Yes I did!

I copied the review over there.  Sorry about that.

PR #302, opened 2022-08-27
ARPA integration: sync yarn/npm dependency versions

mattbroussard:
In preparation to begin moving code from arpa-reporter into GOST, this reconciles dependency versions. Where ARPA has a dependency that GOST doesn't, it's now been added in GOST (this does mean there are now currently some declared dependencies that are not yet used). Where ARPA had a newer version, I've upgraded that dependency in GOST. A few ARPA deps that seemed unused there are omitted (namely `fs-extra` comes to mind).

Notable changes to get things working:
- [Upgrade of Vue CLI from 4.x to 5.x](https://cli.vuejs.org/migrations/migrate-from-v4.html) (which in turn upgraded Webpack from 4.x to 5.x and Webpack Dev Server from 3.x to 4.x)
  - They added some lint rules that currently fail in GOST
    - I looked at trying to fix the `no-mutating-props` one but decided to punt on it because it requires substantive changes to how several components interact and I'm not really a Vue expert. 
    - I just disabled `multi-word-component-names` because it seems silly; the single-word component names we have are not likely to conflict with any built-in HTML tags, etc.
  - Webpack no longer automatically polyfills some node packages; the only place this was a problem was in `dotenv`, which it turns out we don't need on the client anyway (I'd previously [commented](https://github.com/usdigitalresponse/usdr-gost/pull/105#discussion_r906571851) about this where it was added)
  - Format of Webpack Dev Server config options [changed](https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md) slightly

For any other packages that got upgraded by a major version, I went and looked at their changelogs but didn't find anything relevant for us. I manually tested clicking around in the GOST app a bit, as well as running existing tests and linters. I have not yet tested ARPA with deps where GOST had a newer version already, but none of these were newer by major version (just minor/patch), so I suspect there won't be any problems.

There [used to be](https://github.com/usdigitalresponse/arpa-reporter/blob/01c0f639d5d9a2596efbc31003d91d4712935574/public/index.html#L13-L16) some dependencies directly referenced from CDNs in ARPA's `index.html`, but it doesn't look like there are anymore.

Overall ARPA integration design doc: https://www.notion.so/usdr/GOST-ARPA-integration-technical-plan-cffc5800c0414660816a0920b959da82


packages/client/package.json
```diff
@@ -14,40 +14,55 @@
...
     "vuex": "^3.6.2"
   },
   "devDependencies": {
-    "@vue/cli": "~4.5.0",
-    "@vue/cli-plugin-babel": "~4.5.0",
+    "@typescript-eslint/eslint-plugin": "^5.26.0",
+    "@typescript-eslint/parser": "^5.26.0",
+    "@vue/cli": "^4.5.13",
+    "@vue/cli-plugin-babel": "~5.0.4",
     "@vue/cli-plugin-e2e-nightwatch": "~4.5.0",
```
mattbroussard, line 39: note to self: should probably bump version on `@vue/cli` and these other plugins to match `@vue/cli-service`

ajhyndman (approved):

PR #303, opened 2022-08-29
Treat Postgres `date` fields as strings rather than JS Date objects

### LINKED ISSUE #77: (2) [ID] Posted and Closing dates are off by one

mattbroussard:
Fixes #77 -- My comment there goes into more detail about why this is needed, but tldr: by calling the `Date` constructor by default, we were accidentally assuming grant dates are in server's timezone (on Render, this seems to be UTC), causing off-by-one-day display bugs in the UI (users in the US will always see "midnight UTC" as the previous date).

The places that cared about these fields already were passing them into `Date` constructor, so no changes other than flipping the setting are needed.

Basically equivalent to https://github.com/usdigitalresponse/arpa-reporter/pull/440


ajhyndman (approved):
👍

PR #305, opened 2022-08-30
Fix peer dependency warnings, remove unused Jest deps, and force Yarn usage

mattbroussard:
We don't seem to actually be using Jest, so remove dependencies related to it.

The original reason I started looking at this is because @chrisxkeith [got](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1661890546721169) an error on `npm install`:
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: @vue/cli-plugin-unit-jest@5.0.8
npm ERR! Found: @vue/vue2-jest@28.1.0
npm ERR! node_modules/@vue/vue2-jest
npm ERR!   dev @vue/vue2-jest@"^28.0.1" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peerOptional @vue/vue2-jest@"^27.0.0-alpha.3" from @vue/cli-plugin-unit-jest@5.0.8
npm ERR! node_modules/@vue/cli-plugin-unit-jest
npm ERR!   dev @vue/cli-plugin-unit-jest@"~5.0.1" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: @vue/vue2-jest@27.0.0
npm ERR! node_modules/@vue/vue2-jest
npm ERR!   peerOptional @vue/vue2-jest@"^27.0.0-alpha.3" from @vue/cli-plugin-unit-jest@5.0.8
npm ERR!   node_modules/@vue/cli-plugin-unit-jest
npm ERR!     dev @vue/cli-plugin-unit-jest@"~5.0.1" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

It seems that Yarn ignores this conflicting peer dependency because it's marked "optional", but npm is stricter. We should be using yarn anyway, but good not to have conflicting dependencies regardless (though the package in question [does not use semver in the traditional way](https://github.com/vuejs/vue-jest#readme) anyway). Since we don't seem to be using Jest, it didn't seem like there was any reason to e.g. keep this but adjust the version.

Using git blame:
- The `"@vue/vue2-jest": "^28.0.1"` dependency was added in 3926582cf6e644c6f5ef029653afba828843c9b0, a merge commit. I can't really see any reason why but that commit also contains a `package-lock.json` that no longer exists so I suspect it relates to someone accidentally using npm instead of yarn.
- The `"@vue/cli-plugin-unit-jest"` dependency was added in 68f24617a31f231e30c5e17245899ac21463989e. In that commit there is also a `jest.config.js`, so I suspect we did at one time use Jest and no longer do (now using Mocha). I don't see anything that would invoke Jest, so I'm removing this too.

In this PR, I've also added config that should force using yarn if someone accidentally types `npm install` out of habit.

Finally, I added a few dev dependencies to satisfy warnings about unmet peer dependencies -- these didn't seem to be causing a problem (I think the packages in question also have their own dep that they use if you don't provide a peer?) but the warnings were a little annoying; `yarn install` should be a bit quieter. I removed a few deps related to E2E testing and Chromedriver/Geckodriver since we don't have any such tests today.


package.json
```diff
@@ -35,9 +37,9 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "@commitlint/cli": "^11.0.0",
-    "@commitlint/config-conventional": "^11.0.0",
-    "@commitlint/config-lerna-scopes": "^11.0.0",
+    "@commitlint/cli": "^17.0.0",
```
mattbroussard, line 40: This bump was needed because older version had a peer dependency on a lower version of Lerna. There don't seem to be any breaking changes.

chrisxkeith (approved):

PR #306, opened 2022-09-02
ARPA integration: Configure Express and Vue for multiple entry points (+ simplify API_DOMAIN vs WEBSITE_DOMAIN)

mattbroussard:
In preparation for integration of ARPA Reporter code into GOST, this makes several changes:
 - Adds Vue CLI and webpack-dev-server setup for multiple entry points/pages
 - Adds Express configuration to allow `connect-history-api-fallback` to work correctly for ARPA Reporter, which uses HTML5 History API instead of URL hashes for SPA navigation
 - Reorganizes code in configure.js slightly (and add comments) to make it easier to register ARPA Reporter routes (similar done in sister PR: https://github.com/usdigitalresponse/arpa-reporter/pull/510)
 - Adds a `VUE_APP_IS_GOST` environment variable that can be used to gate behaviors in ARPA Reporter client code (one such usage in https://github.com/usdigitalresponse/arpa-reporter/pull/510)
 - Consolidates the two "static" folders and remove the unused static route
 - Simplifies the `API_DOMAIN` vs `WEBSITE_DOMAIN` distinction to be more similar to what ARPA Reporter had already (this isn't strictly necessary but has been a recent point of [confusion](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1661975156584709?thread_ts=1661962811.434859&cid=C0324KDQSCR) in GOST too)
   - webpack-dev-server is now configured to proxy all `/api` requests so there's no need for XHRs and login links to go to :3000 in dev; everything can use :8080
   - There are no longer `API_DOMAIN` or `VUE_APP_GRANTS_API_URL` variables
   - Because everything's served on the same domain in dev now, the CORS middleware is no longer needed (its previous configuration was potentially insecure anyway: it allowed all routes to be accessed from all origins!)

This PR also contains one "drive-by" fix:
 - I noticed the server crashing in a cron job if the gitignored `static/forms/generated` folder was missing. I added some error handling around this so the server doesn't crash in this case.


packages/client/src/helpers/fetchApi.js
```diff
@@ -7,7 +7,7 @@ function getDefaultHeaders() {
 }
 
 export function addOrganizationId(url) {
-  return `${process.env.VUE_APP_GRANTS_API_URL}${url.replace(':organizationId', store.getters['users/selectedAgencyId'])}`;
+  return url.replace(':organizationId', store.getters['users/selectedAgencyId']);
```
chrisxkeith, line 10: Is 'store' a way to get cookies? Or some other form of storage in the browser?

packages/server/__tests__/api/utils.js
```diff
@@ -2,10 +2,15 @@ const fetch = require('node-fetch');
 require('dotenv').config();
 const { knex } = require('../../src/db');
 
+function getTestDomain() {
+    const { PORT = 3000 } = process.env;
```
chrisxkeith, line 6: Is this equivalent to `if process.env.PORT) { PORT = process.env.PORT } else { PORT = 3000 )` ?

packages/client/src/helpers/fetchApi.js
mattbroussard: We use the [VueX](https://v3.vuex.vuejs.org/) state management library. `store` refers to the VueX store, which manages clientside application state. Most of it is not persisted and gets refetched from the server on each page load, but there are a few parts that get stored in `localStorage` as well. The getter referenced here is defined [here](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/client/src/store/modules/users.js#L21).

This `organizationId` param in API URLs is used to support functionality in GOST where admins can impersonate any of their subagencies. This helper makes it so that every place that does XHR doesn't have to worry about which agency is being impersonated.

packages/server/__tests__/api/utils.js
mattbroussard: Yep. This syntax is called [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) with a default value.

ajhyndman (approved):
Looks good to me!

Issue #129, opened 2022-07-15
[ID] Interested and Rejected Grants views should be Agency-scoped, not Agency+child-scoped


mhuang-usdr:
Video explaining the issue here: https://www.loom.com/share/b98a1d6cd01e46b7a862c264ff07e9d2

mhuang-usdr:
Initial eng context - the Interested Grants table only tracks the user_id

bspates:
@mhuang-usdr Mind if I try this one out?

mhuang-usdr:
It's yours! Once I get you added to this repo you can assign the issue to yourself.

mhuang-usdr:
@bspates I can still see Interested, Rejected, and Assigned grants that are not for the agency I'm currently viewing as. I recorded a video to explain it a bit better. This is on `_staging`

https://www.loom.com/share/7370f6804fc14c59b590da603250e93e

bspates:
@mhuang-usdr I just tried this on staging. Looks like it is scoped to your current agency and any subagencies. This becomes more obvious when switching to an agency with no subs.

mhuang-usdr:
@bspates ah, it should be scoped to only your current agency, not any subagencies.

mhuang-usdr:
This issue is fixed for Interested grants but not for Rejected grants. For Rejected grants, I am able to see all grants that anyone has rejected.

as1729:
@mhuang-usdr I added an MR here that should resolve the issue you may be seeing. Let me know if the[ loom demo](https://www.loom.com/share/21ea1fd0f1be491aaf7cd8be8bfacbb9) looks right.

PR #309, opened 2022-09-06
Tenant scoping apis

bspates:
### Ticket #72 
Also fixes: 
* #129
* #61  

### Description
Re-scopes access on api endpoints from agency hierarchy to the rules specified in #72 

### Screenshots / Demo Video
Did bug bash with Mindy

### Testing
- Updated existing api/db unit tests to match new logic
- Manually tested UI use cases

### Checklist
- [x] Added PR reviewers


packages/server/src/routes/agencies.js
```diff
@@ -16,20 +17,36 @@ const {
...
-    let response;
-    if (user.role.name === 'admin') {
-        response = await getAgencies(req.session.selectedAgency);
-    } else {
-        response = await getAgency(req.session.selectedAgency);
+    const response = await getTenantAgencies(user.tenant_id);
+    res.json(response);
+});
+
+router.get('/impersonable', requireAdminUser, async (req, res) => {
```
mattbroussard, line 24: I'm confused about this endpoint:
1. It's not called anywhere? What is its intended use?
2. I thought based on the changes in access-helpers that we are now allowing impersonation of any agency in the tenant -- but this still uses `getAgencyTree` to return only subagencies of a user's agency.
3. What is the difference between `asAgency` and `organizationId` (aka `selectedAgency`)?

packages/server/src/routes/agencies.js
```diff
@@ -16,20 +17,36 @@ const {
...
+        if (!allowed) {
+            return res.status(403);
+        }
+        agencyId = asAgency;
     }
+    const response = await getAgencyTree(agencyId);
     res.json(response);
 });
 
 router.put('/:agency', requireAdminUser, async (req, res) => {
```
mattbroussard, line 40: nit: rename the path parameter to `:agencyId` to make its downstream use clearer. Same for the other routes in this file that have a `:agency` param.

packages/server/src/routes/agencies.js
```diff
@@ -87,7 +113,8 @@ router.put('/parent/:agency', requireAdminUser, async (req, res) => {
 
 router.post('/', requireAdminUser, async (req, res) => {
     const { user } = req.session;
-    if (!isPartOfAgency(user.agency.subagencies, req.body.parentId)) {
+    const allowed = await isUserAuthorized(user, req.body.parentId);
+    if (!allowed) {
         throw new Error(`You dont have access parent agency`);
```
mattbroussard, line 118: nit: send a 403 instead of throwing, which will send a 500.

packages/server/src/lib/access-helpers.js
```diff
@@ -1,19 +1,19 @@
...
  * Determine if a user is authorized for an agency.
  *
- * @param {Number} userId
+ * @param {Object} user
  * @param {Number} agencyId
- * @returns {Boolean} true if the agency is the user's or a descendant; false otherwise
- */
+ * @returns {Boolean} true if the agency is in the same tenant as the user
+ * */
+async function isUserAuthorized(user, agencyId) {
```
mattbroussard, line 10: Naming of the parameter here and the way it's passed through to `inTenant` doesn't make it clear that it can be a single ID or a list -- as noted elsewhere, would prefer to just make it always a list and name it `agencyIds`.

Also: I'd consider renaming this function to be a bit more explicit (e.g. `isUserAuthorizedForAgency`) since it's now used in many more places outside this file where the context is less obvious about what authorization means.

packages/server/src/routes/tenants.js
```diff
@@ -6,14 +6,8 @@ const {
...
 
-router.get('/', requireUser, async (req, res) => {
-    const { user } = req.session;
-    let response;
-    if (user.role.name === 'admin') {
-        response = await getTenant(req.session.selectedAgency);
-    } else {
-        throw new Error(`You dont have access to tenants`);
-    }
+router.get('/', requireAdminUser, async (req, res) => {
```
mattbroussard, line 9: Note: will merge conflict with my changes in #189

packages/server/src/routes/users.js
```diff
@@ -1,29 +1,39 @@
...
+    const { user } = req.session;
     if (!req.body.email) {
         res.status(400).send('User email is required');
         return;
     }
 
+    let agencyId = user.agency_id;
+    if (Number.isFinite(req.body.agency)) {
+        agencyId = req.body.agency;
+    }
```
mattbroussard, line 18: Should we just require `agency` in the body and not have this fallback that defaults to creating a user within creator's own agency? Client UI could handle setting that as a default value.

packages/server/__tests__/api/agencies.test.js
```diff
@@ -36,20 +36,34 @@ describe('`/api/organizations/:organizationId/agencies` endpoint', () => {
...
     });
 
-    context('GET organizations/:organizationId/agencies (list an agency and its subagencies)', () => {
+    context('GET organizations/:organizationId/agencies', () => {
+        it('lists all agencies within a tenant', async () => {
+            // Will default to user's own agency ID
+            const response = await fetchApi('/agencies', agencies.admin.own, fetchOptions.admin);
+            expect(response.statusText).to.equal('OK');
+            const json = await response.json();
+            expect(json.length).to.equal(4);
```
mattbroussard, line 45: I think it's worth asserting on what the specific agency IDs in the response are. Same applies to the other tests added in this file.

packages/server/__tests__/api/grants.test.js
```diff
@@ -312,8 +312,8 @@ describe('`/api/grants` endpoint', () => {
             it('includes this user\'s own agency when it is interested in the grant', async () => {
                 expect(json.some((r) => r.agency_id === agencies.own)).to.equal(true);
             });
-            it('excludes a subagency of this user\'s own agency when the subagency is interested in the grant', async () => {
-                expect(json.find((r) => r.agency_id === agencies.ownSub)).to.be.undefined;
+            it('excludes an agency outside  of this user\'s own tenant', async () => {
+                expect(json.find((r) => r.agency_id === 0)).to.be.undefined;
```
mattbroussard, line 316: Is it guaranteed at this point that agency 0 is interested in some grant so we know that undefined here means it was actually filtered out vs. it just doesn't exist?

packages/server/__tests__/db/db.test.js
```diff
@@ -20,23 +20,22 @@ describe('db', () => {
...
             // in the grants interested table the grant with the most recent created_at has the grant id of 335255
-            expect(result.rows[0]).to.have.property('grant_id').equal('335255');
+            expect(rows[0]).to.have.property('grant_id').equal('335255');
         });
         it('gets the two most recent interested grants', async () => {
             // testing pagination
-            const result = await db.getGrantsInterested({ perPage: 2, currentPage: 1 });
-            expect(result.rows).to.have.lengthOf(2);
+            const rows = await db.getGrantsInterested({ agencyId: fixtures.users.staffUser.agency_id, perPage: 2, currentPage: 1 });
+            expect(rows).to.have.lengthOf(2);
```
mattbroussard, line 32: Can these tests that look at rows.length check the actual IDs of grants returned?

mattbroussard (approved):
Mostly minor comments but some questions mainly about:
1. What is the `/api/organizations/:organizationId/agencies/impersonable` endpoint?
2. What are the new semantics for impersonation / is the query in `inTenant` correct?

There may be some merge conflicts with #189, though probably less now that this PR does not include any "super-admin" logic (for which I added some logic over there). Let me know if you'd like to discuss synchronously (ideally today or Monday; I am traveling 9/13-20).

Marking approved to unblock, but please make requested changes before merge & re-request review if things change substantially.

packages/server/src/db/index.js
```diff
@@ -498,29 +495,46 @@ async function markGrantAsInterested({
...
+        .orderBy('created_at', 'DESC')
+        .offset((currentPage - 1) * perPage)
+        .limit(perPage);
 }
 
-async function getTotalInterestedGrants() {
+async function getTotalInterestedGrants(agencyId) {
     const rows = await knex(TABLES.grants_interested)
         .whereNot('interested_code_id', null)
+        .andWhere('agency_id', agencyId)
```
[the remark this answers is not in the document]
bspates: 🙃

packages/server/src/db/index.js
```diff
@@ -767,6 +795,25 @@ async function sync(tableName, syncKey, updateCols, newRows) {
...
+        .leftJoin('agencies', 'users.tenant_id', 'agencies.tenant_id')
+        .leftJoin('tenants', 'tenants.id', 'users.tenant_id')
+        .where('users.id', userId)
+        .andWhere('users.tenant_id', tenantId);
+
+    if (Array.isArray(agencyId)) {
+        q.andWhere((qb) => qb.whereIn('agencies.id', agencyId));
+    } else {
+        q.andWhere('agencies.id', agencyId);
+    }
```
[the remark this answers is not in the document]
bspates: There is actually a place that is calling isUserAuthorized outside of the access-helper that was using list of agency ids and couldn't be converted to just using tenant id. https://github.com/usdigitalresponse/usdr-gost/blob/tenant-scoping-apis2/packages/server/src/routes/grants.js#L185 

I'll just make it always an array param, and handle the distinction at the isUserAuthorized level.

packages/server/src/routes/users.js
bspates: Since both cases were covered in the existing unit tests it seemed better to follow the pattern than change it and increase the scope of the PR to more front end changes.

packages/server/src/db/index.js
```diff
@@ -549,26 +563,40 @@ async function getAgency(agencyId) {
     return result.rows;
 }
 
-async function getAgencies(rootAgency) {
+async function getTenantAgencies(tenantId) {
+    return knex(TABLES.agencies)
```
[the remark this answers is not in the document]
bspates: This is now changed below this comment.

packages/server/__tests__/api/grants.test.js
bspates: Yeah it is setup in the db seed run before the tests https://github.com/usdigitalresponse/usdr-gost/blob/tenant-scoping-apis2/packages/server/seeds/dev/ref/grants.js#L119

packages/server/src/lib/access-helpers.js
```diff
@@ -23,6 +23,11 @@ async function requireAdminUser(req, res, next) {
     }
 
     const user = await getUser(req.signedCookies.userId);
+    if (!user) {
```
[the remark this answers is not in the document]
bspates: Seems like it

packages/server/src/lib/access-helpers.js
```diff
@@ -1,19 +1,19 @@
...
  *
- * @param {Number} userId
+ * @param {Object} user
  * @param {Number} agencyId
- * @returns {Boolean} true if the agency is the user's or a descendant; false otherwise
- */
+ * @returns {Boolean} true if the agency is in the same tenant as the user
+ * */
+async function isUserAuthorized(user, agencyId) {
+    return inTenant(user.id, user.tenant_id, agencyId);
```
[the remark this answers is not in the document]
bspates: Good point, I think I missed this nuance after removing the super tenant concept.

packages/server/src/routes/agencies.js
bspates: So originally I had front-end changes to use this endpoint for the only purpose of selecting another agency to impersonate. I decided to move them to a later PR when taking out the super-tenant stuff and they were a little scrambled together. Should have mentioned that somewhere. I can remove these backend changes from this PR.

packages/server/src/lib/access-helpers.js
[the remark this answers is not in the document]
mattbroussard: Huh? So what are the semantics _supposed to be_?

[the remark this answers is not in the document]
bspates: Impersonation is limited to children in your agency tree, which is important when checking the selectedAgency, not just when choosing an agency to impersonate. In almost all other situations users can see objects across all agencies in a tenant. Like the user listing api is not limited by membership in a child agency anymore. So I'll be add a different function that checks for allowable impersonation vs allowing access to a resource... OR ask @mhuang-usdr If its okay if we let admin users (which are the only ones that can impersonate) impersonate any agency in a tenant, to help simplify things.

packages/server/src/db/index.js
```diff
@@ -767,6 +795,25 @@ async function sync(tableName, syncKey, updateCols, newRows) {
     }
 }
 
+async function inTenant(userId, tenantId, agencyId) {
+    const q = knex(TABLES.users)
+        .select('users.id as user_id', 'agencies.id as agency_id')
+        .leftJoin('agencies', 'users.tenant_id', 'agencies.tenant_id')
+        .leftJoin('tenants', 'tenants.id', 'users.tenant_id')
+        .where('users.id', userId)
+        .andWhere('users.tenant_id', tenantId);
```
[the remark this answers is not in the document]
mattbroussard: Your comment after the updates says this function

> Determines if a user is in an agency and that agency is in the same tenant

However, I'm not sure the query is quite right after playing with it in Postico a bit: it checks that the agency and user both match the specified tenant ID, and but not that the user belongs to that agency. I think you need to add a clause like `.where('users.agency_id', knex.ref('agencies.id'))`. That will ensure you only get rows if the user is part of the one of the specified agencies.

packages/server/src/db/index.js
```diff
@@ -767,6 +795,34 @@ async function sync(tableName, syncKey, updateCols, newRows) {
...
+
+    if (agencyIds.length > 1) {
+        q.andWhere((qb) => qb.whereIn('agencies.id', agencyIds));
+    } else if (agencyIds.length === 1) {
+        q.andWhere('agencies.id', agencyIds[0]);
+    } else {
+        return false;
+    }
```
mattbroussard, line 819: nits on this section:
 - no need to gate on one vs. multiple IDs; a `whereIn` with one item is just fine
 - I don't think you need the querybuilder callback, doesn't `q.whereIn(...)` work?
 - I think empty `agencyIds` ought to throw instead of just return false; it seems like that would always be a programming mistake?

packages/server/src/lib/access-helpers.js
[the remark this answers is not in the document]
mattbroussard: Got it. I suspect the answer is we want impersonation to still be subagencies only.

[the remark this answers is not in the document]
mhuang-usdr: Clarifying question on what "allowable impersonation vs allowing access to a resource" means -

Let's say we have this agency tree:

- main agency
   -  A
      - sub-A
   - B

So,  if an admin in Agency A is impersonates sub-A, they should have full write/edit access to sub-A.

**Question:**
- in this current PR, if an admin in Agency A is impersonates Agency B, what is the experience? Is it a read-only experience, or does Agency A have write access?

From the product perspective, read-only impersonation should be fine - what's important is that the user is not able to write to an agency they're not supposed to - so I'd lean towards the simplest engineering solution until we hear future feedback.

However, I'd like to test the user experience of such an impersonation to make sure it's not totally broken!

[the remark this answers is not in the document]
mhuang-usdr: @bspates can I check via the same ngrok link you sent earlier?

packages/server/src/db/index.js
```diff
@@ -116,7 +118,7 @@ async function getUser(id) {
         };
         let subagencies = [];
         if (user.role.name === 'admin') {
-            subagencies = await getAgencies(user.agency_id);
+            subagencies = await getAgencyTree(user.agency_id);
```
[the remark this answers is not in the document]
bspates: It works with any agency ID, which for a recursive tree function it makes sense to refer to any node as a possible "root".

packages/server/src/lib/access-helpers.js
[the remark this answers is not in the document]
bspates: Talked with @mhuang-usdr, the answer is that write access in all cases needs to be limited by the agency tree. The nuance here is that with multi-tenancy we have admin users able to see across agency trees in a tenant, BUT not edit them. This PR makes it so that admins can write/edit any object in the tenancy. The work to move to a read but not write across the tenancy is not small, so we're spinning that out to another issue, as this is more of a future use case than a current one.

packages/server/src/db/index.js
bspates: I didn't realize where and "andWhere" were aliases, so I thought I had to do an an andWhere => whereIn, thanks! An in clause can be a much slower query due to not activating the index or forcing a file sort, sometimes the query engine will optimize it away for a single record sometimes not.

packages/server/src/db/index.js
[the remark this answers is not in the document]
bspates: I think I need to change the comment, its actually whether they have access to said agency, not whether they are apart of it, so its just that both the user and agency are in the same tenant.

packages/server/__tests__/db/db.test.js
bspates: I changed this one to check the date ordering, so the created_at values, since they reference the same grant id. The tests below are actually stats endpoints so they can only verify the count as its all that is returned by the api.

packages/server/src/routes/agencies.js
bspates: I'd rather not grow the scope of this PR with a bunch of consistency changes that are currently outside of the diff. I'll make another issue for it. The repo in general has a lot of inconsistencies. I've tried to just leave the code in one place consistent with itself.

bspates: https://github.com/usdigitalresponse/usdr-gost/issues/328

packages/server/seeds/dev/ref/agencies.js
```diff
@@ -8,6 +8,33 @@ module.exports = [
         main_agency_id: 0,
         tenant_id: 1,
     },
+    {
+        id: 400,
+        abbreviation: 'TSDR',
+        code: 'TSDR',
```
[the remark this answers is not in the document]
bspates: Having a _ would mean its not an abbreviation.

packages/server/src/lib/access-helpers.js
```diff
@@ -1,19 +1,19 @@
...
+}
+
 async function isAuthorized(userId, agencyId) {
     const user = await getUser(userId);
-    return isPartOfAgency(user.agency.subagencies, agencyId);
+    return isUserAuthorized(user, agencyId);
 }
```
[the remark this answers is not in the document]
bspates: Both are still in use. For the case when an api is trying to do something on behalf of a different user than the one in session.

PR #310, opened 2022-09-06
ARPA Integration: GOST-side changes to allow running ARPA Reporter tests

mattbroussard:
This will accompany changes in the ARPA repo to make the ARPA Reporter unit tests that will eventually be copied in by the script in #297 possible to run in GOST (and automatically run by CI).

#308 (not for review) is a branch that includes these changes, the copied files, plus some ARPA-side changes that will shortly be put up in a sister PR, validating that the tests can run and pass on CI.

Without those other changes, the new test script added here no-ops.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,9 @@ jobs:
       - name: Run unit tests
         env:
           POSTGRES_TEST_URL: "postgresql://postgres:password@localhost:${{ job.services.postgres.ports[5432] }}/usdr_grants_test"
+          # This is intentional to set POSTGRES_URL=POSTGRES_TEST_URL; ARPA Reporter test runner gates
+          # dev vs. CI differences based on whether these are the same.
+          POSTGRES_URL: "postgresql://postgres:password@localhost:${{ job.services.postgres.ports[5432] }}/usdr_grants_test"
           AWS_ACCESS_KEY_ID: "Fake AWS Key"
           AWS_SECRET_ACCESS_KEY: "Fake AWS Secret"
           NOTIFICATIONS_EMAIL: grants-identification@usdigitalresponse.org
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,5 +53,4 @@ jobs:
           # The .env file needs to be present; the example file is good enough.
           cp packages/server/.env.example packages/server/.env
           cp packages/client/.env.example packages/client/.env
-          cd packages/server/; yarn test:apis
           export CI=true; yarn test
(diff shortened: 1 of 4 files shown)


packages/client/package.json
```diff
@@ -11,7 +11,7 @@
     "start": "yarn run serve",
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
-    "test": "vue-cli-service test:unit --passWithNoTests",
```
mattbroussard, line 14: This argument belongs to Jest, not Mocha, and seems to be an vestige of a time when GOST used Jest (other references to such removed in #305). In Mocha, for some reason it prevented ARPA tests from running -- I think it got interpreted as a path that didn't match anything?

.github/workflows/ci.yaml
```diff
@@ -46,12 +46,14 @@ jobs:
...
+          # dev vs. CI differences based on whether these are the same.
+          POSTGRES_URL: "postgresql://postgres:password@localhost:${{ job.services.postgres.ports[5432] }}/usdr_grants_test"
           AWS_ACCESS_KEY_ID: "Fake AWS Key"
           AWS_SECRET_ACCESS_KEY: "Fake AWS Secret"
           NOTIFICATIONS_EMAIL: grants-identification@usdigitalresponse.org
         run: |
           # The .env file needs to be present; the example file is good enough.
           cp packages/server/.env.example packages/server/.env
           cp packages/client/.env.example packages/client/.env
-          cd packages/server/; yarn test:apis
```
mattbroussard, line 56: Removed because `yarn test` called on the next line amended to also call `yarn test:apis`.

packages/server/__tests__/run_arpa_reporter_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
ajhyndman, line 1: Have you got an action item tracked anywhere to clean this up post-merge?

ajhyndman (approved):
I keep forgetting to click "approve" the first time I submit a review.

packages/server/__tests__/run_arpa_reporter_tests.sh
mattbroussard: Not explicitly, though we don't technically need to do anything with this; I suppose it's a slight readability improvement to remove a level of indirection.

Eventually, it might be nice to have GOST and ARPA tests use the same Mocha instance, same seeds, etc. There are some other things to reconcile there, e.g. ARPA's route tests start a server in the same process and use Supertest to interact with it but GOST's start a server in a separate process. I like ARPA's approach better (I guess I would since I added it 😛) since:
1. It doesn't require you to stop your dev server to run tests (in GOST, they compete for port 3000)
2. You could theoretically mock/spy on things since it's in the same process (can't do this in GOST)
3. Easier to use a debugger with the test + tested code if they're in the same process

PR #312, opened 2022-09-07
ARPA Integration: Script to import ARPA reporter data dumps

mattbroussard:
This is a sister PR of https://github.com/usdigitalresponse/arpa-reporter/pull/516. The script implemented here ingests the ZIP file generated by the script in that PR and updates the database and filesystem accordingly.

When creating a new tenant, we ask the user whether they want to designate an existing (to-be-copied) `main_agency_id`, or whether to create a new root agency. Then all other agencies and users are by default children of that one root agency.

Tables are carefully processed in order of foreign key dependencies, and foreign keys are re-keyed based on the autoincrement IDs in the destination table. We also check for duplicate user emails (which have a `UNIQUE` constraint) and ask the user interactively to reassign email addresses for any duplicates (for Gmail addresses, the script will suggest a plus-address).

Everything is done in a transaction, so if any error occurs, the whole thing is rolled back. There is also a `dryRun` option that causes the transaction to rollback even if everything went OK.

Some questions:
- Right now we assign all users to the main agency (either new or copied). But I wonder if we should instead let copied users with non-null `user.agency_id` keep that agency association and only reassign users with null `agency_id`?
- When creating a new main agency, should we also create a new admin user for that agency? Right now all users get assigned to the main agency so as long as source has an admin, destination will; but from question above, there's a possibility of changing that? The new tenant flow implemented in #193 and #189 automatically creates a new main agency and new admin user, but that flow is starting from nothing -- the situation when migrating is a little different.


packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+            tenant.__unmappedFutureMainAgencyId = mainAgencyId;
+        } else {
+            const agencyToInsert = {
+                id: 1000000 + tenantId,
+                tenant_id: tenantId,
+                name: tenantName,
+                code: `CHANGEME_${tenant.id}`,
+                __isMainAgency: true,
+            };
+            dbContents.agencies.splice(0, 0, agencyToInsert);
```
trumanc, line 176: Can we print an output statement here mentioning that we're adding in a new agency, so that it's clear why the import counts don't match the export counts?

I guess it isn't super critical because the inquirer question lets them know we are creating a new agency, but might be less confusing to be explicit when we actually add it in the code

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+
+    return { inserted, idLookup };
+}
+
+async function importAgencies(
+    dbContents,
+    idLookupByTable,
+    insertedRowsByTable
+) {
+    // First, create all agencies, defaulting their parent pointer to point to themselves.
```
trumanc, line 194: nit: This comment should mention both the parent pointer and the main_agency pointer

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+        rekeyForeignKeys(
+            {
+                email: user.email,
+                name: user.name,
+                // Note: ARPA Reporter had a "reporter" role. For our purposes,
+                // anything non-admin becomes "staff"
+                role_id: user.role === 'admin' ? adminRole : staffRole,
+                tenant_id: user.tenant_id,
+                // Copied users belong to the main agency of their tenant
+                agency_id:
```
trumanc, line 269: In the ARPA tables admin users have an empty agency id, and reporter type users have specific ids set. Are there likely to be consequences for changing that for either/both of these user types?

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+    const usersWithDupeEmails = await knex("users")
+        .select("email")
+        .whereIn("email", _.map(usersToCreate, "email"));
+    if (usersWithDupeEmails.length != 0) {
+        console.error(
+            "Found duplicate emails",
+            _.map(usersWithDupeEmails, "email")
+        );
+        throw new Error(
+            "found users with duplicate emails! Delete them first, or update script to handle."
```
trumanc, line 287: There are definitely duplicates in the tables, especially for the USDR employees/volunteers that show up in most of the dbs, so we'll definitely need to address this before we can run it

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+        ];
+    }
+
+    return { idLookupByTable, insertedRowsByTable };
+}
+
+// based on function of the same name in ARPA's services/persist-upload module
+// with some adjustments. Ideally, we'd just import it, but this migrate script
+// is going to be committed before ARPA's code is copied in.
+function uploadFSName(upload) {
```
trumanc, line 344: Since these are being duplicated here, can you write some big caps lock warnings in the ARPA repo PR saying that logic cannot be modified, just to warn people off in case they haven't seen this?

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+    );
+
+    // Create tenants
+    let inserted = [];
+    for (const tenantId of allTenantIds) {
+        const tenantName = tenantNames[tenantId];
+        const mainAgencyId = mainAgencies[tenantId];
+
+        const tenant = await knex("tenants")
+            .insert({
```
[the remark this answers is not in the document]
mattbroussard: Good idea. I think probably the easiest way to do that would be to put all of the DB inserts in a transaction and make it so that the file copy portion of the script can be run separately.

packages/server/src/scripts/import_arpa_reporter_dump.js
mattbroussard: Oh, good point -- I didn't think about our own users on partners' instances.

We can probably have the script prompt for a remapped email and use a plus-address (at least for usdigitalresponse.org and other Gmail addresses). Or we could omit such users, but there might be rows in other tables that refer to them (e.g. `validated_by`).

packages/server/src/scripts/import_arpa_reporter_dump.js
mattbroussard: I guess I should take a second pass over ARPA code to see if there's any behavior that's gated on `user.agency_id` -- my impression before was that all permissions are already at tenant scope in ARPA.

packages/server/src/scripts/import_arpa_reporter_dump.js
mattbroussard: Sure. In practice, the existing deployments have a single tenant (ID `0`) so there would be at most a single new agency created, so I don't think it's too confusing, but adding logging is easy.

packages/server/src/scripts/import_arpa_reporter_dump.js
mattbroussard: Yeah, I'll do that in the ARPA-side PR.

In practice though, these are unlikely to be changed since they are used to reference files already on disk and any substantial change to the logic here would break those references too.

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+async function importTable(tableName, rows, idLookupByTable) {
+    console.log("Importing table", tableName, "...");
+
+    const rowsToInsert = rows.map((row) =>
+        rekeyForeignKeys(_.omit(row, "id"), idLookupByTable)
+    );
+
+    const inserted = await knex(tableName).insert(rowsToInsert).returning("*");
+    const idLookup = _.chain(inserted)
+        .map((insertedRow, idx) => [rowsToInsert[idx].id, insertedRow.id])
```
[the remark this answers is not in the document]
mattbroussard: Oops, yes, you're right!

I banged out this script pretty quickly and have not tested far enough to catch such a bug yet.

mattbroussard:
Thanks for the comments, @trumanc! Some great callouts here.

packages/server/src/scripts/import_arpa_reporter_dump.js
```diff
@@ -0,0 +1,469 @@
...
+) {
+    // First, create all agencies, defaulting their parent pointer to point to themselves.
+    const agenciesToCreate = dbContents.agencies.map((agency) =>
+        rekeyForeignKeys(
+            {
+                tenant_id: agency.tenant_id,
+                name: agency.name,
+                code: agency.code,
+                abbreviation: agency.code,
+                parent: knex.ref("id"),
```
mattbroussard, line 202: Note to self: root should have null parent, not self ref

packages/server/src/scripts/import_arpa_reporter_dump.js
mattbroussard: As far as I can tell from grepping for `.agency_id`, `user.agency_id` is only used for display -- do you know of anything more consequential I've missed @ajhyndman?

mattbroussard:
@trumanc @ajhyndman this should be ready to look at again -- I've tested locally and everything seems right to me, but we should probably do a test in staging environments where there's slightly more "real" data.

packages/server/src/scripts/import_arpa_reporter_dump.js
ajhyndman: Huh, it looks like you're right.

I had thought that we restricted upload visibility to uploads belonging to your agency.  I just tested though, and that is definitely not the case.

trumanc (approved):

PR #318, opened 2022-09-07
[DO NOT REVIEW] Update db seed files with new default values

### LINKED ISSUE #317: [Infra] Update db seed files

mhuang-usdr:
### Description
Closes #317.

Assuming that the entire app is stood up from scratch, there should be a pretty minimal set of tenants, root agencies, and users by default. Additional tenants, agencies, and users should be populated by the user instead.

Thus, this change deletes a lot of legacy values left over from when the app was intended to be deployed in a distributed manner.


mhuang-usdr:
I didn't realize the seed files were so intricately tied to the tests. @as1729 @TylerHendrickson you can peruse this PR to see what the default state of a new instance of this tool should be, but overall this PR should be scrapped and rethought alongside our testing strategy.

PR #319, opened 2022-09-08
ARPA Integration: Add seed for ARPA tables reporting_periods and application_settings

mattbroussard:
Like the title says, this adds a seed that backfills `reporting_periods` and `application_settings` tables based on existing tenants. This serves two purposes:
- Populate these tables in new dev setups (`yarn db:seed` as prescribed in readme will run this)
- One-time backfill for existing tenants in staging/prod environments (`npx knex seed:run --specific=02_backfill_arpa_reporting_periods.js`)

As far as ARPA Reporter's own seeds:
- We won't be copying any of the "dev" seeds; instead we'll use GOST's for shared tables and this new seed for the ARPA tables (these are the only two tables beyond users/agencies/roles that have to be populated in order for the app to work)
- We will use ARPA Reporter's test seeds while running ARPA Reporter tests. In https://github.com/usdigitalresponse/arpa-reporter/pull/515 we update those to play nicely in GOST environment (addressing differences in shared tables)


packages/server/seeds/dev/02_backfill_arpa_reporting_periods.js
```diff
@@ -0,0 +1,95 @@
...
+        });
+
+        start.add(3, 'months');
+    }
+
+    await knex('reporting_periods').insert(periods);
+}
+
+async function seedApplicationSettings(knex, tenantId) {
+    // TODO(mbroussard): Is this right? Should we instead pick one close to the current date?
```
ajhyndman, line 58: No, this is still a good assumption.  We're expecting new partners to have to backfill old data for previous reporting periods.

packages/server/seeds/dev/02_backfill_arpa_reporting_periods.js
```diff
@@ -0,0 +1,95 @@
...
+    const { firstPeriod } = await knex('reporting_periods')
```
ajhyndman, line 59: Nit: Being incredibly picky, I would prefer the name `firstPeriodId`.

packages/server/src/scripts/create_tenant.js
```diff
@@ -3,6 +3,7 @@ require('dotenv').config();
 const inquirer = require('inquirer');
 const { validate: validateEmail } = require('email-validator');
 const knex = require('../db/connection');
+const { seedReportingPeriods, seedApplicationSettings } = require('../../seeds/dev/02_backfill_arpa_reporting_periods');
```
ajhyndman, line 6: I think I would rather import from a script to seed file than the reverse, but that's a super easy refactor.  No reason to hold up merging.

ajhyndman (approved):

PR #320, opened 2022-09-09
Fix email.test.js for dev

mattbroussard:
In dev environment, this test was failing because the default `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` from `.env.example` are blank. In CI, we set fake values for them, so I did the same thing here.

I also reorganized the `process.env` modifications in the test to make the cleanup a bit cleaner (using Mocha `beforeEach` and `afterEach` hooks).

### DIFF
--- a/packages/server/__tests__/email/email.test.js
+++ b/packages/server/__tests__/email/email.test.js
@@ -19,23 +19,47 @@ const {
 } = process.env;
 
 const testEmail = {
-    toAddress: TEST_EMAIL_RECIPIENT || 'test@gmail.com',
+    toAddress: TEST_EMAIL_RECIPIENT || 'nobody@example.com',
     subject: 'Test email',
     body: 'This is a test email.',
 };
 
 describe('Email module', () => {
+    function restoreEnvironmentVariables() {
+        process.env.TEST_EMAIL_RECIPIENT = TEST_EMAIL_RECIPIENT;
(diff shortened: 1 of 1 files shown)


packages/server/__tests__/email/email.test.js
```diff
@@ -19,23 +19,47 @@ const {
...
+
+    function clearSESEnvironmentVariables() {
+        delete process.env.AWS_ACCESS_KEY_ID;
+        delete process.env.AWS_SECRET_ACCESS_KEY;
+        delete process.env.SES_REGION;
+        delete process.env.NOTIFICATIONS_EMAIL;
+    }
+
+    afterEach(() => {
+        restoreEnvironmentVariables();
```
bspates, line 55: Was there some specific bug you ran into where this was needed?

bspates (approved):
LGTM, just a question about the motivation for the env vars change.

packages/server/__tests__/email/email.test.js
mattbroussard: No; I did this while debugging the original failure because I thought the issue might be related to side effects from previous tests. It turned out not to be, but this felt cleaner anyway -- before I think we were depending on the order of tests executing since some of the setup/cleanup was in `it(...)` blocks.

PR #323, opened 2022-09-12
Load environment variables from .env in knexfile.js

ajhyndman:
I was trying to set up a working GOST environment this evening, and was very confused that none of the variables I set in my .env file were respected.

Finally, I realized that this knexfile wasn't actually loading dotenv, and therefore the .env file is ignored.   I assume that's just an oversight?  Any reason not to remedy that here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,12 +88,6 @@ Then export the environment variables
 > cd packages/server && export $(cat .env)
 ```
 
-Set environment variable pointing to local postgres DB, this is used for migrations (knex does not load .env file)
-
-`export POSTGRES_URL=postgresql://localhost:5432/usdr_grants` (individual vars) or `export $(cat .env)` (whole file)
-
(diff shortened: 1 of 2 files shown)


mattbroussard (approved):
The main server calls it [here](https://github.com/usdigitalresponse/usdr-gost/blob/d8eea7cb6b6bb4145d8a8857fa97892f2bc5b058/packages/server/src/configure.js#L2), but other scripts (and particularly things run by the `knex`) CLI do not consistently and I've had to manually set POSTGRES_URL when calling some of them in the past. I don't think this change should hurt anything and could improve usability for new devs.

It looks like the README currently [mentions](https://github.com/usdigitalresponse/usdr-gost/blob/3ae931363fcbbfeaf2418c66ccb2bd92a81f75ca/README.md?plain=1#L91) this, so would be good to update that as well with this change.

Issue #199, opened 2022-08-02
[ID Tool] Adding a user that already exists should show an error.


mhuang-usdr:
Current behavior:
- Add a user that already exists
- Click Ok
- Nothing happens - the modal doesn't close, there's no reaction.

Expected behavior:
- Add a user that already exists
- Click Ok
- There's a pop-up, or some sort of error message that says "email@email.com already exists."

slants3:
I can take this one

slants3:
@mhuang-usdr this has been added

mhuang-usdr:
Looks great! Thanks!

PR #345, opened 2022-09-15
Handle and display user creation errors

slants3:
### Ticket
#199
### Description
We need an error alert when attempting to add users with an email that already exists in our database. We ideally wanted to port over the existing stylization from arpa-reporter. Since arpa-reporter and gost have been merged, this has been made far easier.

In this PR, we copy essentially what has already been working for arpa-reporter by adding an `alert` concept in our store. Any errors now can trigger the alert popup by committing the `addAlert()` action which will cause the `AlertBox` to appear anywhere in the gost dashboard.

image


packages/client/src/store/index.js
```diff
@@ -24,4 +28,15 @@ export default new Vuex.Store({
...
+    alerts: {},
+  },
+  mutations: {
+    addAlert(state, alert) {
+      Vue.set(state.alerts, randomId(), alert);
+    },
+    dismissAlert(state, alertId) {
+      Vue.delete(state.alerts, alertId);
+    },
+  },
```
mattbroussard, line 41: nit: please implement as a separate module (like the other parts of GOST store are) rather than adding at top-level.

slants3: makes sense!

packages/client/src/store/modules/users.js
```diff
@@ -57,8 +57,12 @@ export default {
         .then((data) => commit('SET_USERS', data));
     },
     async createUser({ dispatch }, user) {
-      await fetchApi.post('/api/organizations/:organizationId/users', user);
-      await dispatch('fetchUsers');
+      try {
+        await fetchApi.post('/api/organizations/:organizationId/users', user);
+        await dispatch('fetchUsers');
+      } catch (error) {
```
mattbroussard, line 63: nit: there's not really a point in this try/catch since you're just rethrowing

mattbroussard (approved):

PR #372, opened 2022-09-27
129 - fix: ensures rejected grants are filtered appropriately

as1729:
### Description
- Rejected grants are currently not scoped to the agency. After this commit, users can only see rejected grants that are part of their agency.

### Screenshots / Demo Video
https://www.loom.com/share/21ea1fd0f1be491aaf7cd8be8bfacbb9

### Testing
- No unit-test coverage here. Adding Loom video.

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/GrantsTable.vue
+++ b/packages/client/src/components/GrantsTable.vue
@@ -216,7 +216,7 @@ export default {
           currentPage: this.currentPage,
           orderBy: this.orderBy,
           searchTerm: this.debouncedSearchInput,
-          interestedByAgency: this.showInterested,
+          interestedByAgency: this.showInterested || this.showRejected,
           interestedByMe: this.showMyInterested,
           aging: this.showAging,
           assignedToAgency: this.showAssignedToAgency,


mhuang-usdr:
One test I didn't see in the video was checking another agency within the same tenant. For example

1. If USDR rejected a bunch of grants
2. USDR's sub agency should not see that show up in the Rejected tab
3. USDR's peer agencies should also not see that show up in the Rejected tab.

But across tenants, this behavior looked right!

as1729:
@mhuang-usdr great points! I added some additional test cases here: https://www.loom.com/share/573c97f24e1f4f26b8141ff6d5a4fac5

I was a bit unsure of what is meant by a sibling-agency. In the example data below, I took it to mean that `TSDR` and `BSDR` are sibling agency. Since `ASDR` is a sub-agency of `TSDR` I did not consider that to be a sibling agency.

```
 abbreviation | parent | main_agency_id | tenant_id 
--------------+--------+----------------+-----------
 USDR         |        |              0 |         1
 TSDR         |      0 |              0 |         1
 ASDR         |    400 |              0 |         1
 BSDR         |      0 |              0 |         1
```

as1729:
@TylerHendrickson I'm aligned on this functionality being in the server-side. I'm just unsure if all requests _must_ be filtered down to be scoped by agency. I'm still reviewing the client-side usage of the `/grants` endpoint to confirm where this might break from existing functionality. Will update with server-side changes once I have a good answer here.

mhuang-usdr:
The new demo video LGTM. 

Re: client vs server-side filtering - #309 introduced a filtering pattern for the Interested and Assigned tabs already. Given that this ticket is time-sensitive (_can we have it hit prod before the onboarding at 9am PT tomorrow?_) I would match whatever pattern already exists, and create a new issue to think through a better design.

as1729:
@TylerHendrickson or @mattbroussard I think this is worth approving for now given we want to meet the noon ET deploy deadline. I'll follow-up with a better pattern for the grants data retrieval after reviewing bspates old PR.

TylerHendrickson:
@mhuang-usdr @as1729 Agreed. Since not implementing server-side doesn't introduce a regression and doesn't appear to be sensitive, I think we can merge this as-is and follow up with a backend change later. LGTM! 🚀

TylerHendrickson (approved):

Issue #404, opened 2022-09-27
[ARPA/Infra] When new users are added to the ARPA Reporter, they should receive an email directing them to the ARPA tool.


mhuang-usdr:
Right now, adding users via the ARPA UI still directs them to grants.usdigitalresponse.org, which is the ID tool. This caused confusion for one of our partners.

The problem will be mitigated by #360 , but if we have capacity to make this fix I think it'll help the user experience tremendously.

as1729:
@mhuang-usdr can you confirm if this is working as intended in staging? If so, we can push to prod.

mhuang-usdr:
QA'd, LGTM

mhuang-usdr:
@as1729

as1729:
Can you approve this production release? https://github.com/usdigitalresponse/usdr-gost/pull/415

PR #408, opened 2022-09-28
feat: ensure arpa_reporter users are redirected to ARPA tool upon user creation

as1729:
### Ticket #404
https://github.com/usdigitalresponse/usdr-gost/issues/404

### Description
Currently the welcome email for all new users defaults to the ID tool. This PR ensures that when users are being created for the ARPA reporter tool-- then they receive an email redirecting them to the ARPA tool.

### Screenshots / Demo Video
https://www.loom.com/share/35eb95346f1448c38919660de7c9f951

### Testing
TODO: add unit test

### Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/routes/users.js
+++ b/packages/server/src/arpa_reporter/routes/users.js
@@ -67,13 +67,7 @@ router.post('/', requireAdminUser, async function (req, res, next) {
       const updatedUser = await createUser(user)
       res.json({ user: updatedUser })
 
-      let origin = req.headers.origin;
-
-      if (req.headers.referer.indexOf('arpa_reporter')) {
-        origin += '/arpa_reporter'
-      }
-
-      void sendWelcomeEmail(updatedUser.email, origin)
+      void sendWelcomeEmail(updatedUser.email, req.headers.origin + '/arpa_reporter')
     }
   } catch (e) {
     console.dir(e)
--- a/packages/server/src/arpa_reporter/routes/users.js
+++ b/packages/server/src/arpa_reporter/routes/users.js
@@ -67,7 +67,13 @@ router.post('/', requireAdminUser, async function (req, res, next) {
       const updatedUser = await createUser(user)
       res.json({ user: updatedUser })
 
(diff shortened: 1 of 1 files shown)


packages/server/src/arpa_reporter/routes/users.js
```diff
@@ -67,7 +67,13 @@ router.post('/', requireAdminUser, async function (req, res, next) {
       const updatedUser = await createUser(user)
       res.json({ user: updatedUser })
 
-      void sendWelcomeEmail(updatedUser.email, req.headers.origin)
+      let origin = req.headers.origin;
+
+      if (req.headers.referer.indexOf('arpa_reporter')) {
```
mattbroussard, line 72: nit: this route is ARPA Reporter-specific anyway, so this referer check is not necessary

mattbroussard (approved):

packages/server/src/arpa_reporter/routes/users.js
mattbroussard: also, if you are going to keep it, you need to check `!== -1` because `-1` is truthy

mattbroussard: also, does this NPE if there is no Referer header? Might need an optional chain

as1729: great point, I'll go ahead and remove this conditional check entirely. I was wondering whether the `arpa-annual-reporter` endpoint uses the same route so wanted add some type of `if/else` logic.

as1729: @mattbroussard can you re-review I have resolved the issue by removing the unnecessary conditional.

mattbroussard: >I was wondering whether the arpa-annual-reporter endpoint uses the same route
That tool is only a single page and doesn't have its own user admin page etc. -- it just uses GOST's.

>can you re-review
fwiw, "approved with comments" generally means the comments are minor enough that reviewer trusts you to make the changes and merge without another review round (still feel free to ask for additional review if you substantially changed the PR but this change is pretty minor)

PR #418, opened 2022-09-28
Move health route out of ARPA Reporter folder

### LINKED ISSUE #362: Set up health check route in Render
Configure the GOST render deployment such that it can find a health check route on the GOST deployment.

Optionally, move the /health route defined in the ARPA tool to the root of the project, so that it will remain valid after the ARPA grants run their course and the tool is cleaned up.

mattbroussard:
Fixes #362 

Note this doesn't change any functionality because this route already exists (and I've already added Render config in both staging and prod to check it); it just moves the file out of the ARPA Reporter folder so it doesn't get accidentally removed in the future if we end up retiring ARPA Reporter code.

It looks like Render hits the health check route frequently, so I'm excluding it from logging to keep logs useful (otherwise interesting things may get truncated by thousands of health checks...)

ARPA's test for this route is removed entirely and re-added in #419 in order to avoid merge conflicts between these two concurrently open PRs, since that PR changes how route tests in GOST work.


as1729:
> wdyt of migrating that health-check test over to gost repo as well? It may be useful just to have as a test-template for how to test routes.

Confirmed that the other open PR #419 has the test so no changes required here 👍

PR #419, opened 2022-09-29
Run GOST API tests in-process rather than starting a separate server; make it possible to use spies in tests

### LINKED ISSUE #368: Fix the 'test:api' target in server/package.json to make it more reliable

mattbroussard:
Fixes #368

Previously, when running API tests we started a server process separate from the one where Mocha is running and made HTTP calls to it using node-fetch. Now, we start an Express server in the same process with the option to interact with it either through fetch or through Supertest. This is how API tests were working in ARPA Reporter. It would still be nice to unify under a single Mocha run for both sets of tests, but this PR does not attempt that (there are some other issues to contend with such as seeds for user/agency model).

As discussed in [Slack](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1664394088131869?thread_ts=1664383805.406279&cid=C0324KDQSCR), this has several benefits:
 - Avoid [timeout issues](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1663709029578369)
 - Avoid leaving extra processes around if something goes wrong (I think this was fixed in #296's e07ff01c7e4ca54ca06b422c6c82ab6349c702fc, but possible there was still some edge case?)
 - Make it possible to use spies/mocks/etc., such as from [Sinon](https://sinonjs.org/) (added in this PR) or [chai-spies](https://www.chaijs.com/plugins/chai-spies/)
 - Make it possible to run tests while your local dev server is also running (because they don't require the same port number anymore)

In this PR I also added a basic test verifying that it's possible to spy on something using Sinon. The one catch here is that if you're trying to spy on something that's exported at the top level of a module, you need to change the import/call site to import the module as an object and only access the property when it's time to call the function. That is, this works:
```
const email = require('../lib/email');
email.sendWelcomeEmail(...);
```
But this doesn't:
```
const {sendWelcomeEmail} = require('../lib/email');
sendWelcomeEmail(...);
```


chrisxkeith:
I may not be able to look at this until mid-day Friday. Will you still be around?

mattbroussard:
@chrisxkeith no worries! I don't necessarily need review from everyone listed & mainly added multiple reviewers for awareness. I think Ty or Aditya will probably be able to get to it sooner (we were talking about this issue today).

as1729 (approved):
Looks good to me. For the most part it we're transferring existing functionality from arpa over to gost so this all makes sense.

Thanks for adding the sendWelcomeEmail mock as it will be easier for me to add tests in the future and pattern match.

TylerHendrickson (approved):
This looks great- approved! 💯

PR #420, opened 2022-09-29
fix: temporary fix that allows grants to be marked uninterested

as1729:
### Description
The markGrantUninterested action seems to be broken after the implementation of an isAuthorized check which requires a single integer value to be passed in rather than an array.

**Note** - this is a temporary fix that ensures we don't 500 in these requests. A more permanent fix requires us to change the client-side code and clean up what the server should be expecting for this request. I put this up so our users will have the functionality and will plan to revisit this again when I'm back.

### Screenshots / Demo Video
- https://www.loom.com/share/92e6392256aa4c5980743e1d8ae64eae

### Testing
- Adds a happy-path unit test

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/api/grants.test.js
+++ b/packages/server/__tests__/api/grants.test.js
@@ -362,7 +362,7 @@ describe('`/api/grants` endpoint', () => {
                 });
                 expect(response.statusText).to.equal('OK');
             });
-            it('forbids removing grant interest when one of the agencies is not this user\'s agency', async () => {
+            it('forbids removing grant interest when one of the agencies is not in this user\'s tenant', async () => {
                 const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
                     ...fetchOptions.staff,
                     method: 'delete',
@@ -370,7 +370,7 @@ describe('`/api/grants` endpoint', () => {
                 });
                 expect(response.statusText).to.equal('Forbidden');
             });
-            it('forbids removing grant interest when the agency is not this user\'s agency', async () => {
+            it('forbids removing grant interest when the agency is not in this user\'s tenant', async () => {
                 const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.offLimits}`, agencies.own, {
                     ...fetchOptions.staff,
                     method: 'delete',
--- a/packages/server/__tests__/api/grants.test.js
+++ b/packages/server/__tests__/api/grants.test.js
@@ -327,13 +327,56 @@ describe('`/api/grants` endpoint', () => {
     context('DELETE /api/grants/:grantId/interested/:agencyId', () => {
         context('by an admin user', () => {
             const interestEndpoint = `335255/interested`;
-            it('records this admin\'s ability to delete a grant', async () => {
+            it('allows removing grant interest for a single agency', async () => {
                 const response = await fetchApi(`/grants/${interestEndpoint}/undefined`, agencies.own, {
                     ...fetchOptions.admin,
                     method: 'delete',
                     body: JSON.stringify({ agencyIds: [agencies.own], interestedCode: null }),
                 });
                 expect(response.statusText).to.equal('OK');
+            });
+            it('allows removing grant interest for multiple authorized agencies', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/undefined`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own, agencies.ownSub], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('OK');
+            });
+        });
+        context('by a user with a staff role', () => {
+            const interestEndpoint = `335255/interested`;
+            it('allows removing grant interest for a single agency', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('OK');
+            });
+            it('allows removing grant interest for a multiple authorized agencies', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own, agencies.ownSub], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('OK');
+            });
+            it('forbids removing grant interest when one of the agencies is not this user\'s agency', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.offLimits, agencies.own], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('Forbidden');
+            });
+            it('forbids removing grant interest when the agency is not this user\'s agency', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.offLimits}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('Forbidden');
             });
         });
     });
--- a/packages/server/__tests__/api/grants.test.js
+++ b/packages/server/__tests__/api/grants.test.js
@@ -327,7 +327,7 @@ describe('`/api/grants` endpoint', () => {
     context('DELETE /api/grants/:grantId/interested/:agencyId', () => {
         context('by an admin user', () => {
             const interestEndpoint = `335255/interested`;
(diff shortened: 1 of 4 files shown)


packages/server/src/routes/grants.js
```diff
@@ -235,8 +235,15 @@ router.delete('/:grantId/interested/:agencyId', requireUser, async (req, res) =>
...
-    console.log(agencyIds);
-    const allowed = await isUserAuthorized(user, agencyIds || agencyId);
+    let authorizedAgencyId = agencyId;
+    if (agencyIds.length > 0) {
+        // Note: ran into a ES lint error so had to declare the variable first
+        // and then assign it to authorizedAgencyId.
+        const [firstAgency] = agencyIds;
+        authorizedAgencyId = firstAgency;
+    }
+    const allowed = await isUserAuthorized(user, authorizedAgencyId);
```
mattbroussard, line 245: you could do something like this:
```
const allowed = _.every(await Promise.all(
  agencyIds.map(agencyId => isUserAuthorized(user, agencyId))
));
```
...but the query that `inTenant` does has some joins, it's not amazing to do n of them concurrently.

Better yet, I'd just create a version of the helper like `isUserAuthorizedForMultipleAgencies`; should be easy since `isUserAuthorized` just calls `inTenant` right now, which already supports multiple agencies.

packages/server/src/routes/grants.js
```diff
@@ -235,8 +235,15 @@ router.delete('/:grantId/interested/:agencyId', requireUser, async (req, res) =>
     const { user } = req.session;
     const { grantId, agencyId } = req.params;
     const { agencyIds } = req.body;
-    console.log(agencyIds);
-    const allowed = await isUserAuthorized(user, agencyIds || agencyId);
+    let authorizedAgencyId = agencyId;
+    if (agencyIds.length > 0) {
+        // Note: ran into a ES lint error so had to declare the variable first
+        // and then assign it to authorizedAgencyId.
+        const [firstAgency] = agencyIds;
```
mattbroussard, line 242: I know it's "temporary" but I don't really like this idea of just using the first agency; doing so basically negates the whole permission check since someone could pass one valid/authorized agency, followed by anything else

mattbroussard (commented):
>A more permanent fix requires us to change the client-side code and clean up what the server should be expecting for this request.

I don't think clientside code necessarily has to change if you just implement the multi-agency permission check on the server side. It is a little odd that this takes multiple agency IDs when the only callsite on the client always gives a list of 1 though.

packages/server/src/routes/grants.js
TylerHendrickson: I agree with your hesitation to run extra `inTenant` queries concurrently. 

Rather than introduce a new helper function, I made an update that keeps the existing `isUserAuthorized` compatible with existing calls by making the second parameter of that function variadic. That way, it's still possible to only pass a single agency ID value as the second parameter (as any extant calls will be doing), but also allow multiple agency IDs to be passed in for cases like this one.

This way, if more than zero agency IDs are passed in the body, we will use (all of, and only) those values for both the permission check and the subsequent `DELETE` query, and the `agencyId` request param is completely ignored. Otherwise, if request body's `agencyIds` value is empty, we use the singular `agencyId` request param value as a "fallback" and only evaluate that ID.

packages/server/src/routes/grants.js
```diff
@@ -235,14 +235,16 @@ router.delete('/:grantId/interested/:agencyId', requireUser, async (req, res) =>
     const { user } = req.session;
     const { grantId, agencyId } = req.params;
     const { agencyIds } = req.body;
-    console.log(agencyIds);
-    const allowed = await isUserAuthorized(user, agencyIds || agencyId);
+    // If agencyIds is not empty, use that. Otherwise, use the single agencyId value.
+    const selectedAgencyIds = agencyIds.length > 0 ? agencyIds : [agencyId];
```
mattbroussard, line 239: nit: I'd name this something else because `selectedAgencyId` has a specific meaning: the agency that an admin is acting as (see `requireAdminUser`)

packages/server/__tests__/api/grants.test.js
```diff
@@ -324,6 +324,19 @@ describe('`/api/grants` endpoint', () => {
...
         });
     });
+    context('DELETE /api/grants/:grantId/interested/:agencyId', () => {
+        context('by an admin user', () => {
+            const interestEndpoint = `335255/interested`;
+            it('records this admin\'s ability to delete a grant', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/undefined`, agencies.own, {
+                    ...fetchOptions.admin,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own], interestedCode: null }),
```
mattbroussard, line 334: Can you add tests that:
1. Attempt to unmark a grant as interested from multiple agencies (although nothing does this now, the API theoretically supports it, so we should test it)
2. Attempt to unmark a grant as interested when at least one of the specified agencies is of a different tenant (should fail)

mattbroussard (approved):
Variadic args are not my favorite thing for readability but I get why you don't want to go update every single callsite to pass a list. I think it's fine here.

packages/server/__tests__/api/grants.test.js
TylerHendrickson: Done ✅

TylerHendrickson:
@mattbroussard I know you approved already, but if you can spare them, I'd appreciate your 👀 for another go-round on this PR. Hopefully this will be the last time!

packages/server/src/db/index.js
```diff
@@ -790,31 +790,29 @@ async function sync(tableName, syncKey, updateCols, newRows) {
...
+
+    const result = await knex(
+        knex('agencies')
+            .select('id as agency_id')
+            .where('tenant_id', tenantId)
+            .whereIn('id', uniqueAgencyIds)
+            .as('agencies_in_tenant'),
+    ).select(
+        knex.raw(
+            'coalesce(? <@ array_agg(agencies_in_tenant.agency_id), false) as same_tenant',
```
mattbroussard, line 810: This clause is not super readable; please add a comment.

packages/server/src/db/index.js
```diff
@@ -790,31 +790,29 @@ async function sync(tableName, syncKey, updateCols, newRows) {
...
 
 /**
- * Determines if a user's tenant is the same as the agency's tenant
+ * Determines if all given agencies belong to the same given tenant
  * @param  int        userId
  * @param  int        tenantId
  * @parm   Array[int] agencyIds
  * @return boolean
  * */
 async function inTenant(userId, tenantId, agencyIds) {
```
mattbroussard, line 799: It appears we're no longer using the `userId` arg and all existing callsites pass `user.tenant_id` for `tenantId` anyway (thus checking `user.tenantId === tenantId` would be sorta useless). Should we just remove the `userId` param then?

packages/server/__tests__/api/grants.test.js
```diff
@@ -324,6 +324,62 @@ describe('`/api/grants` endpoint', () => {
...
+                });
+                expect(response.statusText).to.equal('OK');
+            });
+            it('allows removing grant interest for multiple authorized agencies', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/undefined`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own, agencies.ownSub], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('OK');
```
mattbroussard, line 344: should we be asserting on the actual database state after the call here, not just the HTTP response? (and same for the other cases here)

packages/server/__tests__/api/grants.test.js
```diff
@@ -324,6 +324,62 @@ describe('`/api/grants` endpoint', () => {
...
+            });
+            it('allows removing grant interest for a multiple authorized agencies', async () => {
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.own, agencies.ownSub], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('OK');
+            });
+            it('forbids removing grant interest when one of the agencies is not this user\'s agency', async () => {
```
mattbroussard, line 365: nit: should say "is not in this user's tenant"

packages/server/__tests__/api/grants.test.js
```diff
@@ -324,6 +324,62 @@ describe('`/api/grants` endpoint', () => {
...
+                const response = await fetchApi(`/grants/${interestEndpoint}/${agencies.own}`, agencies.own, {
+                    ...fetchOptions.staff,
+                    method: 'delete',
+                    body: JSON.stringify({ agencyIds: [agencies.offLimits, agencies.own], interestedCode: null }),
+                });
+                expect(response.statusText).to.equal('Forbidden');
+            });
+            it('forbids removing grant interest when the agency is not this user\'s agency', async () => {
```
mattbroussard, line 373: nit: should say "is not in this user's tenant"

mattbroussard (approved):

packages/server/__tests__/api/grants.test.js
TylerHendrickson: I like where you're going with this, but I think it's worth first deciding what purpose/proof we think these tests provide. Right now, they come across like integration tests more than unit tests, and I don't think integration tests should need to check DB state. 

NB: I'd prefer to have unit tests without DB connectivity as well, but I think that's a larger effort beyond the scope of this PR.

Issue #423, opened 2022-09-29
Updated Guidance for Oct 2022 Reporting Period


jcomeau1:
There are changes specific to EC 5.19 - 5.21 that must be made to both the Input Template and the Output template, project519521BulkUpload.

**Input Template EC5** 
- [x] Remove visibility (via Logic sheet) of Planned_Funded_Locations_Minimum_1_Gbps__c (column AX) and Actual_Funded_Locations_Minimum_1_Gbps__c (column AY)
- [x] Adjust column AM to **Is project designed to provide service to households?**
	Pick List, Yes/No
	Conditional on whether project has started
- [x] o	Adjust column BN to **If yes, Confirm that the service provider for the project has, or will upon completion of the project, either participated in the Federal Communications Commission (FCC)s Affordable Connectivity Program (ACP) or otherwise provided access to a broad-based affordability program that provides benefits to households commensurate with those provided under the ACP to low-income consumers in the proposed service area of the broadband infrastructure** 
	Pick List, Yes/No
	Conditional on column AM=”Yes”
- [x] Add Planned_Sum_Speed_Types_Explanation__c (column BP) and visibility set (via Logic sheet)
	Conditional, though no conditional logic has been added
	String with max length of 1500
- [x] Add Actual_Sum_Speed_Types_Explanation__c (column BQ) and visibility set (via Logic sheet)
	Conditional, though no conditional logic has been added
	String with max length of 1500

**Render**
- [ ] Update Input Template to v20220929
- [ ] Update output template **project519521BulkUpload**

PR #424, opened 2022-09-30
423: Update arpa-reporter for new output templates and updated workbook

trumanc:
### Ticket #
#423 
### Description
Treasury has sent out their updated guidance for the upcoming reporting period, and this PR applies the changes needed so that our provided workbook is updated to ask new questions, and the output is formatted to meet the new specifications.

This PR takes all the work done in a previously abandoned PR in the legacy repo (https://github.com/usdigitalresponse/arpa-reporter/pull/512) but it includes a few more updates on an even newer workbook version, plus updates to the output templates.

See the ticket linked above for description of the major changes, but in summary:
- The 5.19-5.21 ec codes had a couple fields removes, deleted a couple redundant copies of other fields, moved/clarified one field, and added two new questions.
- Subrecipients have a new id field that shold be null for any new subrecipient records. We'll eventually update this to ingest/use the id coming from the treasury portal.
- All the output templates for projects use `Q3_2022` instead of Q2 for their obligation/expenditure fields.
- Some dropdown values have been updated after testing in the portal to see what values it accepts.

The `generateRules.mjs` script had incorrect imports (directory structure was different when it was imported over into this repo) so I corrected those imports in order to run the script.
That script was run on the new workbook file (which was configured in the `environment.js` file, and generated the `templateDropdowns.json` and `templateRules.json` files and the changes in those.

The changes in `validation-rules.js` were made after hand-checking what dropdown values are accepted by the treasury portal, using a system that lets us re-write incorrect dropdown values into the correct version.

All the output templates were compared to their old versions to verify any changes in field-ids, and then had those changes reflected in the `generate-arpa-report.js` file

### Checklist
- [x] Added PR reviewers


TylerHendrickson (approved):
@trumanc Changes make sense as far as I can tell. 

I recognize that the test suites for this repo in general are a little bit lacking, and perhaps for the reporter tool (specifically around report generation) especially; rather than attempt to address that here, I think strategies for strengthening tests for changes like these might be a good topic for an upcoming team discussion.

PR #427, opened 2022-09-30
426: Add tooling for docker-based development environments

TylerHendrickson:
### Description

This PR adds support for using Docker to create and manage development environments. 

In particular, it includes the following changes:
- Defines container image build instructions for backend and frontend containers (see `docker/Dockerfile`)
- Defines container service dependencies for managing containerized development environments with Docker Compose (see `docker-compose.yml`)
- Adds automation for Postgres setup when using Docker Compose (see `docker/postgres/`, `docker-compose.yml`, and `.gitignore`)
- Add documentation for using Docker and Docker Compose for development (see `docker/README.md`)
- Conditionally sets Vue development server's config options at `devServer. allowedHosts` and `devServer.proxy['/api'].target` with values derived from `VUE_ALLOWED_HOSTS` and `VUE_APP_API_URL` env vars, respectively (see `packages/client/vue.config.js`)

#### Notes for reviewers

- The `app_base` image defined in `docker/Dockerfile` recursively sets permissions on the `/root` directory to `777` within the container. I don't love this, but it was the only way that I could ensure that `lerna`/`npx` wouldn't run into permissions issues, and I wasn't able to find a definitive set of best practices that demonstrated an alternative approach. If you know of a better way to handle configuring Node in Docker environments, I'd love to hear it!
- Pretty much everyone on the internet has decided that [this StackOverflow answer](https://stackoverflow.com/a/32785014) is the canonical recommendation for mounting `node_modules/` in Docker containers that install dependencies during image build. I can't see a problem with the approach, but something does feel a bit "off" about it to me. Again, please enlighten me if you know better :) 
- Probably obvious, but just in case it isn't: this does not (should not) change anything related to Production! If you see anything that you think _would_ impact anything other than dev environments, please let me know.
- I noticed another Docker Compose file at `packages/server/docker-compose.yml`. I left it alone for this PR because I didn't want to break anyone's workflow, but it would be nice to clean this up if we're confident its usefulness is covered by the new `docker-compose.yml` file added in the repository root.

### Screenshots / Demo Video

N/A

### Testing

Follow the instructions documented in the newly-added `docker/README.md` file. 

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/vue.config.js
+++ b/packages/client/vue.config.js
@@ -16,12 +16,13 @@ module.exports = {
     },
   },
   devServer: {
+    allowedHosts: process.env.VUE_ALLOWED_HOSTS || 'auto',
     client: {
       progress: false,
     },
     proxy: {
       '/api': {
-        target: 'http://localhost:3000',
+        target: process.env.VUE_APP_API_URL || 'http://localhost:3000',
       },
     },
     historyApiFallback: {
--- a/packages/client/vue.config.js
+++ b/packages/client/vue.config.js
@@ -22,7 +22,7 @@ module.exports = {
     },
     proxy: {
       '/api': {
-        target: process.env.VUE_APP_API_URL || 'http://localhost:3000',
+        target: process.env.GOST_API_URL || 'http://localhost:3000',
       },
     },
     historyApiFallback: {
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,6 @@ packages/server/src/static/forms/generated
 
 # ARPA Reporter uploads
 packages/server/data
+
+# Locally-persisted postgres development data
+docker/postgres/persistence
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ Each folder inside packages/ is considered a workspace. To see a list of all wor
 
 These steps are for an install on a Mac. The Windows instructions are [here](https://github.com/usdigitalresponse/usdr-gost/wiki/Setting-up-a-development-environment-on-Windows-(native)).
 
+Instructions for using Docker to manage development environments can be found
+[here](docker/README.md).
+
 1). Ensure using the correct version of NODE Version
 
 First, check the [`.nvmrc` file](./.nvmrc) to make sure you have the correct version of Node.js installed. If you are using [Nodenv](https://github.com/nodenv/nodenv) or [NVM](https://nvm.sh/), it should pick up on the correct version.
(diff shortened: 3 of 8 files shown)


packages/client/vue.config.js
```diff
@@ -16,12 +16,13 @@ module.exports = {
...
   },
   devServer: {
+    allowedHosts: process.env.VUE_ALLOWED_HOSTS || 'auto',
     client: {
       progress: false,
     },
     proxy: {
       '/api': {
-        target: 'http://localhost:3000',
+        target: process.env.VUE_APP_API_URL || 'http://localhost:3000',
```
mattbroussard, line 25: why was it necessary to make this an env var?

nit: name it differently, this use has nothing to do with Vue and `VUE_APP_*` environment variables have [special semantics](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables) (namely that Vue CLI configures webpack to embed those environment variables in the client bundle)

chrisxkeith:
I propose the documentation be structured to put the Docker setup first and foremost, since that is going to be our recommended way of working. This can be done by:

- Merging the non-Windows-specific instructions in [this wiki page](https://github.com/usdigitalresponse/usdr-gost/wiki/Setting-up-development-environment-in-Docker-on-Windows) into the Docker README.md.
- Moving the contents of the current README.md file somewhere else (a wiki page?)
- Putting the contents of the Docker README.md into the top level README.md.

That said, I am fine with accepting this PR as it stands. I don’t want it to get stale (or add a bunch more work for Ty). I will then create another issue to do the documentation restructuring described above, and assign that issue to myself.

Let me know (in this conversation) if that works for you folks.

packages/client/vue.config.js
TylerHendrickson: > why was it necessary to make this an env var?

Within the Docker Compose runtimes, the cross-container hostnames are expose according to the respective service names defined in `docker-compose.yml` rather than `localhost`, so (e.g.) the `frontend` container contacts the backend at `http://app:3000` rather than `http:localhost:3000`. I didn't want to break anyone's existing dev setup by assuming they'd be using Docker, so I opted to have this configurable via environment variable rather than hard-coding the container service name.

Regarding the nit- that's a good callout. I'll do something like `GOST_API_URL` instead.

chrisxkeith (approved):
> I'd prefer to get this merged first, be able to keep my ear to the ground as usage ramps up, and then come back with another PR that focuses on documentation updates.
Works for me!

PR #475, opened 2022-10-14
fix: permissions to run docker setup locally

as1729:
### Description
My local machine ran into permissions issues when running the Docker setup. Specifically:
1. Issue with bitnami/postgresql not having the permissions to execute example.sh
2. Persistent data folder of the postgresql container was not mounted to the local machine.

Both these are now resolved and the dev setup functions as expected.

### Testing
Tested with @TylerHendrickson live on a google meet.

### Checklist
- [x] Added PR reviewers


docker-compose.yml
```diff
@@ -53,7 +53,7 @@ services:
     volumes:
       - ./docker/postgres/docker-entrypoint-initdb.d:/docker-entrypoint-initdb.d
       - ./docker/postgres/docker-entrypoint-preinitdb.d:/docker-entrypoint-preinitdb.d
-      # - ./docker/postgres/persistence:/bitnami/postgresql
+      - ./docker/postgres/persistence:/bitnami/postgresql
```
TylerHendrickson, line 56: I'd leave this part commented out, since I don't think it's generally useful/desired behavior to have DB state persist between tear-downs. As long as we think it will work if someone chooses to uncomment it locally, I think we're good.

TylerHendrickson (approved):
Requesting one tiny change, but shouldn't need a re-review.

PR #525, opened 2022-10-25
Revert "Revert "Merge _staging into main""

as1729:
Reverts usdigitalresponse/usdr-gost#523


as1729:
```
BEGIN;

alter table tenants drop column uses_spoc_process;

/* verify that the column is dropped */
select * from tenants;


/* if previous query looks good go ahead and commit */
commit;
```

I will be running this statement prior to running migrations.

TylerHendrickson (approved):

Issue #386, opened 2022-07-18
Invalid subrecipient in Period 1 should not prevent uploading in Period 2


ajhyndman:
OH saw a case in which a subrecipient was part of an upload to the ARPA tool in Period 1.  Even though this subrecipient was never part of a "valid" upload, it was excluded (incorrectly) from subrecipient exports in Period 2.

mhuang-usdr:
@trumanc can this be closed?

as1729:
This does not seem to be a good first issue. @as1729 to work with @jcomeau1 and figure out the requirements and eng-plan prior to assigning this over to a developer.

as1729:
When the file is invalidated, any subrecipients created by this invalidated upload must also be removed from the subrecipients tab as a valid subrecipient.

as1729:
Same as #1675

PR #533, opened 2022-10-26
Simple db migration + missing migration checker

trumanc:
### Ticket #
https://github.com/usdigitalresponse/usdr-gost/issues/386

### Description
This is the first step in implementing a planned improvement to subrecipient reporting, described in slack here https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1665694828940029 (with a linked google doc)

This PR performs the db migration and adds a couple columns to the database, without any code that actually requires those columns yet. That will allow us to safely perform this migration at our leisure without breaking any deployments.

It also adds a migration version checker that will crash local/dev deployments and tell you to run the latest migration before continuing.
### Screenshots / Demo Video

These screenshots show that when there is an outstanding migration, the `npm run serve` command fails with a useful error message. After running the migration, the deployment works:
Screen Shot 2022-10-26 at 10 21 57 AM
Screen Shot 2022-10-26 at 10 22 24 AM

This screenshot shows the migration was correctly reversible by running up/down/up without any errors

Screen Shot 2022-10-26 at 10 22 40 AM

### Checklist
- [x] Added PR reviewers


as1729 (approved):
Thanks for adding the additional database migration check!

packages/server/src/index.js
```diff
@@ -39,4 +40,14 @@ const cleanGeneratedPdfCron = new CronJob(
 );
 cleanGeneratedPdfCron.start();
 
+hasOutstandingMigrations().then((hasMigrations) => {
+    if (!hasMigrations) {
+        return;
+    }
+    console.error('There are outstanding db migrations. Run \'yarn db:migrate\' before trying again');
+    if (process.env.NODE_ENV === 'development') {
+        process.exit(1);
```
TylerHendrickson, line 49: Instead of crashing, would it be helpful to instead just run the migrations? I like the check, and it's definitely an improvement- just wondering if it could go a step further and eliminate the chore. 

Basically, I'm wondering if there's ever a scenario where this check crashes the app and a developer's immediate next step _wouldn't_ be running the migrations manually.

TylerHendrickson (approved):

PR #578, opened 2022-11-08
494: Add Export CSV Functionality to Recent Activities

slants3:
Add new endpoint in grants for downloading an CSV for recent activities. The endpoint largely follows the conventions set in the existing `/exportCSV` endpoint. 

image

image


TylerHendrickson (changes requested):
Functionality all looks good! Two requests ahead of approval:
1. Please add some unit tests that cover the server-side execution path for the `/exportCSVRecentActivities` endpoint. 
2. I feel like there should be some (reasonable) upper bounds enforced for the `perPage` and `currentPage` query parameters. My concern is that the amount of data fetched from the DB and the resulting CSV file could get quite large, since I don't see anything that naturally limits the amount of data in the export.

packages/server/src/db/index.js
```diff
@@ -74,6 +74,10 @@ async function createUser(user) {
     };
 }
 
+async function getUsersEmailAndName(ids) {
+    return knex.select('id', 'name', 'email').from('users').whereIn('id', ids);
```
as1729, line 78: I believe we'll need to add an `await` here since this query will be run async. Can you confirm?

packages/server/src/db/index.js
```diff
@@ -74,6 +74,10 @@ async function createUser(user) {
     };
 }
 
+async function getUsersEmailAndName(ids) {
```
as1729, line 77: would recommend adding a unit test for this function 🙏

packages/server/src/routes/grants.js
```diff
@@ -154,6 +154,50 @@ router.get('/exportCSV', requireUser, async (req, res) => {
     res.send(csv);
 });
 
+router.get('/exportCSVRecentActivities', requireUser, async (req, res) => {
```
as1729, line 157: Ideally we can add some tests around this route- let me know if you'd like any assistance or patterns since there might not be much available in the repo.

as1729 (changes requested):

packages/server/src/db/index.js
slants3: Wouldn't prefacing calling the endpoint with await produce the same output? I just used examples set by other endpoints but can change it

packages/client/src/views/RecentActivity.vue
```diff
@@ -1,7 +1,15 @@
 <template>
   <section class="container">
     <b-card class="border-0">
+    <div class="d-flex">
       <h4 class="card-title gutter-title1 row">Recent Activity</h4>
+      <div class="justify-content-end left-margin">
+        <b-button @click="exportCSV" :disabled="loading" variant="outline-secondary">
+        <b-icon icon="download" class="mr-1 mb-1" font-scale="0.9" aria-hidden="true" />
+          Export to CSV
+          </b-button>
```
TylerHendrickson, line 10: Nit: the closing `</b-button>` tag is indented further than the opening `<b-button>` tag on line 7.

TylerHendrickson (approved):
One (tiny) nit, but these changes look great! 🚀

packages/client/src/views/RecentActivity.vue
slants3: You're totally right - missed that!

PR #597, opened 2022-11-14
feat: add static email templates

as1729:
### Description
- Adds base email template static assets.

Note: This PR only adds the html. Follow-up PRs will build the templates.

### Screenshots / Demo Video
image

### Testing
N/A

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/static/email_templates/_grant_assigned_body.html
+++ b/packages/server/src/static/email_templates/_grant_assigned_body.html
@@ -1,0 +1,43 @@
+<table class="es-content" cellspacing="0" cellpadding="0" align="center"
+    style="mso-table-lspace:0pt;mso-table-rspace:0pt;border-collapse:collapse;border-spacing:0px;table-layout:fixed !important;width:100%">
+    <tr style="border-collapse:collapse">
+        <td align="center" bgcolor="#FFFFFF" style="padding:0;Margin:0;background-color:#ffffff">
+            <table class="es-content-body"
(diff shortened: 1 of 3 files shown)


TylerHendrickson (approved):
Templates look great! 👏 

Doesn't necessarily have to block this PR, but I'm wondering what you think about trying to get rid of all (or as much as possible) `style` attributes on the various HTML tags and instead using `class` and `id` attributes for our CSS selectors. Prior experience tells me that HTML templates are much easier to maintain (and tend to be less buggy) over time when style definitions are restricted to CSS, rather than inline.

NB: I realize that styles must be inline with the HTML when the emails get sent; there are various "CSS inliner" libraries ([here's one](https://github.com/Automattic/juice) I found that's built with JavaScript) that can be used to render inline styles at runtime (or possibly build-time).

Thoughts?

as1729:
> Templates look great! 👏
> 
> Doesn't necessarily have to block this PR, but I'm wondering what you think about trying to get rid of all (or as much as possible) `style` attributes on the various HTML tags and instead using `class` and `id` attributes for our CSS selectors. Prior experience tells me that HTML templates are much easier to maintain (and tend to be less buggy) over time when style definitions are restricted to CSS, rather than inline.
> 
> NB: I realize that styles must be inline with the HTML when the emails get sent; there are various "CSS inliner" libraries ([here's one](https://github.com/Automattic/juice) I found that's built with JavaScript) that can be used to render inline styles at runtime (or possibly build-time).
> 
> Thoughts?

The inliner library is a great idea! I'll look into implementing it in follow-up PRs.

PR #624, opened 2022-11-22
602: Decouple client/server hostnames

TylerHendrickson:
### Description

So as to not bury the lede: **Once this PR is merged, all new API calls made by the client application need to ensure that the new `apiURL()` function is called before the API request is made.** Read on to learn why!

This PR introduces a few relatively minor changes that facilitate scenarios where the deployed client frontend application is hosted separately from the API server backend. In particular, it introduces a new helper function, `apiURL()`, within the `packages/client/src/helpers/fetchApi.js` module. When called with an endpoint path, `apiURL()` checks for a `window.apiURLForGOST` value and prepends it to the endpoint path. If `window.apiURLForGOST` is undefined or false-y, `apiURL()` will return the endpoint path after ensuring that it is an absolute path (begins with `/`). Therefore, these changes preserve existing behavior in both local development environments as well as environments (like Render) where the API and client are hosted together.

It's worth noting that the `window.apiURLForGOST` is not set in this code. Instead, the Reporter and ID tool frontends have been modified to load a `/deploy-config.js` file via a `<script>` added to the public HTML base templates, which enables hosting environments to dynamically configure the backend URL during the initial bootstrapping phase of the website by ensuring that a `/deploy-config.js` file is served from the web root. An example `deploy-config.js` file would simply look like this:
```js
window.apiURLForGOST = 'https://example.com/somewhere/';
```

This behavior has the benefit of keeping builds hosting-environment agnostic, which enables "build once, run many times" behavior from testing through production deployment, while avoiding hard-coded assumptions about hosting that are specific to USDR-maintained environments.

### Testing

- New unit tests added in `packages/client/tests/unit/helpers/fetchApi.spec.js`

### Checklist
- [X] Run automated tests (docker compose exec app yarn test)
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/public/index.html
+++ b/packages/client/public/index.html
@@ -1,26 +1,33 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-     <!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=G-WCDTMFM6RG"></script>
-<script>
-  window.dataLayer = window.dataLayer || [];
-  function gtag(){dataLayer.push(arguments);}
-  gtag('js', new Date());
 
-  gtag('config', 'G-WCDTMFM6RG');
-</script>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <link rel="icon" href="<%= BASE_URL %>favicon.ico">
-    <title><%= htmlWebpackPlugin.options.title %></title>
-    <link rel="stylesheet" href="https://unpkg.com/vue-multiselect@2.1.0/dist/vue-multiselect.min.css">
-  </head>
-  <body>
-    <noscript>
-      <strong>We're sorry but <%= htmlWebpackPlugin.options.title %> doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
-    </noscript>
-    <div id="app"></div>
-  </body>
+<head>
+  <script src="/deploy-config.js"></script>
+  <!-- Global site tag (gtag.js) - Google Analytics -->
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-WCDTMFM6RG"></script>
+  <script>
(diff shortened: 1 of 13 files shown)


packages/client/public/index.html
```diff
@@ -1,26 +1,33 @@
...
+</head>
+
+<body>
+  <noscript>
+    <strong>We're sorry but <%= htmlWebpackPlugin.options.title %> doesn't work properly without JavaScript enabled.
+        Please enable it to continue.</strong>
+  </noscript>
+  <div id="app"></div>
+</body>
+
```
TylerHendrickson, line 32: Note: modifications from lines 6-32 are just formatting.

packages/client/public/index.html
```diff
@@ -1,26 +1,33 @@
...
-    <link rel="stylesheet" href="https://unpkg.com/vue-multiselect@2.1.0/dist/vue-multiselect.min.css">
-  </head>
-  <body>
-    <noscript>
-      <strong>We're sorry but <%= htmlWebpackPlugin.options.title %> doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
-    </noscript>
-    <div id="app"></div>
-  </body>
+<head>
+  <script src="/deploy-config.js"></script>
```
as1729, line 5: Where is this file defined? I don't see it in the repository anywhere or this PR. Am I missing something?

TylerHendrickson: Yep, the file is not defined in this repository because its contents are specific to the deployment environment; its absence is what enables builds that are agnostic to their runtime configuration. 

I considered adding a "dummy" version of the file defined in the repo, which would simply get overwritten at deploy-time, but I couldn't think of a strong argument for doing so and worried it would be even more confusing to serve a different file than what's tracked in the repo.

as1729 (approved):
Thanks for the comments! The changes look good!

packages/client/public/index.html
TylerHendrickson: Noted! This PR auto-merged, but I will include an example file as a side-addition in my upcoming PR to adds a website build pipeline. Thanks for the suggestion 👍

PR #676, opened 2022-12-06
feat: adds ability to serve a digest email

as1729:
### Description
This PR does a few things: 
1. Adds the grants digest email
2. Re-uses components across multiple emails
3. Ensures all emails have consistent styling
4. Ensures there's a placeholder for the notification management changes in the front-end

### Screenshots / Demo Video
Demoed during show and tell
https://www.notion.so/usdr/Show-Tell-d4ba5e125d2b49df844eda046d298fd3

### Testing
- Added unit tests and integration tests

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -285,7 +285,7 @@ function deleteKeyword(id) {
 }
 
 async function getNewGrantsForAgency(agency) {
-    console.log(`${agency.name}`);
+    console.log(`Getting grants for ${agency.name}`);
     const query = knex(TABLES.grants)
         .limit(5);
     const result = await query;
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -289,10 +289,12 @@ async function getNewGrantsForAgency(agency) {
     const agencyCriteria = await getAgencyCriteriaForAgency(agency.id);
 
     const rows = await knex(TABLES.grants)
+        .select(knex.raw(`${TABLES.grants}.*, count(*) OVER() AS total_grants`))
         .modify(helpers.whereAgencyCriteriaMatch, agencyCriteria)
         .modify((qb) => {
             qb.where({ open_date: moment().subtract(1, 'day').format('YYYY-MM-DD') });
-        });
+        })
+        .limit(3);
 
     return rows;
 }
@@ -633,8 +635,8 @@ async function getAgenciesSubscribedToDigest() {
             'agencies.abbreviation',
             'agencies.code',
         )
-        .from(TABLES.agencies)
-        .join(TABLES.keywords, 'keywords.agency_id', '=', 'agencies.id')
+        .from('agencies')
+        .join('keywords', 'keywords.agency_id', '=', 'agencies.id')
         .orderBy('agencies.id')
         .groupBy(
             'agencies.id',
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -622,7 +622,7 @@ async function getAgenciesByIds(agencyIds) {
 
 async function getAgenciesSubscribedToDigest() {
     const query = knex.select('agencies.*')
-        .from(TABLES.agencies);
+        .from(TABLES.agencies).join(TABLES.keywords, 'keywords.agency_id', '=', 'agencies.id');
     const result = await query;
 
     return result;
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -621,8 +621,22 @@ async function getAgenciesByIds(agencyIds) {
 }
 
 async function getAgenciesSubscribedToDigest() {
-    const query = knex.select('agencies.*')
-        .from(TABLES.agencies).join(TABLES.keywords, 'keywords.agency_id', '=', 'agencies.id');
+    const query = knex
+        .select(
+            'agencies.id',
+            'agencies.name',
+            'agencies.abbreviation',
+            'agencies.code',
+        )
+        .from(TABLES.agencies)
+        .join(TABLES.keywords, 'keywords.agency_id', '=', 'agencies.id')
+        .orderBy('agencies.id')
+        .groupBy(
+            'agencies.id',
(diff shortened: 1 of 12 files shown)


packages/server/src/lib/email.js
```diff
@@ -123,11 +186,74 @@ async function sendGrantAssignedEmail({ grantId, agencyIds, userId }) {
     agencies.forEach((agency) => module.exports.sendGrantAssignedNotficationForAgency(agency, grantDetail, userId));
 }
 
+async function sendGrantDigestForAgency(agency) {
+    const newGrants = await db.getNewGrantsForAgency(agency);
```
as1729, line 190: Add some additional logging here for important events.

packages/server/src/lib/email.js
```diff
@@ -123,11 +186,74 @@ async function sendGrantAssignedEmail({ grantId, agencyIds, userId }) {
...
+
+    if (newGrants.length === 0) {
+        return undefined;
+    }
+
+    const grantDetails = [];
+    newGrants.slice(0, 3).forEach((grant) => grantDetails.push(module.exports.getGrantDetail(grant)));
```
as1729, line 197: Check with knex if it can get total count while getting the first 3 results only

as1729: make some screenshots of metrics dashboard in render to get an idea for typical usage and memory usage.

packages/server/src/lib/email.js
```diff
@@ -123,11 +186,74 @@ async function sendGrantAssignedEmail({ grantId, agencyIds, userId }) {
...
+        body_detail: `There are ${newGrants.length} new grants matching your agency's keywords and settings.`,
+        additional_body: additionalBody,
+    });
+
+    const emailHTML = module.exports.addBaseBranding(formattedBody, {
+        tool_name: 'Grants Identification Tool',
+        title: 'New Grants Digest',
+        notifications_url: 'mailto:grants-helpdesk@usdigitalresponse.org?subject=Unsubscribe&body=Please unsubscribe me from the grant digest notification email.',
+    });
+    const emailPlain = emailHTML;
```
as1729, line 221: Just leave this empty instead.

TylerHendrickson (changes requested):
(Changes discussed in call)

packages/server/src/db/index.js
```diff
@@ -284,6 +285,20 @@ function deleteKeyword(id) {
         .del();
 }
 
+async function getNewGrantsForAgency(agency) {
+    const agencyCriteria = await getAgencyCriteriaForAgency(agency.id);
+
+    const rows = await knex(TABLES.grants)
+        .select(knex.raw(`${TABLES.grants}.*, count(*) OVER() AS total_grants`))
```
as1729, line 292: @TylerHendrickson thanks for pushing on this! I was able to ensure that we only returned at most 3 grants while also getting a count of the rows directly from postgres in a single query.

I implemented it as a window-function and will take a next step once we are using the feature to analyze the query a bit more to see how postgres is processing this.

packages/server/src/lib/email.js
as1729: Done

as1729:
@TylerHendrickson I added 3 main changes:
- Simple plaintext parser using regex. This isn't great and I added a TODO to revisit this in the next PR. But for now its trending in the right direction.
- Added important event logs
- Ensured that we retrieve grant totals in a single query so we aren't overloading application memory with unnecessary grants data.

packages/server/src/lib/email.js
```diff
@@ -159,7 +160,9 @@ async function sendGrantAssignedNotficationForAgency(assignee_agency, grantDetai
         title: 'Grants Assigned Notification',
         notifications_url: 'mailto:grants-helpdesk@usdigitalresponse.org?subject=Unsubscribe&body=Please unsubscribe me from the grant assigned notification email.',
     });
-    const emailPlain = emailHTML;
+
+    // TODO: add plain text version of the email
+    const emailPlain = emailHTML.replace(/<[^>]+>/g, '');
```
TylerHendrickson, line 165: This is a clever solution for the interim. I'd replace the matches here with a single space so that words/sentences that are separated by newlines aren't smooshed together.

Example:
```
Some information is here.
And this is also useful.
```
Should output as `Some information is here. And this is also useful.` rather than `Some information is here.And this is also useful.`

packages/server/src/lib/email.js
```diff
@@ -187,29 +190,37 @@ async function sendGrantAssignedEmail({ grantId, agencyIds, userId }) {
...
-    if (newGrants.length > 3) {
+    if (newGrants[0].total_grants > 3) {
         const additionalButtonTemplate = fileSystem.readFileSync(path.join(__dirname, '../static/email_templates/_additional_grants_button.html'));
         additionalBody += mustache.render(additionalButtonTemplate.toString(), { additional_grants_url: process.env.WEBSITE_DOMAIN });
     }
 
     const formattedBody = mustache.render(formattedBodyTemplate.toString(), {
         body_title: 'New grants have been posted',
-        body_detail: `There are ${newGrants.length} new grants matching your agency's keywords and settings.`,
+        body_detail: `There are ${newGrants[0].total_grants} new grants matching your agency's keywords and settings.`,
```
TylerHendrickson, line 223: IMO it's fine for now (I think people are generally used to this type of "default plural" behavior), but as we add more and more emails, it would be nice to have some kind of conventional/parameterized way of conditionally formatting plural/singular words so that we can avoid errant plurals in cases when the content is singular (e.g. "There is 1 new grant matching..." instead of "There are 1 new grants matching..."). 

Perhaps mustache has something akin to Django's [pluralize filter](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#pluralize)?

TylerHendrickson (approved):
@as1729 Looks good to me! I added a few notes for your consideration, but nothing I think is a show-stopper. 

Great work putting this together! 👏 👏 👏

Issue #680, opened 2022-12-08
[ARPA P&E Reporting] Rhode Island is unable to download the Audit Report for FY23Q1


usdr-grants:
**Submitted Description:**
When attempting to download the audit report for the period of FY23Q1 in Rhode Island's Render Site - https://grants.usdigitalresponse.org/arpa_reporter/

When attempting to download this audit report, instead of the audit report I receive a file called audit_report.htm, in which can now be opened/viewed.

**(Optional) Uploaded file - click to download:**
https://www.jotform.com/uploads/grants.dev/212007960331041/5463386484525592032/Screenshot%202022-12-08%20145636.png

PR #687, opened 2022-12-08
#680 Fix null rules error that is blocking audit report generation

trumanc:
### Ticket #
https://github.com/usdigitalresponse/usdr-gost/issues/680

### Description
Fix for a bug reported by RI saying they are unable to download the audit report currently.
I was able to repro the bug and saw an error message in the 500 response coming back from the audit_report endpoint:
Screen Shot 2022-12-08 at 2 36 13 PM

I looked for anywhere in code that reads the persistentFormatters field in the audit report codepath, and made sure to handle null values more gracefully.

I also added logging to the endpoints error handing so that in addition to sending that error message, it will log the entire error (including the callstack) so that we can examine it more thoroughly.

I also cleaned up some issues with the formatter methods that were generating unnecessary logs that occurred when string methods were applied to values that were instead parsed as numbers (e.g. zipcodes). These errors were being caught, logged, and ignored, and they were making up an overwhelming majority of the log volume.

### Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/services/records.js
+++ b/packages/server/src/arpa_reporter/services/records.js
@@ -125,6 +125,10 @@ async function loadRecordsForUpload (upload) {
       const formattedRow = {}
       Object.keys(row).forEach(fieldId => {
         let value = row[fieldId]
+        if (rulesForCurrentType[fieldId] === null) {
+          // No known rules for this type, so we can't format it.
+          return
+        }
         for (const formatter of rulesForCurrentType[fieldId].persistentFormatters) {
           try {
             value = formatter(value)
--- a/packages/server/src/arpa_reporter/services/validation-rules.js
+++ b/packages/server/src/arpa_reporter/services/validation-rules.js
@@ -1,12 +1,13 @@
 
 const srcRules = require('../lib/templateRules.json')
+const _ = require('lodash');
 
 const recordValueFormatters = {
   makeString: (val) => String(val),
-  trimWhitespace: (val) => val ? val.trim() : val,
-  removeCommas: (val) => val.replace(/,/g, ''),
-  removeSepDashes: (val) => val.replace(/^-/, '').replace(/;\s*-/g, ';'),
-  toLowerCase: (val) => val.toLowerCase()
+  trimWhitespace: (val) => _.isString(val) ? val.trim() : val,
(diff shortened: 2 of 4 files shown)


packages/server/src/arpa_reporter/services/validation-rules.js
```diff
@@ -1,12 +1,16 @@
 
 const srcRules = require('../lib/templateRules.json')
 
+function isStr(val) {
+  return val && typeof val === 'string'
+}
```
TylerHendrickson, line 6: I think it's fine either way, but just wanted to mention that we have `lodash` available for use in this package, which also provides an [`_.isString(value)` function](https://lodash.com/docs/#isString). I think the (sole?) advantage in this case would be having one less thing unit to track in coverage metrics 🤷

packages/server/src/arpa_reporter/services/records.js
```diff
@@ -125,6 +125,10 @@ async function loadRecordsForUpload (upload) {
       const formattedRow = {}
       Object.keys(row).forEach(fieldId => {
         let value = row[fieldId]
+        if (rulesForCurrentType[fieldId] === null) {
+          // No known rules for this type, so we can't format it.
+          return;
+        }
```
ajhyndman, line 131: Cool, thanks for fixing this!

I guess now that we are reading field IDs from the sheet, they don't necessarily all exist in the committed ruleset.  Another oversight of mine.  😬

ajhyndman (approved):

trumanc:
@TylerHendrickson I've updated the PR to use the lodash isString method.
I also agree that it would be good to get some test coverage on this file, but I'm going to be traveling for the next few days and won't have time to work on that before our next release (it ended up being a bigger task than expected because we don't have any of the test files set up for the audit report code paths at all yet). Up to you to decide if it's worth delaying merging this fix until the test coverage is there, or if we want to merge it and deal with test coverage later

TylerHendrickson:
@trumanc I don't think this needs to get held up until our next release, so I think this is okay to merge. I do think it would be good to circle back and get some more coverage on that module, though.

FYI, I went go ahead and add a simple set of table/spec tests that just target the member functions of the `recordValueFormatters` map – feel free to keep or replace whenever you come back to this, though.

TylerHendrickson (approved):

Issue #413, opened 2022-09-28
[ID] Add anti-keywords - way to exclude certain keywords from search.


mhuang-usdr:
Requested by a partner

mhuang-usdr:
Mock of what the Keywords page should look like:

<img width="554" alt="image" src="https://user-images.githubusercontent.com/91221265/194423951-b1c5ba20-2358-4a9c-90b6-ca100f97cf67.png">

slants3:
Taking a look at this.

It looks like we insert all keywords into a `keywords` table and then also store its functionality (to add all terms that match said keyword) and the related agency. My thinking at the moment is to continue to use this table and add new ways to interface with it to introduce anti-keywords.

It looks like there was a mass add event on 2/18 and there haven't been new keywords added since? Wanted to see if this aligns with what you expect @mhuang-usdr

mhuang-usdr:
I don't remember a mass add event, we don't really add new keywords on our
users behalf. They usually add keywords themselves - we should have some
new ones over the last few months.

slants3:
Ah you're totally right, I missed the latest entries!

slants3:
Some notes...
1. Not entirely sure what the `mode` column is used for on the `keywords` table. It looks like it was used during development but discarded when real users began entering their keywords around August of last year. Looks like @Rapol was the last person to touch this as part of grantscraper
2. Is anti-keyword an intuitive name for future devs? Looks like the standard is negative keywords, though that is also slightly confusing
3. Currently debating between adding a new column to the keywords table or creating an entirely new anti-keywords table all together. Love to get input if anyone has any @TylerHendrickson @as1729

slants3:
Looks like this should be working and is live in staging. Closing as of now.

PR #693, opened 2022-12-12
413: Separate Keywords Functionality from Grants Store

slants3:
As we look to add new `anti-keywords` as a part of #413, I motion to separate the `keywords` store module from the larger `grants` module. While keywords themselves are intrinsically tied to grants - they assist in the grant name search process - I believe adding the new `anti-keywords` pattern could overload the `grants` module which already handles grants lookup and adding, agencies related to grants, marking grants as interested, exporting grant CSVs and more.

Also, the `keywords` functionality is more or less already distinct - the API endpoints are already namedspaced appropriately for keywords - and given how small of a task it is to separate the two, I choose to do so here.

### DIFF
--- a/packages/client/src/components/Modals/AddKeyword.vue
+++ b/packages/client/src/components/Modals/AddKeyword.vue
@@ -81,7 +81,7 @@ export default {
   },
   methods: {
     ...mapActions({
-      createKeyword: 'grants/createKeyword',
+      createKeyword: 'keywords/createKeyword',
     }),
     resetModal() {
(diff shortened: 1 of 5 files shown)


as1729:
I'm in favor of this change, could you post a few screenshots or a quick loom video on the PR indicating that the functionality of these pages have not changed? This is mostly as a workaround for the lack of tests on the client side.

as1729 (approved):

TylerHendrickson (approved):

Issue #378, opened 2022-07-28
[ARPA] Create an extensible framework for enforcing conditionally required values


trumanc:
There are a number of fields in the input template that are sometimes required, depending on the answer to a previous question.

We should programmatically detect these conditionally required values, and validate that if they are required, the value is present.

mhuang-usdr:
This is causing partners a lot of issues ahead of the Jan'23 reporting cycle

trumanc:
The first and most critical use-case of this feature has been accomplished in these two PRs: https://github.com/usdigitalresponse/usdr-gost/pull/740 https://github.com/usdigitalresponse/usdr-gost/pull/761

The point of those PRs is that there were around 30 fields that are usually required, but become optional if the project is in status 'Not started'. While we've handled those 30 fields, this idea applies to several other fields. Here are some examples:
* `Cancellation_Reason__c` is required if `Completion_Status__c` is `Cancelled`
* `Total_Cost_Capital_Expenditure__c` and `Type_of_Capital_Expenditure__c` are required if `Does_Project_Include_Capital_Expenditure__c` is 'Yes'
* `Type_of_Capital_Expenditure_Other__c` is required if the answer to `Type_of_Capital_Expenditure__c` is `Other (specify)`
* Many many many others

The details of when/how to determine if these fields are required is usually found in the workbook in the hidden row 5 which gives definitions for all the input columns. A good place to check for possibly conditional fields is https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/server/src/arpa_reporter/lib/outputTemplates.json where you can see all the fieldIds that include `"required": "Conditional"`


I'm not sure how useful a validation library would be here, because our validation system has a strong dependency on the inpout workbook. All the rules you see in outputTemplates.json is parsed directly from data found in the excel workbook, in order to try to ensure the workbook and the backend validations are as in-sync as possible. It might be possible to use a tool to improve it, but given the volume of code that already exists, I'm not sure if the benefits of adding it would outweigh the cost of having to move our current validation code to that system

as1729:
Not necessary for the Monday March 20th deadline.

as1729:
@trumanc let us know if you will be working on this while onboarding into the elections team or if its better to reassign.

trumanc:
This ticket has been completed! There is some follow-up work in ticket #1074 that builds on this

PR #740, opened 2022-12-23
#378: First pass at conditionally required fields in ARPA validator

trumanc:
### Ticket #
https://github.com/usdigitalresponse/usdr-gost/issues/378

### Description
Right now, the validator backend can only really handle fields that are either always required for the particular EC code, or always fully optional. This causes a problem because many fields are actually "conditional": They may be required depending on values added for other fields in the row.

The most common form of this condition is that many fields are not required if the project status is 'Not started', but otherwise must be entered.

This PR creates a system that lets us configure a list of field ids that should include this new check: If we see that the data for the field is empty, and then we see that this field is conditional, and then we see that the conditional check says the data is required, then it will add a validation error. I've started by only turning this on for ~2~ *3* particular fields ids (Primary_Project_Demographics__c and Number_Students_Tutoring_Programs__c. *I also added Does_Project_Include_Capital_Expenditure__c as an important test case*) to allow us to test this in a way that is less dangerous than starting it with all ~20 ids.  

### Screenshots / Demo Video
In a test file, I have two projects in EC code 2.25, where the only data set is the project status.
For the project on line 13, the status is 'Completed'
For the project on line 14, the status is 'Not started'.

The only difference in the validation errors when I run it with this new code, is that the line 13 project has two additional errors: 
Screen Shot 2022-12-22 at 4 04 09 PM

Note: This feature depends very strongly on an assumption that making a field id conditionally required in one place means that it will be conditionally required in all places that we ask that question. If this turns out to be untrue, then we need to update the system to configure these functions by ec AND field-id, not just by field-id.

More details on that assumption in https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1671751915108879?thread_ts=1671740558.980399&cid=C031R1Y49KL

### Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/services/validate-upload.js
+++ b/packages/server/src/arpa_reporter/services/validate-upload.js
@@ -246,6 +246,14 @@ async function validateRecord ({ upload, record, typeRules: rules }) {
           `Value is required for ${key}`,
           { col: rule.columnName, severity: 'err' }
         ))
+      } else if (rule.required === 'Conditional') {
+        if (rule.isRequiredFn && rule.isRequiredFn(record)) {
+          errors.push(new ValidationError(
+            // This message should make it clear that this field is conditionally required
+            `Based on other values in this row, a value is required for ${key}`,
+            { col: rule.columnName, severity: 'err' }
+          ))
+        }
       }
 
     // if there's something in the field, make sure it meets requirements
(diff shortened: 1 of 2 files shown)


as1729 (approved):

trumanc:
UpdateL I added a third field id (Does_Project_Include_Capital_Expenditure__c) so that we can test that field since we know it has interestting changes in behavior across ECs.

Issue #781, opened 2023-01-10
docker-compose up fails: Error: Could not find any Python installation to use


NatHillardUSDR:
On current `_staging`, if you run `docker compose up --build`, you get the below error:

```
...
 => ERROR [usdr-gost-frontend app_base 7/7] RUN ["yarn", "run", "setup"]                                                           39.4s 
------                                                                                                                                   
 > [usdr-gost-frontend app_base 7/7] RUN ["yarn", "run", "setup"]:                                                                       
#0 0.455 yarn run v1.22.17                                                                                                               
#0 0.504 $ yarn install && yarn run bootstrap                                                                                            
#0 0.763 [1/5] Validating package.json...                                                                                                
#0 0.772 [2/5] Resolving packages...                                                                                                     
#0 1.234 [3/5] Fetching packages...
#0 24.82 [4/5] Linking dependencies...
#0 37.25 [5/5] Building fresh packages...
#0 38.30 info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
#0 38.30 error /app/node_modules/@parcel/watcher: Command failed.
#0 38.30 Exit code: 1
#0 38.30 Command: node-gyp-build
#0 38.30 Arguments: 
#0 38.30 Directory: /app/node_modules/@parcel/watcher
#0 38.30 Output:
#0 38.30 gyp info it worked if it ends with ok
#0 38.30 gyp info using node-gyp@9.1.0
#0 38.30 gyp info using node@16.14.0 | linux | arm64
#0 38.30 gyp ERR! find Python 
#0 38.30 gyp ERR! find Python Python is not set from command line or npm configuration
#0 38.30 gyp ERR! find Python Python is not set from environment variable PYTHON
#0 38.30 gyp ERR! find Python checking if "python3" can be used
#0 38.30 gyp ERR! find Python - "python3" is not in PATH or produced an error
#0 38.30 gyp ERR! find Python checking if "python" can be used
#0 38.30 gyp ERR! find Python - "python" is not in PATH or produced an error
#0 38.30 gyp ERR! find Python 
#0 38.30 gyp ERR! find Python **********************************************************
#0 38.30 gyp ERR! find Python You need to install the latest version of Python.
#0 38.30 gyp ERR! find Python Node-gyp should be able to find and use Python. If not,
#0 38.30 gyp ERR! find Python you can try one of the following options:
[...]

PR #782, opened 2023-01-10
[DRAFT] fix: docker-compose up fails: error: Could not find any python instal…

NatHillardUSDR:
…lation to use (781)

Docker compose up is currently failing because of missing dependencies. This adds in the necessary dependencies.
I do think larger Docker and Docker-compose changes are probably warranted here, but keeping this fix small for now.

### Ticket #
#781 

### Description
Docker compose up is currently failing because of missing dependencies. This adds those in. It is a bit odd that a js dependency requires python, but apparently it does! 

### Screenshots / Demo Video
See error message from original issue

### Testing
Successfully ran `docker compose up` and did not see a failure. I did not test deployment scenarios as I'm not sure how to do this just yet, but this is definitely worth a check!

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /app
 COPY . .
 RUN ["mkdir", "-p", "/root/.config/yarn"]
 RUN ["chmod", "-R", "777", "/root"]
-RUN ["apk", "--update", "add", "bash"]
+RUN ["apk", "--update", "add", "bash", "python3", "g++", "make", "py3-pip"]
 RUN ["yarn", "run", "setup"]
 
 FROM app_base as client


TylerHendrickson:
@NatHillardUSDR I didn't see mentioned anywhere- which JS dependency requires Python? Also, is it a dependency for runtime, build-time, or both?

NatHillardUSDR:
A very good question, @TylerHendrickson ! 

https://github.com/usdigitalresponse/usdr-gost/issues/781 has the log, but its failing during the build phase for `@parcel/watcher`, because *this* depends on `node-gyp-build`, which, according to [its docs](https://github.com/nodejs/node-gyp#configuring-python-dependency) uses `python` under the hood. 

Specifically the dependencies go:
```
npm ls node-gyp-build 
cares-grant-opportunities@1.0.0 /Users/nhillard/Documents/Code/USDR/usdr-gost
└─┬ lerna@5.6.2
  └─┬ nx@14.8.3
    └─┬ @parcel/watcher@2.0.4
      └── node-gyp-build@4.5.0
```

This, in turn, is because `yarn setup` is `yarn install && yarn run bootstrap`, and `bootstrap` is `lerna bootstrap` in `package.json` .

Arguably here, `lerna bootstrap` is not necessary, as we're not relying on the inter-package dependencies that `lerna` optimizes for. This is maybe a bigger change though and potentially has other fallout. 

As a shorter-term workaround I guess the above works, but maybe I can add a comment about the somewhat convoluted path that led to this being necessary

TylerHendrickson:
@NatHillardUSDR The missing piece for me is still understanding what specific dependency change is now requiring this. I don't fundamentally have an issue with adding Python to the build deps, but I'd like to first understand how the need became introduced. As you noted, there's something of a convoluted dependency chain here, and I'm wondering whether this issue ought to be resolved through simplifying some of these needs.

NatHillardUSDR:
> @NatHillardUSDR The missing piece for me is still understanding what specific dependency change is now requiring this. I don't fundamentally have an issue with adding Python to the build deps, but I'd like to first understand how the need became introduced. As you noted, there's something of a convoluted dependency chain here, and I'm wondering whether this issue ought to be resolved through simplifying some of these needs.

I can look into removing `lerna bootstrap` - based on [the docs](https://github.com/lerna/lerna/tree/main/commands/bootstrap) for this command it seems to not have much benefit, in our scenario, over `yarn install`. I can test this out and see how it might impact things

LieAlbertTriAdrian (approved):

NatHillardUSDR:
Moving back to draft to consider alternative fix for this, likely involving removing lerna dep in this particular instance

TylerHendrickson:
Removing myself as a reviewer while this is in draft state. Please feel free to re-add me once you're ready for another review!

PR #792, opened 2023-01-11
413: Add logic for creating and display excluded keywords

slants3:
This PR introduces the logic for excluded keywords.

We use the a psql specific pattern matching expression of `!~*` to remove grants with descriptions which include the specific pattern.

Keywords with `null` type default to `include`.

image


packages/client/src/views/Keywords.vue
```diff
@@ -1,22 +1,36 @@
...
-  <AddKeywordModal :showModal.sync="showAddKeywordModal"/>
+    <b-row>
+        <b-col><h3>Include results with:</h3></b-col>
+        <b-col class="d-flex justify-content-end">
+            <div>
+                <b-button variant="success" @click="openAddKeywordModal">Add</b-button>
+            </div>
+        </b-col>
+    </b-row>
+    <b-table sticky-header="600px" hover :items="includeKeywords" :fields="fields">
```
as1729, line 11: Add `<hr />`

packages/server/src/db/index.js
```diff
@@ -176,7 +176,18 @@ async function getAgencyCriteriaForAgency(agencyId) {
 
     return {
         eligibilityCodes: enabledECodes.map((c) => c.code),
-        keywords: keywords.map((c) => c.search_term),
```
as1729, line 179: Can we add test coverage to this method?

packages/client/src/views/Keywords.vue
as1729, line 11: Can we add created_at column?

packages/client/src/views/Keywords.vue
```diff
@@ -35,12 +50,31 @@ export default {
...
+        {
+          key: 'created_at',
+          thStyle: { width: '20%' },
+          formatter: (value) => {
+            const date = new Date(value);
+            const month = date.getMonth() + 1;
+            const day = date.getDate();
+            const year = date.getFullYear();
+
+            return `${month}/${day}/${year}`;
```
as1729, line 68: Could be simplified to-
```js
return moment(value, 'YYYY-MM-DD').format('M/DD/YYYY');
```

In line 43 you'll need to add `import moment from 'moment';`

as1729 (commented):
@slants3 the tests look good to me- can you add some screenshots of the current experience?

packages/client/src/views/Keywords.vue
slants3: Good call!

as1729 (approved):

as1729 (approved):

Issue #785, opened 2023-01-10
getNewGrantsForAgency test fails when run locally


NatHillardUSDR:
# Summary
If you run `yarn test` against a locally-running instance of the current `_staging`, the `getNewGrantsForAgency` test fails with the below error:

```
    getNewGrantsForAgency
      ✔ returns zero grants if no grants match criteria and opened yesterday
      1) returns zero grants if no grants match criteria and opened yesterday


  26 passing (111ms)
  1 failing

  1) db
       getNewGrantsForAgency
         returns zero grants if no grants match criteria and opened yesterday:

      AssertionError: expected +0 to equal 1
      + expected - actual

      -0
      +1
      
      at Context.<anonymous> (__tests__/db/db.test.js:247:38)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)



error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This does not occur when you run `docker compose exec app yarn test`.

This appears to be related to [this line/test](https://github.com/usdigitalresponse/usdr-gost/blob/9074b204be4ffd727474227b0205e8694b46fec7/packages/server/__tests__/db/db.test.js#L241) in particular. 

It may be that this is expected for a local environment as opposed to Docker, but figured I'd file in case.

NatHillardUSDR:
Found it! This is timezone related. 

In our tests we do 
```
this.clockFn = (date) => sinon.useFakeTimers(new Date(date));
this.clock = this.clockFn('2022-06-22');
```

which for me in EST is :
```
GMT: Wednesday, June 22, 2022 12:00:00 AM
Your time zone: Tuesday, June 21, 2022 8:00:00 PM GMT-04:00 DST
Relative: 7 months ago
```
or, using `moment()`, `Tue Jun 21 2022 20:00:00 GMT-04001`.

When we do 
```
moment().subtract(1, 'day').format('YYYY-MM-DD')
```
it then yields `'2022-06-20'` , because in my timezone (UTC-4) this is actually "two" days before, but we're expecting the static string `2022-06-21`. 

The fix is to either make this dynamic or to decouple one or both sides from timezones

PR #793, opened 2023-01-11
fix: getNewGrantsForAgency test fails when run locally (#785)

NatHillardUSDR:
The getNewGrantsForAgency test was failing locally for machines with different timezones due to a hard-coded date string. This fixes date math by moving to use Date objects instead.

### Ticket #
#785

### Description
See above - due to timezone differences, a hard-coded date string meant to represent one day ago was in fact two days ago for anything less than GMT. Using `Date` for the desired end-date as well resolves this. 

### Screenshots / Demo Video
```
x  yarn test
 getNewGrantsForAgency
      ✔ returns zero grants if no grants match criteria and opened yesterday
      ✔ returns a grant whose modification date is one day ago
```

### Testing
Run `yarn test` with a locally-running setup, verify that `getNewGrantsForAgency` passes. Run `docker compose exec app yarn test` ensure that docker-run tests also still pass. 

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/db/db.test.js
+++ b/packages/server/__tests__/db/db.test.js
@@ -238,10 +238,11 @@ describe('db', () => {
             const result = await db.getNewGrantsForAgency(fixtures.agencies.accountancy);
             expect(result.length).to.equal(0);
         });
-        it('returns zero grants if no grants match criteria and opened yesterday', async () => {
+        it('returns a grant whose modification date is one day ago', async () => {
             const newGrant = fixtures.grants.healthAide;
             newGrant.grant_id = '444816';
-            newGrant.open_date = '2022-06-21';
+            // Note the use of `Date` -- this ensures compatibility with our mocked time
+            newGrant.open_date = new Date('2022-06-21');
             await knex(TABLES.grants).insert(Object.values([newGrant]));
             const result = await db.getNewGrantsForAgency(fixtures.agencies.accountancy);
             expect(result.length).to.equal(1);


as1729 (approved):
Thanks for fixing this!

NatHillardUSDR:
Thanks @as1729  for the review! 
I'm wondering, the `Build Docker image for GOST API / build-and-push (pull_request)` was originally "skip"d for this job, so I re-ran, but just now it was cancelled (perhaps I shouldn't have re-run?) 

Is it expected that this job should not be running for this PR? I could hit "squash and merge", but I'm seeing "Some checks were not successful" as the current status still -- should I merge anyway, and/or are there changes we can make in the future to ensure such jobs don't run for certain fixes (e.g. fixes to `__tests__`?

as1729:
Let me get @TylerHendrickson 's eyes on this one. I'm pretty sure the Build Docker Image job is for merges from `_staging` to `main` and not feature branches that get merged into `_staging`.

Can you confirm Ty?

TylerHendrickson:
@as1729 The "Build Docker image for GOST API" workflow is configured to run when changes are pushed to `_staging` and `main`, as well as on pull request events. However, the step that actually publishes the built docker image is meant to be blocked for PRs, so that it essentially makes sure that the image _can_ be built as a QA check, but doesn't actually send the built image anywhere until the changes are merged.

The `push` parameter defined on line 60 of the workflow, `push: ${{ github.event_name != 'pull_request' }}`, is meant to prevent publishing for unmerged (still in PR) changes, but I'm wondering if `github.event_name` was something different when @NatHillardUSDR manually re-ran. I'll look into this.

TL;DR It's fine if the workflow runs during a PR- there should be no need to cancel, as it's meant to build the image but not publish it anywhere. However, we should check to make sure that the `${{ github.event_name != 'pull_request' }}` conditional is working as-intended when the workflow is manually executed (if not, it probably needs to be inverted so that it only evaluates to `true` when the event is `"push"` or `"release"`).

TylerHendrickson:
Update: It looks like everything is working as-intended. You can see [here](https://github.com/usdigitalresponse/usdr-gost/actions/runs/3894581181/jobs/6648767819#step:7:4) that the `push` parameter for the `docker/build-push-action@v3` action evaluated to `false` as it should. 

@NatHillardUSDR I'll leave it to you to go ahead and re-run the workflow. Should be fine to let it run to completion (as stated, it will not actually publish anything), and then Github will allow you to merge.

NatHillardUSDR:
I wanted to follow up on this as part of this comment thread, as I hit this with another PR just now as well - it seems that the two build jobs are being triggered simultaneously, and this is resulting in github auto-canceling one of them. I filed an issue with a bit more detail about this here: https://github.com/usdigitalresponse/usdr-gost/issues/821 . 

In short, @TylerHendrickson , I was initially confused by your comments referencing a cancelation, as I had not myself canceled the build job! Looking into this, I realized that in fact github is auto-canceling due to the other job being higher priority. I'm not sure exactly how to fix this but figured I'd note down my findings so we can address later.

Issue #832, opened 2023-01-19
[ARPA] Application keeps timing out when audit report is generated


as1729:
https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1674139190721159

as1729:
Nobody has any issues with this anymore!

PR #833, opened 2023-01-19
ARPA: Add column and script for caching workbook json data

trumanc:
### Ticket #
First step on ticket #832 

### Description
This PR begins the process of using cached data so that we can avoid the expensive excel parsing step, especially for tasks that read a huge number of files (like generating the audit report).

This PR adds the db migration to add a new column to the db, specifically of the jsonb (or json-binary) type.
This type is a little slower to write to the database, but faster to read out because it is alreadd parsed into a valid json shape.

This PR punts on the questions of when we should use/avoid the cache and when we need to invalidate data by not reading/writing to it during real user requests. It adds a manual script that can be run to cache parsed data for a single upload id. It then also does a comparison to see if the cached data exactly matches the parsed data. So far the only difference is in how timestamps are used (the cached data has converted the timestamp to a string, where the parsed data is using a native js time type).

### Screenshots / Demo Video
Screen Shot 2023-01-19 at 12 39 49 PM

After running the script, json data is showing in the database:
Screen Shot 2023-01-19 at 12 42 13 PM

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/migrations/20230119175045_add_parsed_data_blod_for_uploads.js
+++ b/packages/server/migrations/20230119175045_add_parsed_data_blod_for_uploads.js
@@ -1,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.schema
+        .table('uploads', (table) => {
+            table.jsonb('parsed_data');
+        });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function (knex) {
+    await knex.schema
+        .table('uploads', (table) => {
+            table.dropColumn('parsed_data');
+        });
+};
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -41,6 +41,7 @@
     "cron": "^1.8.2",
     "csv-stringify": "^6.0.5",
     "date-fns": "^2.23.0",
+    "deep-diff": "^1.0.2",
     "docx": "^7.3.0",
     "dotenv": "^16.0.0",
     "email-validator": "^2.0.4",
--- a/packages/server/src/arpa_reporter/db/uploads.js
+++ b/packages/server/src/arpa_reporter/db/uploads.js
@@ -97,6 +97,12 @@ async function createUpload (upload, trns = knex) {
   return inserted
 }
 
+async function cacheRecords(upload, records, trns = knex) {
+  return trns('uploads')
+    .where('id', upload.id)
+    .update({ parsed_data: JSON.stringify(records) })
+}
+
 async function setAgencyId (uploadId, agencyId, trns = knex) {
   return trns('uploads')
     .where('id', uploadId)
@@ -157,6 +163,7 @@ module.exports = {
   getUpload,
   uploadsInPeriod,
   uploadsInSeries,
+  cacheRecords,
   setAgencyId,
   setEcCode,
   markValidated,
--- a/packages/server/src/arpa_reporter/db/uploads.js
+++ b/packages/server/src/arpa_reporter/db/uploads.js
@@ -46,7 +46,10 @@ function getUpload (id, trns = knex) {
 }
 
 function usedForTreasuryExport (periodId, trns = knex) {
-  const tenantId = useTenantId()
+  return usedforTreasuryExportWithTenantId(useTenantId(), periodId, trns)
+}
+
+function usedforTreasuryExportWithTenantId(tenantId, periodId, trns = knex) {
   requiredArgument(periodId, 'periodId must be specified in validForReportingPeriod')
 
   return baseQuery(trns)
@@ -174,7 +177,8 @@ module.exports = {
   setEcCode,
   markValidated,
   markNotValidated,
-  usedForTreasuryExport
+  usedForTreasuryExport,
+  usedforTreasuryExportWithTenantId
 }
 
 // NOTE: This file was copied from src/server/db/uploads.js (git @ ada8bfdc98) in the arpa-reporter repo on 2022-09-23T20:05:47.735Z
--- a/packages/server/src/arpa_reporter/db/uploads.js
+++ b/packages/server/src/arpa_reporter/db/uploads.js
@@ -100,7 +100,13 @@ async function createUpload (upload, trns = knex) {
 async function cacheRecords(upload, records, trns = knex) {
   return trns('uploads')
     .where('id', upload.id)
-    .update({ parsed_data: JSON.stringify(records) })
+    .update({
+      parsed_data_cached_at: trns.fn.now(),
+      parsed_data: JSON.stringify(records,
+        // This resursively filters out any fields named 'parsed_data' in the records, which is
+        // important for ensuring we don't recursively/redundantly write this data to the db.
+        (k, v) => (k === 'parsed_data') ? undefined : v)
+    })
 }
 
 async function setAgencyId (uploadId, agencyId, trns = knex) {
--- a/packages/server/src/scripts/parse_and_cache_data.js
+++ b/packages/server/src/scripts/parse_and_cache_data.js
@@ -5,6 +5,7 @@ require("dotenv").config();
 
 const { loadRecordsForUpload } = require('../arpa_reporter/services/records')
 const _ = require("lodash");
+const { diff } = require('deep-diff');
 
 const inquirer = require("inquirer");
 
@@ -32,8 +33,6 @@ async function main() {
     ])
 
     console.log("Parsing and caching upload with id:", uploadId);
-
-
     const upload = await getUpload(uploadId)
     const records = await loadRecordsForUpload(upload)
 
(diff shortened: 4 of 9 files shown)


as1729 (approved):

TylerHendrickson:
For posterity, pasting below some options discussed in slack, In case write speed ends up being a concern:

I can think of a few ways to take advantage of the JSONB type without sacrificing write speed if we’re okay with not validating JSON correctness at time of insert/update:
1. Instead of just a jsonb column, create both a text column and a jsonb column (one that executes “later” after a write).
    -  We write directly to the text column
    -  A deferred trigger fires when the text column for a row is modified that converts the text value to jsonb and adds it to the jsonb-specific column. Something like this:
        ```sql 
        CREATE TRIGGER parsed_data_text_to_jsonb 
        AFTER INSERT ON uploads 
        FOR EACH ROW 
        WHEN (OLD.parsed_data IS DISTINCT FROM NEW.parsed_data)
        SET parsed_data_jsonb = parsed_data::jsonb
       ```
2. Create a materialized view that does something similar to the above (essentially a JSONB index that gets populated following writes to uploads table)
3. Create a generated column that does something similar to the above (I’m not sure whether generated columns are/can be deferred, so this might not work as a way to alleviate write speed issues).

packages/server/src/scripts/parse_and_cache_data.js
```diff
@@ -0,0 +1,109 @@
...
+    // Now that the data has been written, check that it matches what we parsed from excel
+    const updatedUpload = await getUpload(uploadId)
+
+    if (_.isEqual(records, updatedUpload.parsed_data)) {
+        console.log("Cached data is equivalent to parsed data")
+    } else {
+        diff = compare(records, updatedUpload.parsed_data)
+        console.log("ERROR: Cached and parsed data are not equivalent")
+        console.log(diff)
+    }
```
TylerHendrickson, line 51: Is this just a spot-check? Do you have a specific reason to believe that the data inserted into the DB might not match what was provided to the insert statement?

trumanc: Just a spot-check, but it has already proved useful: it turns out the json-ification means the timestamps are being treated differently after being read from the cache. The cached timestamps are strings now, where as the excel parser returns them as some other timestamp type

TylerHendrickson: Perhaps something like the [cryo library](https://github.com/hunterloftis/cryo) would be useful here, since it has more advanced (de)serialization support for native types but still ultimately serializes to valid JSON?

packages/server/migrations/20230119175045_add_parsed_data_blod_for_uploads.js
```diff
@@ -0,0 +1,21 @@
+/**
```
NatHillardUSDR, line 1: Shouldn't `blod` in the title be `blob`? (It's possible `blod` is a term of art I'm not familiar with?

packages/server/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "adm-zip": "0.5.9",
+    "async-batch": "1.1.2",
```
NatHillardUSDR, line 36: Presumably this is from the other pr to add `async-batch` and could be dropped from this one?

packages/server/src/arpa_reporter/db/uploads.js
```diff
@@ -97,6 +97,12 @@ async function createUpload (upload, trns = knex) {
   return inserted
 }
 
+async function cacheRecords(upload, records, trns = knex) {
+  return trns('uploads')
+    .where('id', upload.id)
+    .update({ parsed_data: JSON.stringify(records) })
```
NatHillardUSDR, line 103: I wonder actually if it makes sense to add a `cached_date` and/or `cached_version` property here to anticipate needing to do more advanced invalidation or update/change the json structure in the future (to Andrew's point in Slack?)

I'm not sure of best practices here but I've seen some combination of date storage and light cache versioning be potentially helpful for similar scenarios

packages/server/src/scripts/parse_and_cache_data.js
NatHillardUSDR: Ooh, this is very interesting and could cause other challenges also, in that it introduces a difference between serialization and deserialization...

I wonder if there's a way for knex and/or pg to set a parser for the data? I see `setTypeParser` on `pg` though this doesn't quite get there / is not this use case per se.

packages/server/src/scripts/parse_and_cache_data.js
```diff
@@ -0,0 +1,109 @@
...
+
+}
+
+// This method is copied verbatim from https://stackoverflow.com/a/41431685/21046021
+// and allows us to see every json path that differs between two objects
+var compare = function (a, b) {
```
NatHillardUSDR, line 57: This works, but I wonder about using a library like [deep diff](https://www.npmjs.com/package/deep-diff) ? I like the additional info this provides as well. It is yet another library, but it seems to handle this particular use case well and seems to have pretty good adoption overall (1mm downloads per week!)

yarn.lock
```diff
@@ -4449,6 +4449,11 @@ astral-regex@^2.0.0:
   resolved "https://registry.yarnpkg.com/astral-regex/-/astral-regex-2.0.0.tgz#483143c567aeed4785759c0865786dc77d7d2e31"
   integrity sha512-Z7tMw1ytTXt5jqMcOP+OQteU1VuNK9Y02uuJtKQ1Sv69jXQKKg5cibLwGJow8yzZP+eAc18EmLGPal0bp36rvQ==
 
+async-batch@1.1.2:
+  version "1.1.2"
+  resolved "https://registry.yarnpkg.com/async-batch/-/async-batch-1.1.2.tgz#ecac7dee0f02a1a3a846438ba7d3f96746239d79"
+  integrity sha512-tBR4YIKi7s1cCtNHuguyXVBdRSkS6N8OhM1czrpE0W2j2yovj/IAxXD187cNX0d+kGxBbZLPwW5ASonYuZJprA==
+
```
NatHillardUSDR, line 4456: Similar to the above comment, is this from the other patch?

NatHillardUSDR (approved):
Overall looks good, seems a straightforward implementation of an initial cache-like solution. 

I wonder about potentially introducing a basic notion of cached versioning and/or timestamping as a future-proofing mechanism - once we start caching it would be good to have some means of understanding when we initially did this. 

I also wonder about client-side caching within vuex. We may want to explore how this fares vs. this option for a subsequent iteration?

packages/server/src/scripts/parse_and_cache_data.js
trumanc: I think a quicker solution (and one that matches up with some processing we already have to do in places) is to add a "formatter" in our code that is run on the read-in data and ensures it is in the right format. 

We already have some of these in place to make sure certain fields are strings or numbers, depending on the field (for instance the zipcode looks like a number but has to be treated as a string in case it has a leading zero).

packages/server/migrations/20230119175045_add_parsed_data_blod_for_uploads.js
trumanc: whoops, I typo'd that. Will fix it now!

packages/server/package.json
trumanc: oh yeah, looks like a commit from Aditya's pr got merged into my branch somehow. I'll remove it.

packages/server/src/arpa_reporter/db/uploads.js
trumanc: yeah, a cached date here would be super useful. Especially so that if we have a breaking change in how data is parsed, we should be able to query for any cached versions from before that deploy and refresh them (or at least consider them invalid so they'll refresh manually later). I'll add that column!

packages/server/src/scripts/parse_and_cache_data.js
```diff
@@ -0,0 +1,109 @@
...
+        },
+    ])
+
+    console.log("Parsing and caching upload with id:", uploadId);
+
+
+    const upload = await getUpload(uploadId)
+    const records = await loadRecordsForUpload(upload)
+
+    await cacheRecords(upload, records)
```
ajhyndman, line 40: I do see some advantage to caching the records like this, but I feel like it would keep the system a good deal simpler if we just cache the whole workbook object.

That is to say, the expensive part of the audit report operation is this one line:

https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/server/src/arpa_reporter/services/records.js#L66-L70

If we abstract just that operation with an attempted cache read and fallback I don't think you need to tweak other parts of the record processors to accommodate running outside of the server context.

packages/server/src/scripts/parse_and_cache_data.js
ajhyndman: I would also prefer to declare a dependency over copy-pasting code.  👍

trumanc: sounds good, I'll update this!

yarn.lock
trumanc: yup, I'll remove this too

packages/server/src/scripts/parse_and_cache_data.js
trumanc: I'd be open to exploring that, but I have a couple major questions first:

1. Are we really certain that the other XLSX methods (including 4 separate calls to `XLSX.utils.sheet_to_json` are not also slow and expensive?
2. Since the fact we are calling `XLSX.utils.sheet_to_json` implies that the sheet isn't yet in a valid json form, then we presumably can't use the jsonb data type in the database column. What type would you prefer for this? Just a plain serialized string of the object?

I'm also worried that storing the full sheet will bog down our json data with a bunch of unnecessary formatting info and tons of data about hidden sheets.

packages/server/src/scripts/parse_and_cache_data.js
trumanc: neat! I'll explore this parser and make sure it fixes the diffs I'm seeing and then push it to this PR

trumanc:
@NatHillardUSDR, in response to your question about maybe using vuex for client-side caching:

I think that while that could help in some places in the project, it isn't well suited for this problem. This problem involves a very rare operation that has to be go back through all historical uploads and look at their data. The operation is so rare that we wouldn't ever expect a user to do the operation twice in a single web session (unless the first request failed). And the sheer volume of data that has to be processed makes it poorly suited for client-side handling, I think.

packages/server/src/scripts/parse_and_cache_data.js
ajhyndman: Good clarifying questions!

1. I can validate this, but I'm pretty confident that's what my profiling exposed.  XLSX.read does some gnarly, repeteitive, on-the-fly regex compilation to actually consume the Microsoft .xlsx file format.  Once it's been transformed to the XLSX (library) workbook format, it is JavaScript data and can be manipulated and transformed much more cheaply.
2. `XLSX.utils.sheet_to_json` is a poorly named function.  From the docs, this method "converts a worksheet object to an array of JSON objects."  The workbook object is itself absolutely valid JSON already.  I have a demo (and screenshot) of one of these objects in #842!

Your last point is definitely a big advantage to caching the parsed records.

I like the idea of keeping the our record logic as ignorant of the cache's existence as possible.  And I also like the idea of only having to invalidate the cache when we update the parameters passed to XLSX.read itself.   But those advantages definitely come at the cost of a somewhat larger cache volume.  A sample workbook file that I saved to disk as JSON was about 1 MB in size.  I haven't measured the equivalent records yet, but I'm sure they are a good deal smaller!

packages/server/src/scripts/parse_and_cache_data.js
ajhyndman: Note that I suspect you will need to use Cryo both to stringify _and_ parse to get the benefit of preserving date objects!

NatHillardUSDR:
> @NatHillardUSDR, in response to your question about maybe using vuex for client-side caching:
> 
> 
> 
> I think that while that could help in some places in the project, it isn't well suited for this problem. This problem involves a very rare operation that has to be go back through all historical uploads and look at their data. The operation is so rare that we wouldn't ever expect a user to do the operation twice in a single web session (unless the first request failed). And the sheer volume of data that has to be processed makes it poorly suited for client-side handling, I think.

This makes a lot of sense. I still need to better familiarize myself with how all of this fits together and what is happening where!

NatHillardUSDR:
> I did a little reading up on jsonb.  It looks to me like the main advantage the `jsonb` format affords is faster queries that rely on parsing the _content_ of the JSON data

Right - definitely this - I'm realizing I partially misunderstood the original issue here. I've used jsonb in the past to be able to query into the structure more effectively and directly. 



> Since our use case is just blob storage that will be dumped and pulled in entirety, it seems like we can get away with not paying the write-time cost and just use the `json` type.  Am I missing anything there?

Yes, I agree for this particular use case this makes sense! I didn't realize what was being written and read when - your / Truman / Ty's comments help me to see this now. If we needed to make use of a general-purpose cache of this data for some reason, adding the ability to query into the json might be beneficial, but this isn't that use case! My apologies for my earlier confusion around this.

trumanc:
This PR is being closed without merging, since we've accomplished enough of an improvement with other changes ( like Andrew's disk based cache [here]( https://github.com/usdigitalresponse/usdr-gost/pull/842) and Aditya's async batching changes [here](https://github.com/usdigitalresponse/usdr-gost/pull/834) and [here](https://github.com/usdigitalresponse/usdr-gost/pull/837) that we don't need to pursue this also.

Issue #839, opened 2023-01-20
Linting is disabled for arpa_reporter


NatHillardUSDR:
Github is currently complaining in various PRs about line length violations. 

We should fix both the violation and the linter rule that is prompting them

Original theory:

> in `.editorconfig` we currently have the line `max_line_length = 100`. This is presumably intended to offer cross-IDE standardization of line-length. 
> 
> Locally, however, this restriction is also enforced by eslint, which uses more sophisticated notions of appropriate line length per file-type and even per function context. As it stands, I’m not sure this line is adding much!
> Github is apparently also picking this up, and seems to be complaining on all current PRs. 
> It seems the easiest way to cut down on this is to remove the line from `.editorconfig` and just rely on `eslint`

NatHillardUSDR:
D'oh - I discovered after this merge that, while this does help with local IDE errors, it doesn't address the errors on github, which actually are coming from eslint -- this is coming from `Build website deployment artifact` in `Build GOST Website`, as seen [here](https://github.com/usdigitalresponse/usdr-gost/actions/runs/3962245463/jobs/6788661153).

In turn, these aren't being caught by standard processes because of [these lines](https://github.com/usdigitalresponse/usdr-gost/blob/ebc0d8e63a747aea0794357535d13794e5a916b3/packages/client/.eslintrc.js#L15): 
```
    // this is temporary, just to make committing on ARPA integration dev branch a bit
    // less annoying.
```

So - the real fix for the github annoyingness is to 1)undersetand what the original annoyingness was all about, 2) remove this ignore, then 3) fix the errors with eslint

ClaireValdivia:
@ajhyndman can you confirm if this ticket was addressed? Thanks!

ClaireValdivia:
Note - still need to confirm if this one was addressed @as1729

ClaireValdivia:
confirmed in meeting that the was completed.

PR #857, opened 2023-01-25
fix: Linting is disabled for arpa_reporter (#839)

NatHillardUSDR:
### Ticket #
#839

### Description
Removes arpa_reporter linter exemption and re-runs linter. There are two parts to this fix:
1. Automatically-applied fixes, and 
2. Those that required manual intervention

A note that we may want to wait on merging this, as it impacts a lot of files (even if these changes are largely automatic), and there is a risk that one or more of these manual fixes could cause a regression. This may mean we want to sit on this for a time until the end of the current reporting period. Unfortunately this may also mean conflicts when we rebase, though its probably best to be cautious. 

The latest version of this patch covers client and server

### Screenshots / Demo Video
N/A

### Testing
* Ran `yarn test` (side note, `sendGrantDigestForAgency` is failing locally though not with docker)
* Ran app locally

### Checklist
- [x] Added PR reviewers
- [] Ensure at least 1 review before merging


trumanc (commented):
Love that this is getting fixed, and I agree with your assessment that it should probably wait until after the deadline.

You mentioned that most of the updates are automatic, but some required manual intervention. Do you mind pulling out the manual interventions into a separate commit so that it is easier to check those without losing them in the other ~1k lines of diff?

NatHillardUSDR:
Hey @trumanc  - thanks for taking a look! There are currently two commits, the first is automatic and the second is the manual changes - they appear as one diff in the default view but you can separate them out in the per-commit view

packages/client/.eslintrc.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
     'func-names': 'off',
     'vue/multi-word-component-names': 'off',
     'vue/no-mutating-props': 'off',
+    // Allow for-of statements
+    'no-restricted-syntax': 0,
```
ajhyndman, line 27: Nit:

I see you removed one in-line eslint-disable comment and disabled this rule globally here.  Was that the only failing line?

If so, I think I'd just as soon leave the in-line exception in place instead of blanket disabling.

NatHillardUSDR: Thanks for taking a look! 

In addition to the line that had already been excepted, there were a handful of places where this rule came up before I ran the `--fix` command / added the global disable - I would have had to have added exceptions for each of these affected lines. 

I do think `for-of` is a bit more readable overall; that said, starting with baseline compatibility with airbnb might be good and we can then make changes across the board for all files. I can re-disable for now for consistency

NatHillardUSDR:
Rebased against latest `_staging` just now and re-ran for changed files

NatHillardUSDR:
Rebased against latest _staging just now and re-ran for changed files ; also re-enabled forOf restriction and replaced it where relevant

NatHillardUSDR:
This latest set of patches also includes the `server` directory (previously only had client) ,  but I seem to have introduced a bug; will investigate

NatHillardUSDR:
I discovered that I was testing with a non-admin user, thus the bug I was seeing before! I do still need to debug the broken test, but this should be straightforward. I'll have a fix for that tomorrow.

packages/client/.eslintrc.js
ajhyndman: Oh, you make a good point.  I wouldn't want to prevent ourselves from using for-of syntax entirely.  I definitely don't feel strongly here.

NatHillardUSDR:
Tests are passing now, but functionality appears to be broken around validating uploads - I'm seeing the following when I attempt to validate:
Screenshot 2023-02-16 at 10 00 28 PM Small

I'm beginning to wonder if the potential risk is worth the fallout here - if we had full integration tests and snapshot tests, I would feel more comfortable with this change overall, but as things stand I worry there is important functionality that these changes have inadvertently broken that we may not catch. 

I can try to debug this issue -- maybe I can put a time box on this for the next few days (and I could use assistance with this if possible, though realize that this may not be the best overall use of our time). This all said,  it may be that the best path forward is to lint the files we touch as we touch them, and approach this more incrementally. Unfortunately we'd still need to do this manually but it would be less risky.

For now I'm prioritizing the work around the notes field over this particular fix with the above in mind

ajhyndman:
@NatHillardUSDR You're certainly right that upload notes makes sense as the higher priority.

I'd really love to get this over the line if we can though, so I'll try taking another look tonight.

NatHillardUSDR:
Discussed this with Andrew, we decided to take this on more incrementally. This PR covers just the automated fixes: https://github.com/usdigitalresponse/usdr-gost/pull/1036 , and manually comments out files.

PR #871, opened 2023-01-27
Fixes for API URLs pointing the wrong domain

TylerHendrickson:
### Description

This PR addresses a few places in the ARPA Reporter client where the `fetchApi.apiURL()` helper was not being used to format URLs intended to point to the API backend. In Render, where the client and backend share the same domain name, it's easy for this to go unnoticed because browsers default to the current domain. However, this causes issues in environments where the API is hosted on a different domain, such as AWS.

Additionally, this fixes a bug (likely introduced in #864 ) that caused the "Download file" button to stop appearing on the Uploads tab of the ARPA Reporter client.

### Checklist
- [X] Added PR reviewers
- [X] Ensure at least 1 review before merging

### DIFF
--- a/packages/client/src/arpa_reporter/components/DownloadFileButton.vue
+++ b/packages/client/src/arpa_reporter/components/DownloadFileButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <a :href="apiURL(`/api/uploads/${upload.id}/download`)" class="btn-sm btn-primary ml-2">
+  <a :href="href" class="btn-sm btn-primary ml-2">
       Download file
   </a>
 </template>
@@ -11,6 +11,11 @@ export default {
   name: 'DownloadFileButton',
   props: {
     upload: Object
+  },
+  computed: {
+    href: function () {
+      return apiURL(`/api/uploads/${upload.id}/download`)
+    }
   }
 }
 </script>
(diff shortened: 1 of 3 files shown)


packages/client/src/arpa_reporter/components/DownloadFileButton.vue
```diff
@@ -11,6 +11,11 @@ export default {
   name: 'DownloadFileButton',
   props: {
     upload: Object
+  },
+  computed: {
```
as1729, line 15: should this be `methods` instead?

as1729 (approved):
If this change works and displays the `download` button, then I'm good with it.

packages/client/src/arpa_reporter/components/DownloadFileButton.vue
TylerHendrickson: Unclear... I was more or less following the [`DownloadTemplateBtn.vue` component template ](https://github.com/usdigitalresponse/usdr-gost/blob/fd95ebf635a1870a7e9ec214ed0a1e165b1698e0/packages/client/src/arpa_reporter/components/DownloadTemplateBtn.vue) as a guide, which uses `computed` to set the `href` value for the rendered component.

PR #920, opened 2023-02-05
fix: validation error when agency is 0

### LINKED ISSUE #616: [Infra/ARPA] On the ARPA tool, we cannot add a user to the USDR agency
image

This works in the ID tool, so there might be a special check that ARPA tool is doing on `agency_id`?

disambiguator:
### Ticket #616 

Fixes #616 

### Description
Fixes error when agency_id is 0. Since 0 is falsy in JavaScript, we need to explicitly
check for null or undefined.

### DIFF
--- a/packages/server/src/arpa_reporter/routes/users.js
+++ b/packages/server/src/arpa_reporter/routes/users.js
@@ -20,7 +20,7 @@ async function validateUser (user, creator) {
     throw new Error('Role required')
   }
 
-  if (!agency_id) {
+  if (agency_id == null) {
     throw new Error('Cannot create user without agency_id')
   }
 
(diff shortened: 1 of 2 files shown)


packages/server/src/arpa_reporter/services/validate-upload.js
```diff
@@ -650,7 +650,6 @@ async function validateUpload (upload, user, trns = null) {
 
 module.exports = {
   validateUpload,
-  ValidationError
```
disambiguator, line 653: unrelated fix, but this is an unused export

as1729 (approved):
Thanks for doing this!

packages/server/src/arpa_reporter/routes/users.js
```diff
@@ -20,7 +20,7 @@ async function validateUser (user, creator) {
     throw new Error('Role required')
   }
 
-  if (!agency_id) {
+  if (agency_id == null) {
```
as1729, line 23: @disambiguator can we add some test coverage around this as well? I noticed late that there wasn't unit tests around this. Since this function us sed by many endpoints it should help improve overall repo coverage.

PR #942, opened 2023-02-10
832: Use a disk cache (TEMP_DIR) to optimize audit report generation

ajhyndman:
### Ticket #832

NOTE: This re-introduces #842.

### Description

Audit report generation has been getting progressively slower, as we need to process more and more upload workbooks to produce the desired report.

The motivating idea behind this approach is that the single slowest part of audit report generation is not the logic, formatting or even disk access.  It is the `XLSX.read` operation.

Short of digging into this third party code, a minimally invasive way to optimise this expensive step is to introduce a cache layer.

Each upload, parsed as JSON is about a megabyte in size.  Our production ARPA tool manages about 1000 uploads today.  That's a little more data than I think we should be comfortable holding in RAM.  We could persist this cached content in our database, but parsing and printing JSON into a database involves overhead and adds load to the DB service.

In this diff, I try a relatively naive disk cache instead.  This mirrors the way we handle the original .xlsm upload content today.

Operationally, the impact of this diff is:
- No DB migrations required (now or in the future)
- No manual scripts required to populate cache
- Changes to the audit report or record parsing logic do not require manual intervention
- Cache will need to be flushed if and only if we change the parsing options passed to `XLSX.read`.
- To flush the cache: wipe the TEMP_DIR folder from the deployment's disk instance

This pull request is an alternative approach to #833.  We should land one or the other (not both)!

### Screenshots / Demo Video

image

### Testing

To test:
- check out branch
- `yarn install`
- `yarn start`
- Set `VERBOSE=true` in your .env file
- Upload at least one valid workbook, if none present in your instance
- Click the "Download audit report" button on the home screen
- Verify that subsequent downloads are faster, and verbose console logs indicate expected cache usage

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,10 +114,5 @@ dist
 
 .DS_Store
 
-packages/server/src/static/forms/generated
-
-# ARPA Reporter uploads
-packages/server/data
-
 # Locally-persisted postgres development data
 docker/postgres/persistence
--- a/packages/server/.gitignore
+++ b/packages/server/.gitignore
@@ -1,0 +1,7 @@
+src/static/forms/generated
+
+# ARPA Reporter uploads
+data
+
+# ARPA Reporter uploads dir used by mocha tests
+__tests__/arpa_reporter/server/mocha_uploads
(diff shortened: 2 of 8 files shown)


TylerHendrickson (approved):
🚀

PR #957, opened 2023-02-14
951 fix bug with empty emails when there are fewer than 2 new grants

as1729:
### Description
The code was initially ignoring the first two grants and building the email for the remaining. However, the logic should have been to ignore any grant beyond the first 3.

This bug was caused by a mistake in array slicing.

### Screenshots / Demo Video
N/A

### Testing
- Added unit tests

### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/email/email.test.js
+++ b/packages/server/__tests__/email/email.test.js
@@ -351,5 +351,25 @@ describe('Email sender', () => {
 
             expect(sendFake.calledTwice).to.equal(true);
         });
+        it('builds all the grants if fewer than 3 available', async () => {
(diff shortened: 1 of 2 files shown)


TylerHendrickson (changes requested):
Looks like a straightforward fix 👍

Are there any unit test assertions we can add that would have caught this bug earlier?

as1729:
> Looks like a straightforward fix 👍
> 
> Are there any unit test assertions we can add that would have caught this bug earlier?

Great point! Initially I found it hard to test since the function was doing too many things. I refactored the code to make the body-generation into its own function which made it easier for me to add unit tests.

TylerHendrickson (approved):

PR #970, opened 2023-02-15
961: update env variable to be VUE_APP_env

slants3:
Vue expects env variables to have the prefix of `VUE_APP`. Fix previous commit to comply with this requirement.

### DIFF
--- a/packages/client/src/main.js
+++ b/packages/client/src/main.js
@@ -16,9 +16,9 @@ Vue.use(BootstrapVue);
 Vue.use(IconsPlugin);
 Vue.use(Vuelidate);
 Vue.component('v-select', vSelect);
-Vue.prototype.$environment = process.env.GOST_ENVIRONMENT || 'production';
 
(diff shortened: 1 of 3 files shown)


packages/client/src/main.js
```diff
@@ -16,9 +16,9 @@ Vue.use(BootstrapVue);
 Vue.use(IconsPlugin);
 Vue.use(Vuelidate);
 Vue.component('v-select', vSelect);
-Vue.prototype.$environment = process.env.GOST_ENVIRONMENT || 'production';
 
 Vue.config.productionTip = false;
+Vue.prototype.$environment = process.env.VUE_APP_GOST_ENVIRONMENT || 'production';
```
as1729, line 21: Can we change this to 
```js
Vue.prototype.$negative_keywords_enabled = process.env.VUE_APP_NEGATIVE_KEYWORDS_ENABLED === 'true';
```

as1729 (approved):

PR #974, opened 2023-02-16
839 nathillardusdr linter ahyndman

ajhyndman:
Hey @NatHillardUSDR I think I rooted out that last test bug you ran into!

In the knex seed file, I think it's important to return your promises.  If you don't, it's not guaranteed that the next seed file won't start executing before the first finishes.

After resolving that, I found a couple more typos in import statements to requireSrc, but that was super straightforward!

I didn't want to presume to push directly to your branch, so I just pushed a separate branch.  This pull request just merges the changes back into your branch.  🙏


NatHillardUSDR (approved):
Thank you for looking into this, and thank you for catching these issues!

Issue #877, opened 2023-01-30
Migrate VARCHAR columns to TEXT


TylerHendrickson:
**DOD:** Postgres DB columns that previously were type `VARCHAR` are now type `TEXT`, except for columns in the `migrations` table.

**Background:** Many of the columns in our Postgres DB that we use to store string/text data are currently implemented as `VARCHAR(255)`. This pattern has lead to various issues (e.g. as described in #633) that are cause by what is effectively an arbitrary max-character limit on columns used to store text values. It's also worth noting that in Postgres, there is [no performance benefit](https://www.depesz.com/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/) for choosing `VARCHAR` over `TEXT`, as they are both backed by the same underlying (variable-length array) type, which creates a compelling argument for treating `TEXT` as the go-to default type for storing string/text data in Postgres, and resorting to `VARCHAR` only in (presumably rare) cases where business logic would benefit from enforcing a character limit at the DB layer. For this reason, this ticket prescribes migrating all currently-`VARCHAR` columns to `TEXT`, rather than only the columns where the max-length constraint has resulted in a known bug.

**Implementation details:**
- Write a Knex migration script that applies an `ALTER TABLE ... ALTER COLUMN ... TYPE text` operation to every `VARCHAR`-typed column in the database (excluding the `migrations` table, which is managed by Knex).
- Ensure that the migration script runs within a single transaction.
- Run migration script on Staging and ensure that operations that insert/update data in the migrated columns have no apparent regressions.
- Schedule downtime to run the migration script on Production. 

**Other notes:**
- To help identify columns to target, the following SQL will produce a row containing the table and column names for every `VARCHAR` ("character varying") column:
    ```sql
  SELECT  
    c.table_name,  
    c.column_name  
  FROM information_schema.columns c  
    JOIN information_schema.tables t  
      ON t.table_schema = c.table_schema  
      AND t.table_name = c.table_name  
      AND t.table_type = 'BASE TABLE'  
  WHERE c.data_type = 'character varying'  
    AND c.table_schema NOT IN ('information_schema', 'pg_catalog')  
    AND c.table_name NOT IN ('migrations')  
  ORDER BY  
    c.table_name,  
    c.column_name;
    ```
- The above SQL can be modified to produce and execute a dynamic statement that executes within a transaction:
  ```sql
  DO $bulkmigration$
  BEGIN
  EXECUTE (
    SELECT
        string_agg(  
[...]

Issue #633, opened 2022-11-23
Truncate text fields to avoid database insert and update errors


chrisxkeith:
E.g. : 2022-11-23T16:31:13.654667401Z knex error when creating a new row with key 340880 in grants: error: insert into "grants" ("agency_code", "award_ceiling", "cfda_list", "close_date", "cost_sharing", "description", "eligibility_codes", "grant_id", "grant_number", "notes", "open_date", "opportunity_category", "opportunity_status", "raw_body", "reviewer_name", "search_terms", "status", "title") values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18) - value too long for type character varying(255)

Should probably truncate and append ellipses (e.g., '...') to indicate that text has been truncated.

chrisxkeith:
Seems very low occurrence (2 out of 2792)
```
gost-app  | 2022-11-25T17:32:59.744642901Z knex error when creating a new row with key 340880 in grants: error: insert into "grants" ("agency_code", "award_ceiling", "cfda_list", "close_date", "cost_sharing", "description", "eligibility_codes", "grant_id", "grant_number", "notes", "open_date", "opportunity_category", "opportunity_status", "raw_body", "reviewer_name", "search_terms", "status", "title") values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18) - value too long for type character varying(255)
gost-app  | 2022-11-25T17:33:00.782819765Z searching for  @ 1110
gost-app  | 2022-11-25T17:33:20.423877441Z found 15 total results on grants.gov
gost-app  | 2022-11-25T17:33:20.551582546Z knex error when creating a new row with key 340880 in grants: error: insert into "grants" ("agency_code", "award_ceiling", "cfda_list", "close_date", "cost_sharing", "description", "eligibility_codes", "grant_id", "grant_number", "notes", "open_date", "opportunity_category", "opportunity_status", "raw_body", "reviewer_name", "search_terms", "status", "title") values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18) - value too long for type character varying(255)
...
gost-app  | 2022-11-25T17:33:29.740060784Z sync complete!, elapsed: 01:03:58, totalGrantsProcessed: 2792
```

chrisxkeith:
[...]

PR #988, opened 2023-02-19
855 migrate varchar columns to text

chrisxkeith:
### Description
See ticket
### Screenshots / Demo Video
n/a
### Testing
See https://github.com/usdigitalresponse/usdr-gost/issues/877#issuecomment-1435398568
### Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
+++ b/packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
@@ -1,0 +1,39 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        DO $bulkmigration$
+        BEGIN
+        EXECUTE (
+        SELECT
+            string_agg(
+                format (
+                'ALTER TABLE %I ALTER COLUMN %I TYPE text',
+                c.table_name,
+                c.column_name
+                ),
+            E';\n')
+        FROM information_schema.columns c
+            JOIN information_schema.tables t
+            ON t.table_schema = c.table_schema
+            AND t.table_name = c.table_name
+            AND t.table_type = 'BASE TABLE'
+        WHERE c.data_type = 'character varying'
+            AND c.table_schema NOT IN ('information_schema', 'pg_catalog')
+            AND c.table_name NOT IN ('migrations')
+        );
+        END
+        $bulkmigration$;
+    `);
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+// eslint-disable-next-line no-unused-vars
+exports.down = function (knex) {
+    // no going back
+};


packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
```
NatHillardUSDR, line 6: I would wonder about using knex built-ins like `select` / `where` and [alterTable](https://knexjs.org/guide/schema-builder.html#altertable) for this instead? 

There's a tradeoff, in that we'd get further away from the underlying SQL, but we'd also move away from potentially error-prone raw strings, so it could go either way I guess

packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
```diff
@@ -0,0 +1,39 @@
...
+    `);
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+// eslint-disable-next-line no-unused-vars
+exports.down = function (knex) {
+    // no going back
```
NatHillardUSDR, line 38: We should just be able to do `alterTable` in reverse for this migration in `down`

NatHillardUSDR (commented):
Hey, saw this go by, and I had recently implemented a migration for the work with the notes field -- I think we should at least implement a `down` method for this; it might also help to do knex-ified SQL methods but this would probably be a nice to have

packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
NatHillardUSDR: eg 
```js
const tables = await knex('information_schema.tables')
    .select('table_name')
    .where('table_type', 'BASE TABLE')
    .andWhereNot('table_schema', 'information_schema')
    .andWhereNot('table_schema', 'pg_catalog')
    .andWhereNot('table_name', 'migrations');
```

packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
chrisxkeith: Thanks for the suggestion. This approach was suggested by Ty.

If we allowed this migration to be undone, I'm concerned what would happen once there was data in the `cfda_list` column that exceeded 255 characters (see https://github.com/usdigitalresponse/usdr-gost/issues/633). Truncating the data in that column would have a bad effect.

TylerHendrickson: Yeah, reverting the migration in this case would be a destructive change. 

I'm generally wary of the notion of a "downgrade path" in these migration tools outside the context of testing – it generally seems better to determine an upgrade path given the specific nature of an observed problem.

NatHillardUSDR: This makes sense to me - I'd imagine this is probably not something we'd want to run on prod directly, rather we would make another migration.

packages/server/migrations/20230219162719_migrate_varchar_columns_to_text.js
TylerHendrickson: I generally do prefer to use query-builders for the reasons you mentioned, but occasionally I think the query complexity goes beyond what the builder tool does a good job supporting. 

In this case, the inner-most `SELECT` query could theoretically be produced by Knex, requiring some usage of `knex.raw` to produce the `string_agg()` call. However, I don't think that the outer `DO ... BEGIN ... EXECUTE ... END` can be produced natively by Knex, so it would end up requiring (at least one) more `knex.raw()` usages to embed the dynamic subquery. Given that using `knex.raw()` runs counter to the main benefit you mentioned in avoiding "potentially error-prone raw strings", as well as the fact that the resulting contents of the migration file would be (IMO) increasingly non-obvious, _and_ the relative level of effort to produce the inner dynamic query in the first place, I think there are diminishing returns to using the query builder here.

That said, if either of you wants to take a crack at implementing this migration in a more "Knex-native" form, I would be happy to consider it.

chrisxkeith: I'm good with it the way it is.

NatHillardUSDR: Yep this makes sense to me, thank you for reasoning through this! I will consider this as an option in future migrations also!

NatHillardUSDR (approved):

PR #1032, opened 2023-03-01
898: add optional slice for undefined values

misspran:
## Description
Add ? to optionally slice only when value is defined
### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


chrisxkeith (approved):

packages/client/src/views/UpcomingClosingDates.vue
```diff
@@ -213,9 +213,9 @@ export default {
         }
       }
       //                      format date in MM/DD/YY
-      const year = value.slice(2, 4);
```
as1729, line 216: @misspran it looks like the error happens when the `value` is `undefined`.

My recommendation would be to update the top of the `formatDate` function in line 188. This way we can avoid all the unnecessary computation that this function does.
```js
formatDate(value) {
    if (!value) {
         // returns empty string without error if value is undefined
         return '';
    }
    ....rest of code stays the same
}
```

Issue #1044, opened 2023-03-03
Non-numeric values are being accepted for numeric-type fields


trumanc:
This bug was reported in an email titled `[USDR Grants Eng Vols] [Grants HD] USDR ARPA Reporter bug - Non-numerical entries` sent to the grants-helpdesk email on Feb 13th. 

In the linked workbooks, Alysha points out that some users have out in non-numeric values (specifically the string 'N/A' in place of numbers. Here is an example:
![Screenshot 2023-03-03 at 9 35 26 AM](https://user-images.githubusercontent.com/3675290/222789315-b420322e-861c-4e85-b1ff-c7a80e02a709.png)

These values are eventually written to the generated ARPA report, and when that report is uploaded to treasury, that fails with an error because of the unexpected value.

We should add a validation that will throw an error whenever a numeric field has a non-numeric value.

jcomeau1:
Why is this Done? It doesn't look like anything has been addressed.

jcomeau1:
It looks like this was implemented some time ago (March, 2023). This can be closed.

PR #1045, opened 2023-03-03
#1044: Throw error for invalid non-numeric vals

trumanc:
## Description
The backend is incorrectly accepting arbitrary strings as the value for fields that are supposed to be numeric only. This PR adds a validation that will throw an error in that case, while displaying the value that caused the error to help the uploader find the error.

The majority of the code in this PR is actually an expansion of the testing code so that we can more easily write testcases to check validations in the validateRecord method. I added a test for this non-numeric case, but also added a test that verifies that missing required fields will throw an error

## Screenshots / Demo Video
Test workbook includes these values:
Screenshot 2023-03-03 at 9 35 26 AM

Before the change, the validations show no errors:
Screenshot 2023-03-03 at 9 57 58 AM

After the change, we have new errors:
Screenshot 2023-03-03 at 10 01 16 AM

### Automated and Unit Tests
- [X] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [X] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/services/validate-upload.js
+++ b/packages/server/src/arpa_reporter/services/validate-upload.js
@@ -350,7 +350,6 @@ async function validateRecord({ upload, record, typeRules: rules }) {
             }
 
             if (rule.dataType === 'Numeric') {
-                console.log
               if (typeof(value) === 'string' && isNaN(parseFloat(value))) {
                 // If this value is a string that can't be interpretted as a number, then error.
                 // Note: This value might not be exactly what was entered in the workbook. The val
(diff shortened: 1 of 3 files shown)


packages/server/src/arpa_reporter/services/validate-upload.js
```diff
@@ -349,6 +349,19 @@ async function validateRecord({ upload, record, typeRules: rules }) {
                 }
             }
 
+            if (rule.dataType === 'Numeric') {
+                console.log
```
as1729, line 353: was this `console.log` committed by mistake here?

as1729 (approved):
Aside from the minor change this looks good 👍

packages/server/src/arpa_reporter/services/validate-upload.js
trumanc: ah yes, that must've slipped through while I was trying to debug something. Removed now, thanks!

PR #1035, opened 2023-03-02
feat: support explicitly setting agency and reporting period in the ARPA Reporter upload api (#895)

NatHillardUSDR:
Backend change to allow setting agency and reporting period in uploads. Not yet populated on the frontend.

DRAFT until I can verify a few scenarios / have a few examples to work against

## Description
Adding a simple change to apply the logic from the original ticket when uploading values for agencyId and reportingPeriodId

## Screenshots / Demo Video
N/A, backend only change

## Testing
* Not yet possible to test this, as values are not being populated in the UI, but you should be able to upload workbooks as usual without side effects

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer
- [x] Added steps to test feature/functionality manually

## Checklist
- [x] Provided ticket and description
- [x] Provided screenshots/demo
- [x] Provided testing information
- [x] Provided adequate test coverage for all new code
- [x] Added PR reviewers

Issue #786, opened 2023-01-10
[ARPA] Add supplemental data to workbook uploads


cajinajr:
Creating an issue for the Upload Supplemental data feature that I've been working on for engineering. Here's a [link to the 1-pager](https://docs.google.com/document/d/1A3wrYUu1mZIv-WluCW-lsWu1ETyp8U6FRW1ZIEL0flc/edit?usp=sharing) that @trumanc put together for reference.

The core issue that is being addressed is improving the workbook upload experience. One of the ways we're attempting to do this is providing a way for the uploader to include a comment or "note" (my preferred term) along with their workbook upload. Doing so would allow for uploads to outline what may have changed within the file between uploads, as well as any other thoughts they might want to include. With notes appended to the upload, others can review those to get a sense of the history of an upload series. Below are mockup screenshots (taken from my [Figma file](https://www.figma.com/file/dpMckPYtCGwPV5p85wcpTk/ARPA-Reporter?node-id=135%3A12&t=gnAVsBn3RMTIIYh4-1) for the project) that imagine what that could look like based on feedback from the partners. I believe this is ready for engineering to implement.

Backend implementation tasks:
- #895 
- #896 

Technical implementation details for front-end:
- New fields for the upload form
	- **Agency Code**
		- Presents a dropdown of Agency Codes representing Agencies available to the user, mapped to IDs of the respective agency.
		- Submits as `agencyId` in POST request form data (i.e. the user sees and selects an Agency Code from the dropdown, but the corresponding Agency ID is what is submitted to the API).
		- If the user has a default Agency, the corresponding dropdown item for that Agency should be initially preselected, but the user should still be allowed to change the selection to a different available option.
	- **Reporting Period**
		- Displays in upload the form as the name of the reporting period. This value is display-only, and should not be modifiable by the user. It should reflect the current reporting period (the same value displayed at the top of the page).
		- The ID corresponding to the displayed Reporting period should be submitted as `reportingPeriodId` in the POST request form data.
	- **Notes**
		- Freeform text area (multiline/scrollable) for the user to input plain-text notes (no rich formatting). Input should be encoded as UTF-8.
		- Submits as `notes` in the POST request form data.
- Modified schema in response from `GET /api_uploads`:
	- Upload objects (members of the top-level `uploads` array in the response body) will now contain a nullable string `notes` field, which should be displayed in the redesigned **Version History** block for each upload, as well as on the top-level Uploads page.

----

[...]

PR #1048, opened 2023-03-04
Support freeform text notes on uploads (896)

NatHillardUSDR:
## Description
Adds a form to set reportingId, agencyId, and notes in addition to uploading a file, using Vue-Formulate

## Screenshots / Demo Video
Screenshot 2023-03-04 at 4 33 20 PM
Screenshot 2023-03-04 at 4 35 40 PM

## Testing
NOTE! This patch needs to be applied alongside https://github.com/usdigitalresponse/usdr-gost/pull/1035 for the functionality to work! 
Once that has been merged / this is made on top, you should be able to go to Uploads -> Submit Workbook and specify an agency and reporting period

### Automated and Unit Tests
- [ ] Added Unit tests 
(N/A - we do not yet have FE unit tests)

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

NOTE: Setting as DRAFT for now in case @ajhyndman has another patch in the works / is planning on a different fix! Currently coordinating in slack.

### DIFF
--- a/packages/client/src/arpa_reporter/views/NewUpload.vue
+++ b/packages/client/src/arpa_reporter/views/NewUpload.vue
@@ -122,19 +122,9 @@ export default {
       return file;
     },
     async onSubmit(formData) {
-      // This is technically a FileList, allowing for multiple uploads if configured
-      // this way, but for now we have only a single file.
-      const file = formData.spreadsheet[0];
-      const uploadData = {
-        reportingPeriodId: formData.reportingPeriodId,
-        agencyId: formData.agencyId,
-        expenditure_code: formData.expenditure_code,
-        notes: formData.notes,
-      };
       const uploadFormData = new FormData();
-      uploadFormData.append('spreadsheet', file);
-      Object.keys(uploadData).forEach((key) => {
-        const value = uploadData[key];
+      uploadFormData.append('spreadsheet', data.spreadsheet[0]);
+      Object.entries(_.omit(data, ['spreadsheet'])).forEach(([key, value]) => {
         if (value) {
           uploadFormData.append(key, value);
         }
--- a/packages/client/src/arpa_reporter/views/NewUpload.vue
+++ b/packages/client/src/arpa_reporter/views/NewUpload.vue
@@ -82,7 +82,7 @@ export default {
         },
         {
           component: 'div',
-          class: 'row',
+          class: 'd-flex',
           children: [
             {
               name: 'submit',
--- a/packages/client/src/arpa_reporter/views/NewUpload.vue
+++ b/packages/client/src/arpa_reporter/views/NewUpload.vue
@@ -48,6 +48,7 @@ export default {
           placeholder: 'Select one',
           options: this.reportingPeriodSelectItems(),
           validation: 'required',
+          disabled: true,
         },
         {
           type: 'select',
--- a/packages/client/src/arpa_reporter/views/NewUpload.vue
+++ b/packages/client/src/arpa_reporter/views/NewUpload.vue
@@ -1,95 +1,150 @@
 <template>
-  <div class="upload">
-    <h1>Submit Workbook</h1>
+  <div>
 
-    <div>
-      <div v-if="error" class="mt-3 alert alert-danger" role="alert">
-        {{ error }}
-      </div>
-
-      <form
-        method="post"
-        encType="multipart/form-data"
-        ref="form"
-        @submit.prevent="onSubmit"
-      >
-        <div class="form-group">
-          <input
-            class="form-control-file"
-            type="file"
-            id="spreadsheet"
-            name="spreadsheet"
-            @change="changeFiles"
-            ref="files"
-          />
-        </div>
-        <div class="form-group">
-          <button
-            class="btn btn-primary"
-            type="submit"
-            :disabled="uploadDisabled"
-            @click.prevent="onSubmit"
-          >
-            {{ uploadButtonLabel }}
-          </button>
-          <a class="ml-3" href="#" @click="cancelUpload">Cancel</a>
-        </div>
-      </form>
-    </div>
+    <FormulateForm
+      name="new-upload"
+      :form-errors="formErrors"
+      v-model="values"
+      :schema="schema"
+      @submit="onSubmit"
+    >
+    <FormulateErrors
+      class="alert alert-danger"
+    />
+    </FormulateForm>
   </div>
 </template>
 
 <script>
+import Vue from 'vue';
+import VueFormulate from '@braid/vue-formulate';
 import { postForm } from '../store';
 
+Vue.use(VueFormulate, {
+  classes: {
+    prefix: 'formulate-',
+    outer: 'form-group',
+    input: (context, classes) => {
+      // set this to 'form-control' unless it's a file input
+      const inputClass = context.type === 'file' ? '' : 'form-control';
+      return [classes, inputClass].join(' ');
+    },
+    inputHasErrors: 'is-invalid',
+    help: 'form-text text-muted',
+    errors: 'list-unstyled text-danger',
+  },
+});
 export default {
-  name: 'NewUpload',
   data() {
     return {
-      error: null,
-      files: null,
-      uploading: false,
+      formErrors: [],
+      initialValues: () => ({
+        reportingPeriodId: this.$store.getters.viewPeriodID ?? '',
+        agencyId: this.$store.getters.user.agency.id,
+      }),
+
+      values: {},
+      schema: [
+        {
+          type: 'select',
+          name: 'reportingPeriodId',
+          label: 'Reporting Period',
+          placeholder: 'Select one',
+          options: this.reportingPeriodSelectItems(),
+          validation: 'required',
+        },
+        {
+          type: 'select',
+          name: 'agencyId',
+          label: 'Agency Code',
+          options: this.agencySelectItems(),
+          validation: 'required',
+        },
+        {
+          type: 'text',
+          label: 'Expenditure Code',
+          name: 'expenditure_code',
+          validation: 'required',
+        },
+        {
+          type: 'file',
+          label: 'Workbook File',
+          name: 'spreadsheet',
+          validation: 'required',
+          uploader: this.uploadFile,
+          // Don't upload the file until the form is submitted
+          uploadBehavior: 'delayed',
+        },
+        {
+          type: 'textarea',
+          label: 'Notes',
+          name: 'notes',
+        },
+        {
+          component: 'div',
+          class: 'row',
+          children: [
+            {
+              name: 'submit',
+              type: 'submit',
+              label: 'Submit',
+              inputClass: ['btn', 'btn-primary'],
+            },
+            {
+              type: 'button',
+              label: 'Reset',
+              inputClass: ['btn', 'btn-danger', 'ml-2'],
+              '@click': this.onReset,
+            },
+          ],
+        },
+      ],
     };
   },
-  computed: {
-    uploadButtonLabel() {
-      return this.uploading ? 'Uploading...' : 'Upload';
-    },
-    uploadDisabled() {
-      return this.files === null || this.uploading;
-    },
+  created() {
+    this.values = this.initialValues();
   },
   methods: {
-    changeFiles(evt) {
-      this.files = evt.target.files;
+    reportingPeriodSelectItems() {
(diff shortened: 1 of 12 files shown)


packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+          uploadBehavior: 'delayed',
+        },
+        {
+          type: 'textarea',
+          label: 'Notes',
+          name: 'notes',
+        },
+        {
+          component: 'div',
+          class: 'row',
```
ajhyndman, line 85: I noticed there's a slightly wonky layout effect here.  I think the issue is that the `row` class isn't meant to be used without a corresponding `col`.

Probably the easiest fix is just to swap this for `d-flex`!
```suggestion
          class: 'd-flex',
```

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+import VueFormulate from '@braid/vue-formulate';
 import { postForm } from '../store';
 
+Vue.use(VueFormulate, {
+  classes: {
+    prefix: 'formulate-',
+    outer: 'form-group',
+    input: (context, classes) => {
+      // set this to 'form-control' unless it's a file input
+      const inputClass = context.type === 'file' ? '' : 'form-control';
```
ajhyndman, line 29:
```suggestion
      const inputClass = context.type === 'file' ? 'form-control-file' : 'form-control';
```

There is actually a form-control-file class we can use here.  It looks like it doesn't change anything visually, but it does apply `display: block` to the element, creating a larger click target, which seems nice.

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+      return [classes, inputClass].join(' ');
```
ajhyndman, line 30: According to the documentation, if we return an array here, it will automatically join the new input class with existing classes.

```suggestion
      return [inputClass];
```

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
 <template>
-  <div class="upload">
-    <h1>Submit Workbook</h1>
```
ajhyndman, line 3: I'd prefer to bring back the document heading here!

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
 </template>
 
 <script>
+import Vue from 'vue';
+import VueFormulate from '@braid/vue-formulate';
 import { postForm } from '../store';
 
+Vue.use(VueFormulate, {
+  classes: {
+    prefix: 'formulate-',
```
ajhyndman, line 25: What is this prefix class for?  It doesn't obviously change anything in my testing.

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
   data() {
     return {
-      error: null,
-      files: null,
-      uploading: false,
+      formErrors: [],
+      initialValues: () => ({
+        reportingPeriodId: this.$store.getters.viewPeriodID ?? '',
+        agencyId: this.$store.getters.user.agency.id,
+      }),
```
ajhyndman, line 44: Could we move this into `methods` alongside `reportingPeriodSelectItems` and `agencySelectItems`?

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+
+      values: {},
```
ajhyndman, line 46:
```suggestion
      values: this.initialValues(),
```

Instead of populating this in the `created` lifecycle hook, can we just populate it here?

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+    // eslint-disable-next-line no-unused-vars
+    async uploadFile(file, progress, error, options) {
+      /* VueFormulate traditionally calls this method to upload the file
+         to some remote location, but for now, we don't want to do that.
+         It may eventually make sense to do that when we don't have access
+         to the file system, e.g in a lambda context. For now though,
+         we just return the file object. */
+      return file;
+    },
+    async onSubmit(formData) {
```
ajhyndman, line 124:
```suggestion
    async onSubmit(data) {
```

Nit: just for clarity, could we give this a name that doesn't imply it is an instance of the `FormData` class?

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,111 +1,164 @@
...
+        notes: formData.notes,
+      };
+      const uploadFormData = new FormData();
+      uploadFormData.append('spreadsheet', file);
+      Object.keys(uploadData).forEach((key) => {
+        const value = uploadData[key];
+        if (value) {
+          uploadFormData.append(key, value);
+        }
+      });
```
ajhyndman, line 141:
```suggestion
      const uploadFormData = new FormData();
      uploadFormData.append('spreadsheet', data.spreadsheet[0]);
      Object.entries(_.omit(data, ['spreadsheet'])).forEach(([key, value]) => {
        if (value) {
          uploadFormData.append(key, value);
        }
      });
```

Nit: We do have lodash available in this project.  Using that, we can probably simplify this block a little!

packages/client/src/main.js
```diff
@@ -15,6 +15,7 @@ Vue.use(BootstrapVue);
 // Optionally install the BootstrapVue icon components plugin
 Vue.use(IconsPlugin);
 Vue.use(Vuelidate);
+Vue.use(VueFormulate);
```
ajhyndman, line 18: I think it isn't necessary to mount VueFormulate globally, is it?

ajhyndman (changes requested):
Overall, I'm quite happy with this approach!

I went through and left some polish suggestions, but they are all pretty minor.

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: Ah, good catch!

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: Ah cool - thanks for catching!

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: 🤦

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: Ah, interesting - this is actually doubly wrong. 

I believe I had seen a reference to this in an issue in the repo, as a means of prefixing class names so as to avoid conflicts with other library class names. The property is apparently actually named `idPrefix`, and is a sibling to `classes`. 

Also though, by [default](https://github.com/wearebraid/vue-formulate/blob/3fdd6fdc2ed82113e6a78dfeec800b36763d7f95/src/Formulate.js#L96) it's `formulate-`, so i can just remove this line.

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: 👍 I had initially moved it here due to the issue I go into below, but if I move to methods and then set properly it works

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: I had briefly tried this before only to find it didn't populate correctly. That said! Trying it again, I'm realizing now, in having used an arrow function, I hadn't sufficiently captured `this`, which was the true culprit!

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: A good call, fixed

packages/client/src/arpa_reporter/views/NewUpload.vue
NatHillardUSDR: Ooh, I have less familiarity with all the functionality available in lodash - nice to have this!

packages/client/src/main.js
NatHillardUSDR: I tried various variations of this - It seems to only work with mounting it on a per-file basis. It also doesn't play well with exports/imports, or with attempts to register it under `components`... 

Removing it globally and keeping it per-file for now seems to work, I can keep it this way.

NatHillardUSDR:
Thanks @ajhyndman for your review! I appreciate your thoughtful approach and suggestions, and I learned a few things in the process. 

I made your recommended fixes, made a few additional, then rebased to clean up earlier commits / rebased against `_staging` as well. Taking out of draft, as #1035 just landed, which I also rebased against

TylerHendrickson (approved):
Looks great!

NatHillardUSDR:
Just now I:
1. Fixed a few linting errors on files that were noted by the CI job, and
2. Updated the reportingPeriodId dropdown to be readonly, per the AC spelled out in #786 

Here's a screenshot with the field disabled - note it still sends the value in the BE:
Screenshot 2023-03-06 at 1 26 49 PM

Thank you @TylerHendrickson for the review; @ajhyndman can you take a look as well to see if this addresses the issues you called out?

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,113 +1,154 @@
...
+    <FormulateForm
+      name="new-upload"
+      :form-errors="formErrors"
+      v-model="values"
+      :schema="schema"
+      @submit="onSubmit"
+    >
+    <FormulateErrors
+      class="alert alert-danger"
+    />
```
ajhyndman, line 13:
```suggestion
      <FormulateErrors
        class="alert alert-danger"
      />
```

Nit: The indentation looks slightly off to me here.

packages/client/src/arpa_reporter/views/NewUpload.vue
ajhyndman: Ahh, nice catch!  Yeah, that would be a sneaky one.

packages/client/src/arpa_reporter/views/NewUpload.vue
ajhyndman: I have found the lodash docs to be pretty good!  I do still have to check them pretty frequently to double-check the exact syntax.

https://lodash.com/docs/

packages/client/src/main.js
```diff
@@ -2,7 +2,6 @@ import Vue from 'vue';
 import { BootstrapVue, IconsPlugin } from 'bootstrap-vue';
 import vSelect from 'vue-select';
 import Vuelidate from 'vuelidate';
-
```
ajhyndman, line 5: Nit: I think we lost this blank line during a revert.

ajhyndman (approved):
Thanks!  This looks great.

I left two formatting nits, but honestly you should probably just ignore them for now.  We should set up an auto-formatter (like Prettier) at some point.  That would fix this too.

NatHillardUSDR:
Thanks! The formatting issues you call out are odd, I'm not seeing these on my branch - it's possible this arose due to having rebased + force pushed? I can look to avoid this workflow in the future actually, as if you've pulled locally weird things can happen.

packages/client/src/arpa_reporter/views/NewUpload.vue
```diff
@@ -1,113 +1,154 @@
...
+          type: 'select',
+          name: 'agencyId',
+          label: 'Agency Code',
+          options: this.agencySelectItems(),
+          validation: 'required',
+        },
+        {
+          type: 'text',
+          label: 'Expenditure Code',
+          name: 'expenditure_code',
```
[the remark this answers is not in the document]
NatHillardUSDR: Ah interesting - I had gotten this from this screenshot: 
image
in the original supplemental data parent [issue](https://github.com/usdigitalresponse/usdr-gost/issues/786).

Though actually, it's true, this is not used by `createUploadRow` which actually determines the data inserted into the DB. I can remove this for now.

PR #1073, opened 2023-03-07
#378: Improve conditionally-required fields

trumanc:
## Description
This PR aims to improve how we can enforce conditionally-required fields in ARPA reporter data (that is, fields that are sometimes required and sometimes not, depending on your answers to other questions).

This feature already existed in a hard-coded way, where we had a list of field ids that we knew should be treated as optional if the project status was "Not started".

This PR starts by refactoring that hard-coded feature in a way that is easier to add new conditions for new fields in the future. It creates a new `CONDITIONAL_REQS_CONFIGS` list for these configurations, and uses that to enforce the optional-if-not-started rules.

Then, in the second commit, I show how it is easy to add a new configuration. Here I added a rule that enforces that if the project status is "Cancelled", then the Cancellation_Reasons__c field must be required.

## Screenshots / Demo Video
I uploaded a test project that has status 'Cancelled' but no reason included. After commit 1 (refactoring the existing rule) but before commit 2 (adding the new cancellation_reason rule) we see that the file validates without errors:
Screenshot 2023-03-07 at 10 55 21 AM

But after adding the 2-lines in commit 2, there is a new validation error:
Screenshot 2023-03-07 at 10 56 30 AM

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/arpa_reporter/server/services/validation-rules.spec.js
+++ b/packages/server/__tests__/arpa_reporter/server/services/validation-rules.spec.js
@@ -69,8 +69,8 @@ describe('validation rules', () => {
         const configs = validationRulesModule.__get__('CONDITIONAL_REQS_CONFIGS');
         const funcLookup = validationRulesModule.__get__('CONDITIONAL_REQUIREMENTS_BY_FIELD_ID');
         it('has valid configurations', () => {
-            const seenFieldIds = new Set()
-            assert(configs.length > 0)
+            const seenFieldIds = new Set();
+            assert(configs.length > 0);
             for (const config of configs) {
                 assert(config.fieldIDs && config.fieldIDs.length > 0,
                     'Conditional requirement config is missing fieldIDs');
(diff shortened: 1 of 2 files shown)


NatHillardUSDR (commented):
Looks good, this is an interesting approach!

Can we include unit tests for these mini functions as well? Would overall be good to set a precedent for adding tests for this kind of thing in the future.

Also - have we considered something like [JSON Schema](https://json-schema.org/) or similar for this and similar validations? I'd wonder if we could use an out of the box solution to encoding these kinds things

trumanc:
Good call, I'll add some tests for these now.

@NatHillardUSDR can you say a little bit more about what you mean about JSON schema or other library based validations? How do you envision a library being used?

NatHillardUSDR (approved):
Nice! Looks great. Thanks for adding the tests, this also helps contextualize how these might be used.

trumanc:
@NatHillardUSDR in repsonse to your question:
> I'm wondering at the end of the day if what we're really dealing with is schema validation, that is to say.

It's an interesting question, and might require a larger discussion to really flesh out how we think about this.
I haven't taken much time to weigh the pros/cons of this particular instance. If you think there is a compelling case for why adding a library here might add value without too much cost then you're welcome to explore the idea and make the case to the team!

PR #1094, opened 2023-03-13
411: Update GrantsTable to Display 50 Items Per Page

slants3:
## Description
Changed `GrantsTable` to display 50 items per page on the My Grants and Browse Grants pages.


as1729:
> Note it adds a scroll bar to the right but the main page outside of the table itself does not scroll. Is this the behavior we want?

yes 💯  we want the contents of table to scroll, but not the page itself. Looks good!

as1729 (approved):

Issue #1027, opened 2023-02-28
Unique Entity ID Requirement Emphasis


as1729:
Editing the item's issue here.

jcomeau1:
The following requirement will be imposed with the April reporting period (for Q1, 2023):

**Unique Entity ID Requirement Emphasis** – All subrecipients and contractors are required to have a Unique Entity ID (UEI) and have that number included as part of the reporting process. The UEI is the replacement for the previously used DUNS numbers, and they are issued by SAM.gov. While this requirement is not new, starting in April 2023’s P&E Report the report form will now return an error when no valid UEI is provided when creating new Subrecipient or Contractor entities. Please consult SAM.gov [here](https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fsam.gov%2Fcontent%2Fentity-registration&data=05%7C01%7Cjessica.martin%40obm.ohio.gov%7C3c16c5a0b4b54180629608db14556d8c%7C50f8fcc494d84f0784eb36ed57c7c8a2%7C0%7C0%7C638126129365549102%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=3YlxXmr501oGOndiY%2BWA5uf%2F8kmxie4q08DGdCPw2Fk%3D&reserved=0) if you have questions about obtaining a UEI.

NatHillardUSDR:
I took a look at this a few days ago and have some questions about this - some of these are technical and some are product-facing, but I figured I'd write them out here to keep them all in one place. @trumanc  / @jcomeau1 I'd appreciate if you all could take a look and let me know your thoughts / clarifications! Happy to jump on a call as well. The questions:

a. It seems we already have UEI as a field for recipients - is this just a matter of making this a required field?
b. what are the db implications of this - is this the new primary key? Is some combination of this and other fields a the true primary key?
c. Presumably we need a migration for this?
d. what do we do for old records that don’t have a UEI - is it ok that these are now invalid?
e. what spreadsheet changes are necessary around this, and how does this interact with any required UI changes / how can we coordinate these?

jcomeau1:
A. Yes, but it's conditional based on new subrecipients being added and only those with an entity type of subrecipient or contractor.
[...]

PR #1105, opened 2023-03-15
feat: unique entity id requirement emphasis (#1027)

NatHillardUSDR:
UEI is now required for entities of type "Contractor" and "Subrecipient". Update our validation logic to reflect this.

## Description
Updates our validation logic to reflect new treasury rules.

## Screenshots / Demo Video
Coming soon post confirmation of logic over on https://github.com/usdigitalresponse/usdr-gost/issues/1027

## Testing
Upload a sheet with a Subrecipient row that is missing a UEI but is of type "Subrecipient" - you should see a validation failure

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

## NOTE
I am not sure I did the logic correctly for the db update -- see questions over at https://github.com/usdigitalresponse/usdr-gost/issues/1027

### DIFF
--- a/packages/server/src/arpa_reporter/services/validate-upload.js
+++ b/packages/server/src/arpa_reporter/services/validate-upload.js
@@ -265,7 +265,7 @@ function validateIdentifier(recipient, recipientExists) {
  * @returns {boolean} - true if the recipient belongs to the upload
  */
 function recipientBelongsToUpload(existingRecipient, upload) {
-    return existingRecipient && existingRecipient.upload_id === upload.id && !existingRecipient.updated_at;
+    return Boolean(existingRecipient) && existingRecipient.upload_id === upload.id && !existingRecipient.updated_at;
 }
 
 /**
--- a/packages/server/src/arpa_reporter/services/validate-upload.js
+++ b/packages/server/src/arpa_reporter/services/validate-upload.js
@@ -208,13 +208,22 @@ async function validateReportingPeriod({ upload, records, trns }) {
 }
 
 /**
- * Check whether UEI is required for the recipient
+ * Return an already existing record in the db, defined via UEI or TIN
  * @param {object} recipient - the recipient record
- * @returns {boolean} - true if UEI is required, false otherwise
+ * @param {object} trns - the transaction to use for db queries
+ * @returns {Promise<object>} - the existing recipient record
  */
-function isUEIRequired(recipient) {
-    // As of Q1, 2023 we require a UEI for all entities of type subrecipient and/or contractor
-    return (recipient.Entity_Type_2__c.includes('Contractor') || recipient.Entity_Type_2__c.includes('Subrecipient'));
+async function findRecipientInDatabase({ recipient, trns }) {
+    // There are two types of identifiers, UEI and TIN.
+    // A given recipient may have either or both of these identifiers.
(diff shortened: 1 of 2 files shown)


packages/server/src/arpa_reporter/services/validate-upload.js
```diff
@@ -207,18 +207,52 @@ async function validateReportingPeriod({ upload, records, trns }) {
     return errors;
 }
 
+/**
+ * Check whether UEI is required for the recipient
+ * @param {object} recipient - the recipient record
+ * @returns {boolean} - true if UEI is required, false otherwise
+ */
+function isUEIRequired(recipient) {
+    // As of Q1, 2023 we require a UEI for all entities of type subrecipient and/or contractor
```
as1729, line 216: I thought the rule was `As of Q1, 2023 entity of type beneficiary does not require a UEI`. I'm wondering if the logic should be `return recipient.Entity_type_2.doesnotinclude('beneficiary')` instead.

as1729: 🤦  I should have just read the ticket. This logic is correct!

packages/server/src/arpa_reporter/services/validate-upload.js
```diff
@@ -249,27 +283,29 @@ async function validateSubrecipientRecord({
     || existing.updated_at
     ) isOwnedByThisUpload = false;
 
-    // the record has already been validated before this method was invoked. how
-    // did the validation go?
-    const isRecordValid = recordErrors.length === 0;
+    // the record has already been validated before this method was invoked, or we found an error above. how did the validation go?
+    const isRecordValid = recordErrors.length === 0 && errors.length === 0;
```
as1729, line 287: Was this an existing bug? In the past we never checked the length of `errors` to determine if `isRecordValid` so curious if you noticed this was incorrect and changed it.

NatHillardUSDR:
Thanks for the comments, @as1729 ! 
based on Joe's comments over at https://github.com/usdigitalresponse/usdr-gost/issues/1027 I've got to change the logic slightly. 

Moving back to draft for now to reflect this

packages/server/src/arpa_reporter/services/validate-upload.js
NatHillardUSDR: This is supposed to reflect something Truman called out, which is that if we have a validation error we don't even want to create  or update the record in the DB. I can / should actually call this out more explicitly

packages/server/src/arpa_reporter/services/validate-upload.js
NatHillardUSDR: Right, I didn't want to call out "non-beneficiary" because in the future if for instance we add more entity types this logic wouldn't hold whereas explicitly checking for one of the two others would

NatHillardUSDR:
Made a number of changes to accommodate correct logic based on Joe's latest comment; simplified and wrote unit tests. Re-opening for review.

Need to confirm that this works with a working sheet, but at least logic should be correct, and wanted to open up for review to double check

as1729 (approved):
Thanks for refactoring the logic. The new code is much easier to read and follow!

PR #1126, opened 2023-03-21
[ARPA][Bug] Fix issue with "Forbidden" error for closing reporting periods

as1729:
## Description
The following changes ensure:
1. Cookies are sent in the request to the API which contain auth information.
2. There is logging to debug further if the problem continues to happen.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/arpa_reporter/views/ReportingPeriods.vue
+++ b/packages/client/src/arpa_reporter/views/ReportingPeriods.vue
@@ -72,7 +72,7 @@
 </template>
 
 <script>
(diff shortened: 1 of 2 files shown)


as1729:
I believe the reason why this worked in the past was because the API requests were made to the same origin. But now that we have a different origin for the API we need to explicitly set credentials.
https://stackoverflow.com/questions/34558264/fetch-api-with-cookie

As for why this works locally and does not work on AWS- I'm not 100% clear on it.

PR #1154, opened 2023-03-29
feat: adds an s3 bucket to store arpa reporting data

as1729:
## Description
- Adds an S3 bucket to store audit-report files

## Testing
- Confirmed the terraform validate and and plan CI action is able to create the correct resources.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/terraform/modules/gost_api/storage.tf
+++ b/terraform/modules/gost_api/storage.tf
@@ -23,3 +23,37 @@ module "efs_data_volume" {
     }
   }
 }
+
+module "arpa_audit_reports_bucket" {
+  source  = "cloudposse/s3-bucket/aws"
+  version = "3.0.0"
+  context = module.s3_label.context
+  name    = "arpa_audit_reports"
+
+  acl                          = "private"
+  versioning_enabled           = true
+  sse_algorithm                = "AES256"
+  allow_ssl_requests_only      = true
+  allow_encrypted_uploads_only = true
+  source_policy_documents      = []
+
+  lifecycle_configuration_rules = [
+    {
+      enabled                                = true
+      id                                     = "rule-1"
+      filter_and                             = null
+      abort_incomplete_multipart_upload_days = 1
+      transition                             = [{ days = null }]
+      expiration                             = { days = null }
+      noncurrent_version_transition = [
+        {
+          days          = 30
+          storage_class = "GLACIER"
+        },
+      ]
+      noncurrent_version_expiration = {
+        days = 2557 # 7 years (includes 2 leap days)
+      }
+    }
+  ]
+}
--- a/terraform/modules/gost_api/storage.tf
+++ b/terraform/modules/gost_api/storage.tf
@@ -39,6 +39,7 @@ module "arpa_audit_reports_bucket" {
 
   lifecycle_configuration_rules = [
     {
+      enabled                                = true
       id                                     = "rule-1"
       filter_and                             = null
       abort_incomplete_multipart_upload_days = 1
--- a/terraform/modules/gost_api/storage.tf
+++ b/terraform/modules/gost_api/storage.tf
@@ -41,19 +41,9 @@ module "arpa_audit_reports_bucket" {
     {
       enabled                                = true
       id                                     = "rule-1"
-      filter_and                             = null
       abort_incomplete_multipart_upload_days = 1
-      transition                             = [{ days = null }]
-      expiration                             = { days = null }
-      noncurrent_version_transition = [
-        {
-          days          = 30
-          storage_class = "GLACIER"
-        },
-      ]
-      noncurrent_version_expiration = {
-        days = 7
-      }
+      expiration                             = { days = 14 }
+      noncurrent_version_expiration          = { days = 7 }
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


terraform/modules/gost_api/storage.tf
```diff
@@ -23,3 +23,37 @@ module "efs_data_volume" {
     }
   }
 }
+
+module "arpa_audit_reports_bucket" {
```
as1729, line 27: @TylerHendrickson can you confirm if this is the right place to add the s3 bucket definition in order for it to get created in AWS?

I'm still working on whether the specific configurations is what we need. I'm basing this off of `grants_source_data` bucket which I think is close to what we want to achieve with ARPA reports.

Once I get this setup in my `awslocal` I'll open up this draft PR to a live PR.

TylerHendrickson: Yep, perfect.

terraform/modules/gost_api/storage.tf
```diff
@@ -23,3 +23,37 @@ module "efs_data_volume" {
...
+      transition                             = [{ days = null }]
+      expiration                             = { days = null }
+      noncurrent_version_transition = [
+        {
+          days          = 30
+          storage_class = "GLACIER"
+        },
+      ]
+      noncurrent_version_expiration = {
+        days = 2557 # 7 years (includes 2 leap days)
```
TylerHendrickson, line 55: We should check with @mhuang-usdr to see how long we need to retain these reports from a regulatory (or other?) standpoint, and see if we can expire these documents along that guidance.

as1729: updating this to 7 days as per Joe's suggestion

TylerHendrickson (changes requested):

terraform/modules/gost_api/storage.tf
```diff
@@ -23,3 +23,37 @@ module "efs_data_volume" {
...
+      expiration                             = { days = null }
+      noncurrent_version_transition = [
+        {
+          days          = 30
+          storage_class = "GLACIER"
+        },
+      ]
+      noncurrent_version_expiration = {
+        days = 7
+      }
```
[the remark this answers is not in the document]
as1729: Hmm based on the error it looks like we need to transition these noncurrent objects if we want to keep them for an additional 7 days. Should we remove this requirement entirely or do you think we should keep in Glacier storage for these 7 days?

```
Error: Invalid value for input variable

  on modules/gost_api/storage.tf line 40, in module "arpa_audit_reports_bucket":
  40:   lifecycle_configuration_rules = [
  41:     {
  42:       enabled                                = true
  43:       id                                     = "rule-1"
  44:       abort_incomplete_multipart_upload_days = 1
  45:       expiration                             = { days = 14 }
  46:       noncurrent_version_expiration          = { days = 7 }
  47:     }
  48:   ]

The given value is not suitable for
module.api.module.arpa_audit_reports_bucket.var.lifecycle_configuration_rules
declared at
.terraform/modules/api.arpa_audit_reports_bucket/variables.tf:[15](https://github.com/usdigitalresponse/usdr-gost/actions/runs/4566719470/jobs/8059614205#step:10:16)9,1-41:
element 0: attributes "filter_and", "noncurrent_version_transition", and
"transition" are required.
```

TylerHendrickson (changes requested):
Either this PR or https://github.com/usdigitalresponse/usdr-gost/pull/1157 will need to update the ECS task role with a policy that gives access to the new bucket. Something like this should work:

In `terraform/modules/gost_api/storage.tf`:
```terraform
module "access_arpa_reports_bucket_policy" {
  source  = "cloudposse/iam-policy/aws"
  version = "0.4.0"
  context = module.s3_label.context

  name = "access_arpa_reports_bucket"

  iam_policy_statements = {
    ReadWriteBucketObjects = {
      effect = "Allow"
      actions = [
        "s3:PutObject",
        "s3:GetObject",
      ]
      resources = ["${module.arpa_audit_reports_bucket.bucket_arn}/*"]
    }
  }
}
```

In `terraform/modules/gost_api/task.tf` (modify this resource):
```terraform
resource "aws_iam_role_policy" "task" {
  for_each = !var.enabled ? {} : {
    connect-to-postgres   = module.connect_to_postgres_policy.json
    ecs-exec              = module.ecs_exec_policy.json
    send-emails           = module.send_emails_policy.json
    rw-arpa-audit-reports = module.access_arpa_reports_bucket_policy.json
  }

  name   = each.key
  role   = join("", aws_iam_role.task.*.name)
  policy = each.value
}
```

terraform/modules/gost_api/storage.tf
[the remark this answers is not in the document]
TylerHendrickson: Odd, I thought that the module would handle those `null` defaults, but looks like they need to be specified. I think this should work?
```terraform
      id                                     = "rule-1"
      filter_and                             = null
      abort_incomplete_multipart_upload_days = 1
      expiration                             = { days = 14 }
      transition                             = null
      noncurrent_version_expiration          = { days = 7 }
      noncurrent_version_transition          = null
```

TylerHendrickson (approved):

PR #1155, opened 2023-03-29
Improvements to Docker documentation

pearkes:
## Description

This adds some additional documentation for using Docker and gives a few hints for blockers that someone may run into when first using it. Note: I did take a look through open issues and couldn't find one for any of these dev UX improvements, just opening this based on my experience getting started.

### Manual tests for Reviewer

- [ ] In a new environment, go through the set-up flow for Docker as a developer
- [ ] Ensure you can get to a running application by the end of the setup, can view logs, and can recreate the environment when needed

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,7 +1,9 @@
 # Using Docker for Development
 
 You can use Docker to run the database, API server, and Vue frontend during development,
-which should help ensure consistent development environments and avoid workstation clutter.
+which should help ensure consistent development environments and avoid workstation
+clutter. The easiest way to install Docker is their [Desktop Applications](https://www.docker.com/products/docker-desktop/),
+which can be installed via their GUI installers or your systems package manager.
 
 
 ## Setup
--- a/docker/README.md
+++ b/docker/README.md
@@ -35,6 +35,12 @@ that you invoke it by running `docker-compose <subcommand>`.
 Refer to this section for common scenarios and their solutions when using Docker
 for development.
 
+### Install or update dependencies
+
+After your intitial setup or for any update to the `package.json` file that
+tracks application dependencies, you'll need to run yarn. Thisq
+can be executed via `exec`. For example, to install dependencies after initial
+setup: `docker compose exec app yarn`.
 
 ### Seed and apply migrations to the database
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -42,6 +42,9 @@ tracks application dependencies, you'll need to run yarn. Thisq
 can be executed via `exec`. For example, to install dependencies after initial
 setup: `docker compose exec app yarn`.
 
+For more on how dependencies are managed in the application see the documentation
+on how we use [Workspaces](../docs/workspaces.md).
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -17,7 +19,8 @@ steps to prepare your environment:
   - `POSTGRES_TEST_URL=postgresql://postgres:password123@postgres:5432/usdr_grants_test`
   - You may also have to set the `WEBSITE_DOMAIN` hostname if you are not developing on `localhost`
   ([more info](#cookbook-non-localhost)).
-3. Run `docker compose up -d` to start the services.
+4. Run `docker compose up -d` to start the services in the background (the `-d` flag).
+5. Install application dependencies via yarn: `docker compose exec app yarn`.
```
jakekreider, line 23: I didn't need to do this (I think [this ](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/docker/Dockerfile#L15)takes care of it?), but if you needed to then +1 for including!

docker/README.md
```diff
@@ -32,12 +35,56 @@ that you invoke it by running `docker-compose <subcommand>`.
 Refer to this section for common scenarios and their solutions when using Docker
 for development.
 
+### Install or update dependencies
+
+After your intitial setup or for any update to the `package.json` file that
+tracks application dependencies, you'll need to run yarn. Thisq
```
jakekreider, line 41:
```suggestion
tracks application dependencies, you'll need to run yarn. This
```

jakekreider (approved):
LGTM, helpful!

TylerHendrickson (approved):

docker/README.md
pearkes: Yeah I believe so but in my case I ran tests first — so maybe that’s why. I failed out of tests because of missing deps so I added this.

jakekreider: Oh! Okay, I ran into this too but I fixed it by `nohoist: ['mocha']` to the root `package.json`. The problem here was that mocha needs to be in the path, but when yarn workspace invokes commands on that individual workspace, it does so with the workspace's `node_modules`, not the root monorepo. I don't think it affects the deployment one way or another since the app gets packaged separately. 

@TylerHendrickson - mind if I add the nohoist in a PR, or should the workflow be to run the yarn install directly in the workspace?

TylerHendrickson: @jakekreider Please feel free to add `nohoist` to a PR – I'm not terribly familiar with that particular directive, but after reading up on it [here](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) it sounds like a sensible thing to do.

One thing  to mention (disclaimer that I'm not sure how pertinent this is to the change you're proposing) is that we somewhat recently moved away from Lerna in favor of further decoupling `packages/client` and `packages/server` concerns. I'm curious whether you think it would be worth ditching the root-level `package.json` entirely in the spirit of furthering that theme of decoupling?

PR #1177, opened 2023-04-03
feat: adding WAF to website Cloudfront CDN

slapula:
## Description
See title.  Only adding rules to staging right now until I can get a better idea of how they are going to affect production.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/terraform/modules/gost_website/variables.tf
+++ b/terraform/modules/gost_website/variables.tf
@@ -42,9 +42,9 @@ variable "logs_bucket_versioning" {
 }
 
 variable "managed_waf_rules" {
-  description = "List of parameters for managed WAF rules"
-  type = list(object({
-    name              = string
+  description = "Map of rules and associated parameters for managed WAF ACL"
+  type = map(object({
+    managed_rule      = string
     priority          = number
     metric_visibility = bool
   }))
--- a/terraform/modules/gost_website/variables.tf
+++ b/terraform/modules/gost_website/variables.tf
@@ -43,7 +43,7 @@ variable "logs_bucket_versioning" {
 
 variable "managed_waf_rules" {
   description = "List of parameters for managed WAF rules"
-  type        = list(object({
+  type = list(object({
     name              = string
     priority          = string
     metric_visibility = bool
--- a/terraform/modules/gost_website/variables.tf
+++ b/terraform/modules/gost_website/variables.tf
@@ -41,6 +41,15 @@ variable "logs_bucket_versioning" {
   default     = true
 }
 
+variable "managed_waf_rules" {
+  description = "List of parameters for managed WAF rules"
+  type        = list(object({
+    name              = string
+    priority          = string
+    metric_visibility = bool
+  }))
+}
+
 variable "origin_bucket_dist_path" {
   description = "Path to the directory where website build files should be stored in the S3 origin bucket."
   default     = "/dist"
--- a/terraform/modules/gost_website/variables.tf
+++ b/terraform/modules/gost_website/variables.tf
@@ -45,7 +45,7 @@ variable "managed_waf_rules" {
   description = "List of parameters for managed WAF rules"
   type = list(object({
     name              = string
-    priority          = string
+    priority          = number
     metric_visibility = bool
   }))
 }
--- a/terraform/modules/gost_website/waf.tf
+++ b/terraform/modules/gost_website/waf.tf
@@ -1,9 +1,9 @@
 locals {
   expanded_rules = flatten([
     for r in var.managed_waf_rules : {
-      name              = r.name,
-      priority          = r.priority,
-      statement         = {
+      name      = r.name,
+      priority  = r.priority,
+      statement = {
         name        = r.name
         vendor_name = "AWS"
       },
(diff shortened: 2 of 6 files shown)


terraform/modules/gost_website/variables.tf
```diff
@@ -41,6 +41,15 @@ variable "logs_bucket_versioning" {
   default     = true
 }
 
+variable "managed_waf_rules" {
+  description = "List of parameters for managed WAF rules"
+  type = list(object({
+    name              = string
+    priority          = string
```
TylerHendrickson, line 48: Based on the usage, it looks like the `priority` type should be `number`.

slapula:
@TylerHendrickson just a heads up,  these rules may break staging in unique and interesting ways.  If this deploys successfully to staging, lets give that environment a good once-over to make sure nothing is out of sorts.

terraform/modules/gost_website/variables.tf
```diff
@@ -41,6 +41,15 @@ variable "logs_bucket_versioning" {
...
+    metric_visibility = bool
+  }))
```
TylerHendrickson, line 50: This is admittedly a trivial nit, but since the name of each rule needs (I think?) to be unique, it might make sense to implement this variable as a `map` type keyed by the `name`, which would ensure that the same rule name could only be used once. 

```suggestion
  type = map(object({
    priority          = string
    metric_visibility = bool
  }))
```

TylerHendrickson (changes requested):

terraform/staging.tfvars
```diff
@@ -7,6 +7,28 @@ ssm_deployment_parameters_path_prefix = "/gost/staging/deploy-config"
...
+    managed_rule      = "AWSManagedRulesCommonRuleSet",
+    priority          = 20,
+    metric_visibility = true
+  },
+  "KnownBadInputsRuleSet" = {
+    managed_rule      = "AWSManagedRulesKnownBadInputsRuleSet",
+    priority          = 30,
+    metric_visibility = true
+  }
+}
```
TylerHendrickson, line 31: Do you think there's a benefit to having separate `managed_rule` attribute here? Would something like this work?
```suggestion
website_managed_waf_rules = {
  "AWSManagedRulesAnonymousIpList" = {
    priority          = 5,
    metric_visibility = false
  },
  "AWSManagedRulesAmazonIpReputationList" = {
    priority          = 10,
    metric_visibility = false
  },
  "AWSManagedRulesCommonRuleSet" = {
    priority          = 20,
    metric_visibility = true
  },
  "AWSManagedRulesKnownBadInputsRuleSet" = {
    priority          = 30,
    metric_visibility = true
  }
}
```

terraform/modules/gost_website/waf.tf
```diff
@@ -0,0 +1,26 @@
...
+        name        = r.managed_rule
+        vendor_name = "AWS"
+      },
+      visibility_config = {
+        cloudwatch_metrics_enabled = r.metric_visibility,
+        sampled_requests_enabled   = r.metric_visibility,
+        metric_name                = "${r.key}-metric"
+      }
+    }
+  ])
```
TylerHendrickson, line 16: Per my [other comment](https://github.com/usdigitalresponse/usdr-gost/pull/1177/files#r1156376153):
```suggestion
  expanded_rules = flatten([
    for r in var.managed_waf_rules : {
      name     = r.key,
      priority = r.value.priority,
      statement = {
        name        = r.key
        vendor_name = "AWS"
      },
      visibility_config = {
        cloudwatch_metrics_enabled = r.value.metric_visibility,
        sampled_requests_enabled   = r.value.metric_visibility,
        metric_name                = "${r.key}-metric"
      }
    }
  ])
```

terraform/staging.tfvars
slapula: Yeah, I thought about that but AWS considers `name` separate from the actual name of the managed rule set you are selecting.  The `name` attribute is what is checked for uniqueness (thus you can have several different rules using the same managed `AWSManagedRulesAnonymousIpList` ruleset).  The use case here is that you can deploy several different versions the the ruleset all with different parameters (most of which I'm not exposing here since I'm not sure that lever of granularity is what we need at the moment).

terraform/modules/gost_website/variables.tf
slapula: I actually like this idea better than what I originally had. Much cleaner and easier to understand.

terraform/staging.tfvars
TylerHendrickson: Ah, gotcha- makes sense. Carry on!

slapula:
@TylerHendrickson I'm satisfied with this PR for now.  I poked around in the staging environment and it looks good as far as I can tell.  Do we have any automated testing in place to give us some more thorough feedback?

TylerHendrickson:
> @TylerHendrickson I'm satisfied with this PR for now. I poked around in the staging environment and it looks good as far as I can tell. Do we have any automated testing in place to give us some more thorough feedback?

@slapula Currently, we don't have automated tests (a la `terratest`) in place for Terraform changes. Once this PR merges and the changes deploy to staging, the site at https://staging.grants.usdr.dev would be the best place to do functional spot-checks. NB: We are working on adding Cypress for integration tests, but I can't think of any explicit test cases that should be added for this PR.

PR #1185, opened 2023-04-04
fix: WAFv2 must be managed in us-east-1 if globally scoped

slapula:
## Description
See title.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/terraform/modules/gost_website/waf.tf
+++ b/terraform/modules/gost_website/waf.tf
@@ -23,5 +23,10 @@ module "waf" {
 
   managed_rule_group_statement_rules = local.expanded_rules
 
+  # WAFv2 must be managed in us-east-1 if globally scoped
+  providers = {
+    aws = aws.cloudfront_certificate
+  }
+
   context = module.this.context
 }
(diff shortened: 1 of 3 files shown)


terraform/modules/gost_website/waf.tf
```diff
@@ -23,5 +23,10 @@ module "waf" {
 
   managed_rule_group_statement_rules = local.expanded_rules
 
+  # WAFv2 must be managed in us-east-1 if globally scoped
+  providers = {
+    aws = aws.cloudfront_certificate
```
slapula, line 28: Since I'm re-using this now, should we consider renaming it to something less specific? @TylerHendrickson

TylerHendrickson: Yeah, good call – maybe setting the alias to something like `us-east-1` (I think `-` characters are okay? can use underscores if not) would be more appropriate.

Would you also mind moving that aliased `provider` definition block from `certificate.tf` to `main.tf` in this module?

TylerHendrickson (changes requested):

TylerHendrickson (approved):

PR #1186, opened 2023-04-04
feat: enable query logging for Postgres cluster

slapula:
## Description
See title.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


as1729 (approved):
Looks good 👍 

The AWS docs seem to suggest that we should minimize the use of logging for sensitive fields. I think we should make sure PII like email/name/etc are added to the `ENCRYPTED` option when making a query for these fields. wdyt?

slapula:
@as1729 Yeah, I agree with that.  In fact, I'd probably want to hold off on merging this until that masking is rolled out. @TylerHendrickson thoughts?

TylerHendrickson:
@slapula @as1729 Did some thinking on this...

Technically, all of the information stored in our Postgres database is public domain, with the only exception being passwords/access tokens. That said, I'd agree with the best practice recommendation that it's better to keep values like names and email addresses out of logs. One thing I'm not sure about is whether this whether the statement query logs will include this information, since they should be passed to queries as parameters rather than literal values.

My preference would be to start by rolling this out to staging and doing a manual audit to see whether email addresses used in testing show up in the logs. If it turns out that they are getting logged, we can resolve by masking log data by configuring a [CloudWatch data protection policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_data_protection_policy for the log group in a separate PR.

as1729:
Makes sense @TylerHendrickson! 👍

slapula:
@TylerHendrickson @as1729 Makes sense to me as well. With that said, this PR needs another look since I made a few improvements.  I changed it to accommodate a conditional to allow us to control which environments have this feature enabled.

TylerHendrickson (approved):

PR #1202, opened 2023-04-08
1161 aws module

BatMiles:
## Description
This PR condenses two files that deal with creating AWS infrastructure. The resulting file is used in all tests and infrastructure creation.

## Testing
Currently existing tests should run.
Infrastructure should be created as normal.

### Automated and Unit Tests
- [ ] Added Unit tests
^ Unit tests previously existed for both pre-existing files, they've been adjusted to use the new file

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/src/lib/gost-aws.js
```diff
@@ -63,6 +92,7 @@ function send(message) {
         .catch((err) => console.error('Error sending SES email:', err, err.stack));
 }
 
-module.exports = { send };
-
-/*                                  *  *  *                                   */
+module.exports = {
+    getS3Client,
+    send,
```
TylerHendrickson, line 97: I think it would be useful to rename this function and its exported name to something a bit more verbose, like `sendEmail`.

TylerHendrickson (changes requested):
Existing work looks great! I'm requesting a few additional changes that I think are looking more relevant/useful in context of this move. Let me know if you have any questions on any of them.

packages/server/src/lib/gost-aws.js
```diff
@@ -1,11 +1,40 @@
...
+    return s3;
+}
+
+/*
+----------------------------------------------------------
+                        AWS SES
+----------------------------------------------------------
+*/
 
 function createTransport() {
```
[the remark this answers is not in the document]
BatMiles: @TylerHendrickson makes sense! I've addressed this and your other comments

packages/server/src/lib/gost-aws.js
```diff
@@ -0,0 +1,72 @@
...
+                    Charset: 'UTF-8',
+                    Data: message.text,
+                },
+            },
+        },
+    };
+    transport.sendEmail(params).promise()
+        .then((data) => console.log('Success sending SES email:', data))
+        .catch((err) => console.error('Error sending SES email:', err, err.stack));
+}
```
TylerHendrickson, line 67: No change needed for this right now, but noting for future reference: I'm thinking it would be better for the `.catch()` path to log and re-throw the error so that callers of `sendEmail()` are responsible for handling the error in context of the operation that's sending a specific email (which may include re-trying, formulating a response based on the error, etc.).

BatMiles: That would also make this more testable - we can directly test for the contents of an error, but can't check a console log without returning the promise from this function.

PR #1206, opened 2023-04-10
Remove legacy agencies.main_agency_id

pearkes:
## Description

This removes references to `agencies.main_agency_id` as we had to workaround this cyclical dependency in various places.

I wanted to call-out two areas I'd appreciate review, as tests are passing but I'm wanting to be sure I'm not disrupting intent. I will comment inline on those.

### Manual tests for Reviewer
- [ ] Confirm the workflow for a tenant to have a main_agency_id set by creating a tenant and assigning it to an agency

## Checklist
- [x] Added PR reviewers


packages/server/seeds/dev/01_main.js
```diff
@@ -63,14 +63,18 @@ exports.seed = async (knex) => {
...
-                        WHERE main_agency_id = id
-                    ) UPDATE tenants
-                      SET main_agency_id = mal.id
-                      FROM main_agency_lookup mal
-                      WHERE mal.tenant_id = tenants.id`);
+    //
+    // TODO (pearkes): How should this be modified with the removal of
+    // main_agency_id? It does look like we're using this to set
+    // the tenant.main_agency_id. Should we be reading the parent?
+    // await knex.raw(`WITH main_agency_lookup AS (
```
pearkes, line 70: Can this query be removed? I believe this is redundant now but we're setting `tenants.main_agency_id` here with the of the `agencies` lookup. Should we just do a select on `id = id` for agencies here?

packages/server/src/db/index.js
```diff
@@ -138,7 +137,7 @@ async function getUser(id) {
         )
         .leftJoin('roles', 'roles.id', 'users.role_id')
         .leftJoin('agencies', 'agencies.id', 'users.agency_id')
-        .leftJoin('tenants', 'tenants.main_agency_id', 'agencies.main_agency_id')
+        .leftJoin('tenants', 'tenants.main_agency_id', 'agencies.id')
```
pearkes, line 140: Ensuring this join is as intended here. I believe this is the fk relationship but not 100% on the intent.

packages/server/seeds/dev/01_main.js
TylerHendrickson: > Should we just do a select on id = id for agencies here?
I might be missing something here... From what I can see, we still need a step that sets `tenant.main_agency_id` somewhere in the seeding process, so I don't think it can be removed entirely. 

My instinct is that the correct `tenant.main_agency_id` value should be able to be determined by checking for the "root" agency on each tenant, e.g.
```sql
WITH main_agency_lookup AS (
  SELECT id, tenant_id
  FROM agencies
  WHERE parent = 0
) UPDATE tenants 
  SET main_agency_id = mal.id
  FROM main_agency_lookup mal
  WHERE mal.tenant_id = tenants.id;
```
The problem with the above query is that it assumes that there's only one agency per tenant with `parent = 0`, but based on the seed data we're currently using, there are two issues:
1. there are two agencies (one with `name = "Test sub USDR", tenant_id = 1` and another with `name = "B Test sub USDR", tenant_id = 1`) that have the same `tenant_id` value. 
2. It leaves one tenant (with `name = "NPO Tenant"`) without a value for `main_agency_id` because there are no agencies associated with that tenant.

What I'm not sure about is whether the seed data is incorrect (which implies we should have some unique constraint that prevents this situation) or if there's a gap in my understanding of the significance of a "main" agency.

CCing @as1729 to see if his understanding is different than mine. If not, we may want to check with @mhuang-usdr to see if she can clarify the intended business logic here.

packages/server/src/db/index.js
TylerHendrickson: It looks like this query is simply using this join to get information about the tenant associated with the user's agency. The agency associated with a user might not always be the main agency within that tenant, so I think this is supposed to be:
```suggestion
        .leftJoin('tenants', 'tenants.id', 'agencies.tenant_id')
```

TylerHendrickson (changes requested):
One change to request (per your question on `packages/server/src/db/index.js`) and another that I'm pretty sure will require some change, but (as noted in my comment) I'm not positive what it needs to be just yet.

packages/server/src/db/index.js
TylerHendrickson: Actually, on second glance, there's a more direct path to the tenant associated with the user:
```suggestion
        .leftJoin('tenants', 'tenants.id', 'users.tenant_id')
```

TylerHendrickson (changes requested):
@pearkes I had a chance to check in with @as1729 today, who was able to confirm my take an elaborate on things a bit.

A few things to call out:
1. The assumptions behind the SQL I provided in my [previous comment](https://github.com/usdigitalresponse/usdr-gost/pull/1206#discussion_r1162099278) are correct. The fact that the current outcome is **not** correct exposes an issue with the fixtures. 
    - NB: I'm not sure if is only now a problem with the `agencies.main_agency_id` column removed or if this is just surfacing an existing issue that's gone unnoticed until now. 
2. The expected business rule is: each tenant should have exactly 1 agency (the "main" agency) whose parent is USDR (`agency` row with `parent = 0`).
3. This isn't an issue we'd normally run into during tenant creation in production, because the input for that process specifies the main agency to create along with the new tenant, so there's never any ambiguity as far as which agency should be set as the `main_agency_id` on the tenant.
4. However, this scenario could potentially cause problems when evaluating user access:
    - Tenant A has three agencies: Agency `X`, `Y`, and `Z`.
    - `TenantA.main_agency_id = AgencyX.id`
    - `AgencyX.parent  = 0`, `AgencyY.parent = AgencyX.id`, `AgencyZ.parent = 0`

To resolve, I'm proposing the following changes:
1. Fix the invalid seed data for the fixture agency with ID `402` (defined [here](https://github.com/usdigitalresponse/usdr-gost/blob/ad57c55f321eaffeeef3a9aab2c2939f51f604f3/packages/server/seeds/dev/ref/agencies.js#L27-L34)) by setting its `parent` to `400` (currently set to `0`).
2. Either / both:
    - Update the file `packages/server/seeds/dev/ref/tenants.js` so that each tenant's seed data (the member values of the `module.exports` array) has hard-coded `main_agency_id` values. Since all of our fixture data has hard-coded identifiers, this shouldn't cause any problems.
    - Update the file `packages/server/seeds/dev/01_main.js` to incorporate the query I suggested (or something similar to it) in my [prior comment](https://github.com/usdigitalresponse/usdr-gost/pull/1206/files#r1162099278).
3. Add a knex migration to define a new constraint on the `agencies` table that satisfies the requirement that exactly 1 agency per-tenant must have `parent = 0`.
    - NB: It might be more trouble than it's worth to enforce that a tenant with only 1 agency (i.e. rows in `agencies` where `tenant_id` is unique) must have `parent_id = 0` – I think a solution that satisfies a requirement that "**up to 1** agency per-tenant must have `parent = 0`" would be acceptable.
    - NB: I don't love the idea of treating `0` as a magic number in a DB constraint, but offhand I'm having a hard time coming up with a good alternative. Improvements are welcome!

There's an argument that `#3` is out of scope for this PR, so if you'd prefer, we can create a different ticket to address adding the new constraint.

pearkes:
@TylerHendrickson Thanks for the detailed feedback. I agree about #3 with a constraint being a good idea, but going to follow-on with that and have created a new issue here: #1239. My only tweak there would be instead of the magic number possibly just do a boolean on agencies that marks that agency as special in that way.

Pushed changes from feedback. Do you mind taking another look at this @TylerHendrickson?

packages/server/src/db/index.js
TylerHendrickson: I think this one still needs to be updated- might have been resolved prematurely. 

Suggest joining the `tenants` table where `tenants.id = users.tenant_id` ([second suggestion](https://github.com/usdigitalresponse/usdr-gost/pull/1206/files#r1162110418) from the above comments).

TylerHendrickson (changes requested):
One quick change that I think was left out previously (if intentional, please educate me 🙂).

TylerHendrickson (approved):
🚀

PR #1243, opened 2023-04-17
Upgrading AWS SDK to v3

### LINKED ISSUE #653: Upgrade `aws-sdk` to v3
Currently, the server package makes use of the [2.x version of the AWS SDK for JavaScript](https://github.com/aws/aws-sdk-js#aws-sdk-for-javascript), which is scheduled to be deprecated to maintenance mode in 2023. 

The [3.x version of the AWS SDK for JavaScript](https://github.com/aws/aws-sdk-js-v3) is the recommended replacement library. In addition to remaining current, this version of the SDK is [modularly structured](https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/), which helps keep the size of packages down by only building with SDK components that are actually in-use. As the SDK in its entirety is rather expansive, and at present we only use the SES client, migrating to the new version should yield a not-insignificant (roughly 72M) reduction in the `node_modules` footprint.

Upgrade path is documented [here](https://aws.amazon.com/blogs/developer/why-and-how-you-should-use-aws-sdk-for-javascript-v3-on-node-js-18/).

jakekreider:
Added AWS SDK v3 libraries, removed legacy aws-sdk. 

I'm rusty in the stack here and appreciate any feedback!

Closes #653.

## Description
Migrates AWS SDK to v3

## Testing
1. Launch locally with localstack using configured values added in .env.example
2. Request login link. Email should be found in volume/tmp/state/ses/.
3. Navigate to /arpa_reporter
4. Click "Send audit report by email"
5. Aforementioned email location should include an email with localstack S3 link.

### Automated and Unit Tests
- [ ] Added Unit tests **ed note, modified unit tests**

### Manual tests for Reviewer

## Checklist
- [ ] Provided screenshots/demo (**ed note, no functional change**)
- [x] Added PR reviewers


packages/server/src/lib/gost-aws.js
```diff
@@ -18,27 +19,48 @@ function getS3Client() {
...
         console.log('------------ USING LOCALSTACK ------------');
-        const endpoint = new AWS.Endpoint(`http://${process.env.LOCALSTACK_HOSTNAME}:${process.env.EDGE_PORT || 4566}`);
-        s3 = new AWS.S3({
+        const endpoint = `http://${process.env.LOCALSTACK_HOSTNAME}:${process.env.EDGE_PORT || 4566}`;
+        console.log(`endpoint: ${endpoint}`);
+        s3 = new S3Client({
             endpoint,
-            s3ForcePathStyle: true,
+            forcePathStyle: true,
+            region: 'us-west-2',
```
TylerHendrickson, line 27: I'd [prefer](https://github.com/usdigitalresponse/usdr-gost/pull/1202#discussion_r1161909931) to avoid hard-coding regions, and defer to the [AWS SDK's documented behavior](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-region.html#setting-region-order-of-precedence) for selecting the appropriate region (e.g. from `$AWS_REGION`). 

If you think not having a region specified in environments making use of LocalStack is a likely sticky point for developers, I propose setting a default in `packages/server/.env.example`, like this:
```env
AWS_REGION="${AWS_REGION:-us-west-2}"
```

packages/server/src/lib/gost-aws.js
```diff
@@ -18,27 +19,48 @@ function getS3Client() {
...
                         AWS SES
 ----------------------------------------------------------
 */
 
+function getSESClient() {
+    const sesOptions = {};
+
+    if (process.env.LOCALSTACK_HOSTNAME) {
+        sesOptions.endpoint = `http://${process.env.LOCALSTACK_HOSTNAME}:${process.env.EDGE_PORT || 4566}`;
+        sesOptions.region = 'us-west-2';
```
TylerHendrickson, line 54: (See previous comment regarding hard-coded regions.)

docker-compose.yml
```diff
@@ -34,11 +34,13 @@ services:
     environment:
       - DEBUG=${DEBUG-}
       - DOCKER_HOST=unix:///var/run/docker.sock
+      - HOSTNAME_EXTERNAL=localhost.localstack.cloud
```
TylerHendrickson, line 37: Can you elaborate on the reasoning behind this? It looks like the LocalStack [default](https://docs.localstack.cloud/references/configuration/) is `localhost` – wouldn't this require a LocalStack cloud account?

TylerHendrickson (changes requested):

packages/server/src/lib/gost-aws.js
jakekreider: Good call, totally agree.

docker-compose.yml
jakekreider: This shouldn't have made it into my non-draft PR, will remove!

packages/server/.env.example
```diff
@@ -6,6 +6,7 @@ POSTGRES_TEST_URL=postgresql://localhost:5432/usdr_grants_test
 COOKIE_SECRET=itsasecretsecretsecret
 
 WEBSITE_DOMAIN=http://localhost:8080
+API_DOMAIN=http://localhost:8080
```
jakekreider, line 9: I noticed that without this, the email URL has an `undefined` host.

docker-compose.yml
```diff
@@ -34,11 +34,13 @@ services:
     environment:
       - DEBUG=${DEBUG-}
       - DOCKER_HOST=unix:///var/run/docker.sock
+      - AWS_DEFAULT_REGION=${AWS_REGION:-us-west-2}
     networks:
       - app
     volumes:
       - "${LOCALSTACK_VOLUME_DIR:-./volume}:/var/lib/localstack"
       - "/var/run/docker.sock:/var/run/docker.sock"
+      - "${PWD}/localstack/entrypoint/init-aws.sh:/etc/localstack/init/ready.d/init-aws.sh"
```
TylerHendrickson, line 43: I think leaving out `${PWD}` and using a relative path should be basically equivalent, but avoid potential ambiguity/issues from running `docker compose` from another directory.

```suggestion
      - "localstack/entrypoint/init-aws.sh:/etc/localstack/init/ready.d/init-aws.sh"
```

TylerHendrickson (changes requested):
One final nit, but overall looking great!

docker-compose.yml
jakekreider: Sounds good!

Issue #1238, opened 2023-04-14
Use IAM auth when connecting to Postgres


TylerHendrickson:
**DOD:** In AWS environments, we use the [AWS JS SDK's `@aws-sdk/rds-signer` module](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_rds_signer.html) to generate a short-lived token to use as the password component of the Postgres DB connection string. 

**Acceptance criteria:**
- The Postgres cluster is no longer provisioned with static credentials.
- We no longer provision an SSM `SecureString`-type parameter for storing a Postgres connection string.
- The `POSTGRES_URL` environment variable in API containers is set in plaintext (using `map_environment` in the task definition instead of `map_secrets`) without a password.

**Implementation details:**
- Configure the `module.postgres.module.db` module resource without static credentials.
- Deprovision the `module.api.aws_ssm_parameter.postgres_connection_string` resource.
- Update the `POSTGRES_URL` environment variable definition for API containers according to the above acceptance criteria.
- The Knex connection flow in `packages/server/knexfile.js` and/or `packages/server/src/db/connection.js` will need to be updated with the IAM auth flow documented [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_rds_signer.html#generate-authentication-token-for-rds-iam-authentication).

**Background:** 
[...]

PR #1263, opened 2023-04-19
1238: Enable IAM auth for RDS Postgres

TylerHendrickson:
## Description

This PR implements changes outlined in #1238, differing slightly in that it does not completely prevent the use of static credentials – since RDS/Aurora Postgres will _only_ accept IAM-based authentication once enabled (although it can be disabled again), the implementation provided by this PR modify the behavior of the API server so that it first attempts to connect to Postgres using IAM authentication, but gracefully falls back to basic/standard authentication (i.e. how we currently connect) if the initial IAM auth attempt is unsuccessful. The benefit of this approach is that it allows us to roll out these changes before we enable IAM auth on our Postgres DB, thereby avoiding downtime or other complicated/synchronized deployment activities.

Stated more simply, the changes in this PR allow us to transition to IAM auth according to the following steps:
1. Deploy this code to an environment that uses RDS.
2. Wait for ECS to pick up the changes.
3. Turn on IAM auth in Postgres (`GRANT rds_iam TO postgres;`)
4. The next time an ECS task creates a new connection, it will use IAM auth.

Once the above steps are complete, a follow-up PR can be made to remove the "basic auth fallback" behaviors entirely (if we think that's useful).

**Summary of changes:**
- Terraform changes:
  - The `gost_postgres` module is updated to incorporate the major version bump for [terraform-aws-modules/rds-aurora/aws](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora) from 7.7.1 to 8.0.2, initially introduced in [this dependabot PR](https://github.com/usdigitalresponse/usdr-gost/pull/1226). 
      - Breaking changes related to the major version upgrade have been addressed according to the the v7.x -> v8.0 upgrade path documented [here](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/e054f77c700128c206b0912c51712313981d6fae/UPGRADE-8.0.md).
  - The API server ECS task definition in the `gost_api` module has been updated to include a new environment variable, `PGSSLROOTCERT`, which points to the (global) [RDS root certificate bundle](https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem) that Knex will use when connecting to RDS Postgres databases using SSL.
- API server package (JS) changes:
  - Added a new AWS SDK dependency, `@aws-sdk/rds-signer`, which is used to generate signed IAM auth tokens.
  - Added `packages/server/rds-combined-ca-bundle.pem` (downloaded from [here](https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem)) to be tracked in git.
  - Updated `packages/server/knexfile.js` to use IAM auth when available, and fall back to basic auth when it isn't. The overall flow for this behavior is as-follows:
    - Postgres connection parameters will be sourced from [standard Postgres environment variables](https://www.postgresql.org/docs/current/libpq-envars.html) and/or from values parsed from the `$POSTGRES_URL` environment variable (preference is given to the more granular Postgres environment variables).
    - SSL for the Postgres connection is configured as follows:
      - If the `$PGSSLMODE` environment variable is set to `"disabled"`, then SSL configuration is bypassed entirely (which may be useful/necessary in local testing environments).
      - If any of `$PGSSLROOTCERT`, `$PGSSLKEY`, and/or `$PGSSLCERT` environment variables are set, the contents of the file that each env var references are mapped to their corresponding SSL configuration options documented [here](https://node-postgres.com/features/ssl).
    - If an auth token cannot be generated by the `rds-signer` client, Knex uses the password provided by `$POSTGRES_URL`.
    - Once an IAM auth token is generated, a test connection is established and used to issue a `SELECT 1;` test query to check connectivity. If the query/connection fails, Knex is configured with the password provided by `$POSTGRES_URL`.
    - If the test query/connection succeeds, Knex is configured with IAM auth credentials.
- Extra/miscellaneous changes:
  - Added the RDS CA bundle to `CODEOWNERS` so that admin approval is required for PRs where this file is modified (which should be pretty rare).
  - Fixed a bug affecting local Docker dev environments introduced in #1243 (this is my fault- the suggestion I made during PR review had the wrong `docker-compose.yml` syntax for mounting a volume using a relative path).
  - Readability tweak to the API ECS task definition in the `gost_api` terraform module so that environment variables are listed alphabetically.

## Testing

Fallback behavior can be tested locally, but complete testing inherently requires an accessible RDS/Aurora Postgres database that has granted `rds_iam` to the user. I needed to spin up a sandbox environment to do complete manual testing, which included toggling between granting/revoking `rds_iam` to ensure that Knex was successfully configured under both scenarios.

### Automated and Unit Tests
- [ ] Added Unit tests

I did **not** include unit tests for these changes. I'm open to suggestions on how this might be accomplished, although we haven't historically included unit tests around the database-bootstrapping path for production environments.

It's perhaps worth noting that the end-to-end integration tests run by Cypress do cover some parts of the new bootstrapping process – specifically the non-SSL, basic-auth-fallback path.

### Manual tests for Reviewer

## Checklist
- [X] Added PR reviewers


packages/server/knexfile.js
```diff
@@ -33,7 +36,70 @@ module.exports = {
...
+                        return false;
+                    }
+                    const sslConfig = Object.fromEntries(Object.entries({
+                        // We assume that the below env vars will point to a file if/when set.
+                        ca: process.env.PGSSLROOTCERT,
+                        key: process.env.PGSSLKEY,
+                        cert: process.env.PGSSLCERT,
+                    })
+                        // Filter the above keys to those with a truthy value
+                        .filter(([, v]) => v)
```
TylerHendrickson, line 67: Props to @Samcbehrens for helping me avoid spinning my wheels on this!

packages/server/knexfile.js
```diff
@@ -33,7 +36,70 @@ module.exports = {
...
+            const iamAuthConfig = {
+                ...basicAuthConfig,
+                password: token,
+                expirationChecker: () => tokenExpiration <= new Date(),
+            };
+
+            try {
+                // Test IAM authentication
+                // eslint-disable-next-line global-require
+                await require('knex')({ client: 'pg', connection: iamAuthConfig }).raw('select 1');
```
TylerHendrickson, line 96: IMO this is a little gross/overly clever, but given our conventions around requiring/exporting a fully-configured Knex instance from `src/db/connection.js`, this seemed the best method to run this test query and implement the fallback behavior without turning that instance into a promise (which would require quite a bit of refactoring and net additional complexity everywhere we use Knex). I explored a few alternative approaches in `src/db/connection.js`, but they all ultimately required some use of `await` in the course of producing the exported value. 

At the end of the day, this strikes me as the least-worst option, although I'm definitely open to suggestions.

packages/server/rds-combined-ca-bundle.pem
TylerHendrickson, line 1: There are a few options I considered as alternatives to tracking this file in git:
1. Updating `docker/production-api.Dockerfile` to download and add the file to the runtime environment during build-time. Given that this file should rarely (if ever) need to be updated, I didn't like the variability that bundling on every build would entail, especially since it's hard to test the correctness of the file during the build process.
2. Mounting it in the ECS task would require an entire filesystem, which feels unjustified for a single file that basically never changes.

terraform/.tflint.hcl
TylerHendrickson, line 1: This isn't used in any automation yet, but I figured it was worth including since we've started using `tflint` in the [grantes-ingest](https://github.com/usdigitalresponse/grants-ingest/) repo and I'd like to eventually include it in our CI workflows for this repo as well.

terraform/modules/gost_api/task.tf
```diff
@@ -32,25 +32,26 @@ module "api_container_definition" {
...
       NODE_OPTIONS              = "--max_old_space_size=1024"
-      API_DOMAIN                = "https://${var.domain_name}"
+      NOTIFICATIONS_EMAIL       = var.notifications_email_address
+      PGSSLROOTCERT             = "rds-combined-ca-bundle.pem"
       VUE_APP_GRANTS_API_URL    = module.api_gateway.apigatewayv2_api_api_endpoint
       WEBSITE_DOMAIN            = "https://${var.website_domain_name}"
-      NOTIFICATIONS_EMAIL       = var.notifications_email_address
-      DATA_DIR                  = "/var/data"
-      AUDIT_REPORT_BUCKET       = module.arpa_audit_reports_bucket.bucket_id
     },
```
TylerHendrickson, line 47: NB: The only actual change here is the addition of the `PGSSLROOTCERT` env var. Everything else is just the result of sorting the keys alphabetically.

terraform/modules/gost_postgres/main.tf
```diff
@@ -29,10 +29,15 @@ resource "aws_rds_cluster_parameter_group" "postgres13" {
   }
 }
 
+resource "random_password" "postgres_user" {
+  length  = 32
+  special = false
+}
+
```
TylerHendrickson, line 36: This does force a new password to be generated, but should **not** invalidate existing connections, so it shouldn't cause any downtime.

terraform/modules/gost_postgres/main.tf
```diff
@@ -45,15 +50,22 @@ module "db" {
...
+  create_db_subnet_group         = true
   security_group_use_name_prefix = true
-  allowed_security_groups        = var.allowed_security_groups
-  allowed_cidr_blocks            = var.allowed_cidr_blocks
+  security_group_rules = {
+    for k, security_group_id in var.ingress_security_groups : k => {
+      type                     = "ingress"
+      source_security_group_id = security_group_id
+    }
+  }
```
TylerHendrickson, line 60: The [v7.x -> v8.x docs for this module](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/e054f77c700128c206b0912c51712313981d6fae/UPGRADE-8.0.md#security-group-rules-migration) includes a [section](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/e054f77c700128c206b0912c51712313981d6fae/UPGRADE-8.0.md#security-group-rules-migration) for highlighting the caveats involved in this particular change. 

In my experience from testing in a sandbox environment, this actually caused `terraform apply` to fail, although the failure itself didn't cause any perceivable issues. While not ideal, the most straightforward way to deal with this is to let `terraform apply` fail initially, and then run `terraform apply` again- I didn't need to actually do any manual state manipulation (e.g. `terraform state mv ...`) to resolve things.

terraform/modules/gost_postgres/variables.tf
```diff
@@ -25,15 +25,9 @@ variable "subnet_ids" {
...
 
-variable "allowed_security_groups" {
-  description = "Security Group IDs which should be allowed to connect (ingress) to Postgres."
-  type        = list(string)
-}
-
-variable "allowed_cidr_blocks" {
-  description = "CIDR blocks which should be allowed to connect (ingress) to Postgres."
-  type        = list(string)
-  default     = []
```
TylerHendrickson, line 36: Removed this because we aren't using it – all our ingress rules are based around security groups.

jakekreider (approved):
I don't have a sandbox to test IAM, but I tested both fallback scenarios locally and the changes make sense to me. Thanks for the detailed explanation!

packages/server/knexfile.js
```diff
@@ -33,7 +36,70 @@ module.exports = {
...
+            };
+
+            // Attempt to get a signed token for IAM auth or fall back to basic auth
+            const tokenExpiration = new Date();
+            const signer = new Signer({ hostname, port, username });
+            const token = await signer.getAuthToken().then((authToken) => {
+                // Auth tokens are valid for 15 minutes (900 seconds)
+                tokenExpiration.setSeconds(tokenExpiration.getSeconds() + 900);
+                return authToken;
+            }).catch(() => { });
```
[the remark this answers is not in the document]
TylerHendrickson: The failure modes basically come down to a misconfigured SDK/environment – an unset `AWS_REGION`, for example, or a missing/malformed parameter in the `Signer` instantiation. However, having the right information on-hand would be very useful in a situation where both IAM auth failed and then the subsequent fallback path failed as well – I could see a scenario where we've already enabled IAM auth on the Postgres side but "for some reason" we can't get a signed token, and having some logs here would help narrow down that reason. 

Thanks for the call-out, I'll add some logs here.

Issue #1218, opened 2023-04-11
[SPIKE] State Management


NatHillardUSDR:
# Description
As we move toward a new React-based stack, we should look into the various state management libraries that are available and make a decision about which to go forward with.

- We currently use [Vuex](https://vuex.vuejs.org/) - we would need to move toward a different library in the React world.
- Tentative Example options:
    - Just [useReducer](https://react.dev/reference/react/useReducer) 
    - [Redux](https://react-redux.js.org/) + [Redux Toolkit](https://redux-toolkit.js.org/)
    - [Zustand](https://docs.pmnd.rs/zustand/getting-started/introduction)

# Acceptance Criteria
* We have copied [this](https://www.notion.so/usdr/Duplicate-me-Architecture-Decision-Record-Template-79ad657a737f47838741017c743a4b4f) Notion template to a subpage of the [Tech Modernization page](https://www.notion.so/usdr/Tech-Modernization-b0f8e3736ebc4e9a898dad2b4ca2f858) named the same as this issue and have filled out the various sections appropriately
* We have added it as a link in the top-level [Tech Modernization page](https://www.notion.so/usdr/Tech-Modernization-b0f8e3736ebc4e9a898dad2b4ca2f858)

Samcbehrens:
[Associated Notion doc](https://www.notion.so/usdr/WIP-Migrating-State-management-1ccff6b630894709bdc394aac560e81d)

NatHillardUSDR:
PR [here](https://github.com/usdigitalresponse/usdr-gost/pull/1314) for GitHub version

PR #1314, opened 2023-05-03
[DECISION] Use Redux for State Management

NatHillardUSDR:
Issue: #1218\n Ports Notion decision doc around using Redux for state management


Samcbehrens (approved):
Thanks for porting this over. 👍

Samcbehrens:
I just want to highlight my final point in this doc: I recommend using redux only because I am not sure complicated our current state management tooling is. Redux gives us the tools for more complicated state management so I am mostly erring on the side of caution.

 I believe Zustand is a great option but it tends to work well with more straightforward apps, it may be a good fit for our needs, but since I have little knowledge of our current state management complexity, I am recommending redux.

Issue #1395, opened 2023-05-19
BUG: FK constraint violation when deleting agencies


TylerHendrickson:
**Description:** An agency that has assigned users assigned cannot currently be deleted. Instead these requests to the `DELETE /api/organizations/:organizationId/agencies/del/:agency` endpoint fail with a 500 status code because of the following error being thrown on the backend: 
> `error: delete from "agencies" where "id" = $1 and "parent" = $2 and "name" = $3 and "abbreviation" = $4 and "warning_threshold" = $5 and "danger_threshold" = $6 - update or delete on table "agencies" violates foreign key constraint "email_subscriptions_agency_id_foreign" on table "email_subscriptions"`

**Steps to reproduce:**
1. Login as a user with permission to create agencies (e.g. tenant admin).
2. Create a new agency for the tenant.
3. Create a new user assigned to the newly-created agency.
4. Ensure that the browser's developer console is recording requests, and attempt to delete the newly-created agency. Observe that although the modal closes, the request fails with a 500 response code and the error message documented above is logged.
5. Delete the newly-created user.
6. Once again, attempt to delete the newly-created agency. Observe that the deletion is successful.

**Impacts:**
- User experience: there is no message communicating that the agency failed to delete.
- User experience: it is unclear that an agency with associated users cannot be deleted (or whether that an intentional constraint).

**Desired behavior:**
Currently unclear what the best UX would be – when deleting an agency, should all associated users be automatically deleted as well (and the request succeeds), or should we inform the admin user performing the deletion that agencies with users cannot be deleted (and that they need to manually delete those users first)?

**Implementation details**
Depending on desired behavior:
1. Users should be automatically deleted when their associated agency is deleted:
    -  Update the handler's execution path to check for and batch-delete users associated with the agency being deleted prior to issuing the `DELETE from "agencies" ...` query. 
    - Alternatively (and more atomically), adjust the `DELETE from "agencies" ...` query to appropriately cascade to records associated with the agency (including users and their associated records).
2. The UI should convey an error message that an agency cannot be deleted because it has assigned users.
    - Update the handler's execution path to check for associated users before attempting to delete the agency. If associated users are found, issue a 400 response that conveys the nature of the problem.
    - Update the UI to react appropriately when (400 status) responses containing errors are received by the client upon form submission.

TylerHendrickson:
cc @caitlinwinner for input on desired behavior.

mhuang-usdr:
[...]

PR #1432, opened 2023-05-26
Fix: #1395 delete agencies (short term solution)

### LINKED ISSUE #1431: BUG: Canceling "Are you sure..." confirmation dialogues doesn't cancel the operation
**Description:** On the "Edit Agency" modal, clicking the "Admin Delete Agency" button at the bottom of the form triggers a confirmation dialogue that warns the user that the action cannot be undone, and also provides instructions for deleting sub-agencies before deleting a parent. The confirmation dialogue offers "Cancel" and "OK" options, but both buttons currently have the same effect (as well as pressing the `[Esc]` key) of issuing a `DELETE /api/organizations/:organizationId/agencies/del/:agency` request to the API.

**Steps to reproduce:**
1. Login as a user with permission to delete agencies (e.g. tenant admin).
2. Create a new agency for the tenant.
3. Click the pencil icon next to the agency to bring up the "Edit Agency" modal.
4. Ensure that the browser's developer console is recording requests, then click the "Admin Delete Agency" button at the bottom of the form.
5. When the "Are you sure you want to delete this agency?" confirmation dialogue appears, click "Cancel" or press the `[Esc]` key.
6. Observe that the `DELETE /api/organizations/:organizationId/agencies/del/:agency` request fires, and that the agency has been deleted (no longer appears in the list of agencies).

**Impacts:**
- User experience: The behavior of the confirmation dialogue is misleading. Presumably users only click the "Admin Delete Agency" button with the intent of deleting an agency, but the ultimate selection the user makes through the additional confirmation dialogue should still be honored. Currently, the dialogue only exists to inform the user that _if_ the delete operation fails, they may need to ensure it has no child agencies before trying again, rather than giving the user a final opportunity to back out of the operation.

**Desired behavior:**
The dialogue warning should not provide a "cancel" option if the operation can't actually be canceled. Depending on whether we intend to provide this option in the first place, the desired behavior is one of the following:
1. If users should to be able to cancel the operation, the delete request should sent on dialogue close if the user clicked "OK".
2. If users shouldn't be able to cancel the operation, the dialogue should not present a "Cancel" button to the user.

**Implementation details:**
Per the two options presented in the "Desired behavior" listed above...
- **Option 1:**
  - Update the flow control in the `handleDelete()` method of the Vue component to inspect the result of the `msgBoxConfirm()` method call in order to determine whether to issue a request to delete the agency.
- **Option 2:**
  - The modal should be launched via by calling the `$bvModal.msgBoxOk()` method from the Vue component, rather than `$bvModal.msgBoxConfirm()`. Using `msgBoxOk` will only provide an "OK" button in the dialogue, rather than "OK" and "Cancel".
  - Additionally, the "Are you sure you want to delete this agency?" part of the dialogues message should be removed.

TylerHendrickson:
### Ticket #1395 #1431

Fixes #1431 

## Description

This PR implements the [short-term fix](https://github.com/usdigitalresponse/usdr-gost/issues/1395#issuecomment-1561730199) recommended for #1395. As described in that issue, users cannot delete agencies while they have associated users or sub-agencies. While these restrictions are expected behavior, the API currently does not check for these conditions or handle errors thrown when attempting to delete an agency with associated users and/or sub-agencies, nor does it inform users that the delete operation failed. This PR addresses those issues in the following ways:
- In the API: Before attempting to delete an agency from the database, the endpoint handler checks whether there are currently associated users or sub-agencies. If these conditions are met, the API responds with a 400 (Bad Request), along with a message stating the reason why the agency is not eligible for deletion (either `agency has assigned users` or `agency has sub-agencies`).
- In the UI: When a request to delete an agency is not successful, a [Vue modal message box](https://bootstrap-vue.org/docs/components/modal#modal-message-boxes) is displayed, which contains a message informing the user of the error and a single OK button.

Additionally, while working on this fix, I found another issue (#1431). In an exciting twist, it seems that selecting the "Cancel" option provided by the "Are you sure you want to delete this agency?" confirmation dialogue was not being honored, meaning that regardless of the user's selection, a request would always be issued to delete the agency. This PR follows the "Option 1" approach from that issue and continues to provide "OK" and "Cancel" options to the user, ensuring that a `DELETE` request is only sent when the user selects "OK".

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/1851017/02c737e1-6fec-49b5-8708-1b66e962cd3b

## Testing

1. Create a new agency
2. Create another agency, with the agency created in Step 1 as the parent
3. Create a user associated with the agency created in Step 1
4. Open the edit modal for the agency created in Step 1, and click the "Admin Delete Agency" button. 
5. Click "Cancel", and confirm the agency was not deleted.
6. Click "Admin Delete Agency" again, then click "OK". You should get an error because the agency has associated users.
7. Delete the user created in Step 3.
8. Do the same actions as in Step 6, but this time, the error should inform that there are associated sub-agencies.
9. Delete the sub-agency created in Step 2.
10. Do the same actions as in Steps 6 and 8. This time, the modal should close with no error.
11. Confirm that the agency created in Step 1 has been deleted.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/components/Modals/EditAgency.vue
+++ b/packages/client/src/components/Modals/EditAgency.vue
@@ -197,7 +197,7 @@ export default {
       const msgBoxConfirmResult = await this.$bvModal.msgBoxConfirm(
         'Are you sure you want to delete this agency? This cannot be undone. '
       + 'If the agency has children, reassign child agencies to continue deletion.',
-        { okTitle: 'Delete', okVariant: 'danger', title: 'Delete Agency' }
+        { okTitle: 'Delete', okVariant: 'danger', title: 'Delete Agency' },
       );
       if (msgBoxConfirmResult === true) {
         await this.deleteAgency({
--- a/packages/client/src/components/Modals/EditAgency.vue
+++ b/packages/client/src/components/Modals/EditAgency.vue
@@ -194,23 +194,28 @@ export default {
       if (this.$v.formData.$invalid) {
         return;
       }
-      await this.$bvModal.msgBoxConfirm(
-        'Are you sure you want to delete this agency? This cannot be undone. If the agency has children,'
-      + ' reassign child agencies to continue deletion.',
-      ).then(() => {
-        this.deleteAgency({
+      const msgBoxConfirmResult = await this.$bvModal.msgBoxConfirm(
+        'Are you sure you want to delete this agency? This cannot be undone. '
+      + 'If the agency has children, reassign child agencies to continue deletion.',
+        { okTitle: 'Delete', okVariant: 'danger', title: 'Delete Agency' }
+      );
+      if (msgBoxConfirmResult === true) {
+        await this.deleteAgency({
           agencyId: this.agency.id,
           parent: this.agency.parent,
           name: this.agency.name,
           abbreviation: this.agency.abbreviation,
           warningThreshold: this.formData.warningThreshold,
           dangerThreshold: this.formData.dangerThreshold,
+        }).then(() => {
+          this.resetModal();
+        }).catch(async (e) => {
+          await this.$bvModal.msgBoxOk(`Could not delete agency: ${e.message}`, {
+            title: 'Error',
+            bodyTextVariant: 'danger',
+          });
         });
-        this.resetModal();
-      })
-        .catch((err) => {
-          console.log(` ${err}`);
-        });
+      }
     },
     async handleSubmit() {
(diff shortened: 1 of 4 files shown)


jakekreider (approved):
Changes LGTM and tested out locally with both paths for both users and agencies.

PR #1495, opened 2023-06-08
938 consume grants from sqs

TylerHendrickson:
## Description

This PR adds a new library and associated script for consuming SQS messages originating from events published by the `grants-ingest` service. 

It's worth stating up-front that due to forthcoming schema design definitions in https://github.com/usdigitalresponse/grants-ingest/issues/114 (and subsequent incorporation into the published events), the new `grants-ingest` library code will ~likely~ definitely need to be updated soon – specifically, the logic involved in mapping data from the `grants-ingest`-provided format to the format used in the GOST `grants` table. However, given that I think everything else in this branch is here to stay, along with the fact that we won't be running the script in Production immediately, I it would be preferable to get these changes merged in sooner rather than later and avoid maintaining a long-running branch.

Summary of changes:
- Adds a new library module at `packages/server/src/lib/grants-ingest.js` that exports two functions:
  - `receiveNextMessageBatch()`: Long-polls an SQS queue for messages containing data for recently-updated grant opportunities. These messages will be delivered to the queue from events emitted by the `grants-ingest` service to an event bus.
  - `processMessages()`: Turns a batch of SQS messages (returned by `receiveNextMessageBatch()`) into GOST `grant` objects that are inserted/updated into the `grants` table.
- Adds a new Node.js script at `packages/server/src/scripts/consumeGrantModifications.js` that orchestrates the usage of the library, continuously long-polling for and processing any received SQS messages until a shut down is requested (which it attempts to perform in a graceful manner). This script serves as the entrypoint for the work in #1296, which will start ECS tasks that run the script according to the number of messages in the source SQS queue.
- Defines a new helper function, `getSQSClient()`, in the `gost-aws` module that initializes SQS clients from the AWS SDK with conditional configuration for LocalStack. This follows the pattern we've established for SES and S3 client initialization.
- Creates a Knex migration that adds a new `award_floor` column to the `grants` table. A few things to mention here:
  - Although it's used to store dollar values, it's implemented as a text column to stay consistent with the existing `award_ceiling` column. I'm not sure it matters much given our current usage, but we may want to migrate to a more appropriate data type for these columns in the future.  
  - I was surprised to learn that we weren't already persisting this data to a dedicated column; apparently, we just retrieve it from the JSON data stored in the `raw_body` column/field in the following scenarios:
    -  [On the server-side](https://github.com/usdigitalresponse/usdr-gost/blob/b0c8cd447c4b0edbfe0eb6d87b1c8a529f0cb0f3/packages/server/src/routes/grants.js#L13-L33), when [generating CSV exports](https://github.com/usdigitalresponse/usdr-gost/blob/b0c8cd447c4b0edbfe0eb6d87b1c8a529f0cb0f3/packages/server/src/routes/grants.js#L125).
    - [On the client-side](https://github.com/usdigitalresponse/usdr-gost/blob/b0c8cd447c4b0edbfe0eb6d87b1c8a529f0cb0f3/packages/client/src/components/GrantsTable.vue#L282-L302), when [preparing the grants table view](https://github.com/usdigitalresponse/usdr-gost/blob/b0c8cd447c4b0edbfe0eb6d87b1c8a529f0cb0f3/packages/client/src/components/GrantsTable.vue#L186).
  - Once we've started populating the `grants` table from the `grants-ingest` service, we should revisit and replace the usages of `raw_body` (side note: this is the only use-case I can find where we actually read that column) with the new `award_floor` column.
- Misc:
  - Adds tests for the `grants-ingest` module.
  - Adds `@aws-sdk/client-sqs` as a new dependency to the server package.
  - Updated the `localstack/entrypoint/init-aws.sh` script to create an SQS queue named `grants-ingest-events` on startup.
  - Add a new `$GRANTS_INGEST_EVENTS_QUEUE_URL` env var to `packages/server/.env.example` that points to the queue on LocalStack for development environments.

## Screenshots / Demo Video

938-demo

## Testing

**Note:** When testing with docker compose, you can start the script by prefixing the command in Step 3 with `docker compose exec -w /app/packages/server app`. However, it's not possible to observe the graceful shutdown behavior by running the script in this manner since the signal is sent to `docker compose` rather than the script running in the container. Instead, you have to access a shell inside the container by running `docker compose exec app bash` and then running `yarn consume-grants` or `node src/scripts/consumeGrantModifications.js` from the `packages/server` directory. Additionally, to stop the script by issuing `kill -SIGTERM <pid>`, you also need to `docker compose exec app bash` in (yet another) terminal.

1. Ensure that the `grants-ingest-events` SQS queue exists in LocalStack and that the `$GRANTS_INGEST_EVENTS_QUEUE_URL` env var is set to its queue URL.
2. Create a file named something like `sqs-message.json` that contains JSON resembling what we expect to receive from the queue (or some invalid data if you want to test how errors behave). 
    - **Note:** I'll drop an example in a comment on this PR that you can use to test.
4. Start the script by running `yarn consume-grants` from the `server` package or `node /path/to/consumeGrantModifications.js` (I recommend doing this!). 
    - Once it enters the main loop, it should log a message like `Long-polling next SQS message batch from <queue url>`.
5. Send a message to the queue by running `awslocal sqs send-message --queue-url $GRANTS_INGEST_EVENTS_QUEUE_URL --message-body file:///path/to/sqs-message.json`
    - If desired, you can run this command multiple times, either providing the same message or different messages (e.g. invalid JSON or other grants).
7. Observe the logs from the script. If valid JSON is received, it should indicate that it successfully parsed its content and inserted/updated the `grants` table accordingly. Once a batch has been processed, it should go back to logging the same `Long-polling next SQS message batch...` message about every 20 seconds (the configured long-poll duration).
8. At this point, you can keep sending different messages to the queue, or continue to the next step.
9. Check that the database has been updated according to the data you provided to the queue.
10. Request that the script shut down in any of the following ways:
    - Issue a `SIGINT` by pressing `CTRL+C` in the shell where the script is running. 
      - Note that when you do this, your shell will likely go back to a normal prompt but the script will continue to run until the current loop iteration has finished. When it finally stops, you'll get a detached `Shutting down` log message printed to your console.
    - In a separate shell, find the PID for the script (e.g. `ps aux | grep consume`) and issue a `SIGTERM` by running `kill -SIGTERM <pid>`.
      - Note: If you started the script with `yarn` rather than directly with `node`, your shell will probably also detach from the process but continue to print script logs. It's also important to target the PID belonging to the `node` process, rather than the PID associated with the `yarn` command. Frankly, I found it much less cumbersome to run `node /path/to/consumeGrantModifications.js` for this reason, but you do you :)

### Automated and Unit Tests
- [X] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [X] Added PR reviewers


TylerHendrickson:
Here's an example payload to save (e.g. as `sqs-message.json`) for manual testing:
```json
{
    "grant_id": "345856",
    "AdditionalInformationOnEligibility": "Eligible&#160;applicants&#160;include&#160;public&#160;and&#160;non-profit&#160;private&#160;agencies&#160;and&#160;coordinated&#160;networks&#160;of such&#160;entities.&#160;For-profit&#160;organizations&#160;are&#160;not&#160;eligible. Private institutions of higher education must be non-profit entities. &#160; Applications from individuals (including sole proprietorships) and foreign entities are not eligible and will be disqualified from competitive review and funding under this  funding opportunity. Faith-based and community organizations that meet the eligibility requirements are eligible to receive awards under this funding opportunity. Faith-based organizations may apply for this award on the same basis as any other organization, as set forth at and, subject to the protections and requirements of 45 CFR Part 87 and 42 U.S.C. 2000bb&#160;et seq., ACF will not, in the selection of recipients, discriminate against an organization on the basis of the organization&apos;s religious character, affiliation, or exercise.",
    "AdditionalInformationText": "",
    "AdditionalInformationURL": "",
    "AgencyCode": "HHS-ACF-FYSB",
    "AgencyName": "Administrationz for Children & Families - ACYF/FYSB",
    "ArchiveDate": "08302023",
    "AwardCeiling": "350000",
    "AwardFloor": "250000",
    "CategoryExplanation": "",
    "CategoryOfFundingActivity": [
      "ISS"
    ],
    "CFDANumbers": [
      "93.623"
    ],
    "CloseDate": "07292023",
    "CloseDateExplanation": "Electronically submitted applications must be submitted no later than 11:59 pm Eastern Standard Time on the listed application due date.",
    "CostSharingOrMatchingRequirement": "Yes",
    "Description": "The Runaway and Homeless Youth Prevention Demonstration Program (RHY-PDP) program supports the design and delivery of community-based demonstration initiatives to prevent youth and young adults from experiencing homelessness. Through development and coordination of partnerships with youth and young adult services providers, community organizations and private and public agencies the RHY-PDP will: 1) Identify young people at risk of experiencing homelessness; 2) Design and develop a comprehensive community prevention plan to prevent youth homeless; 3) Implement the robust, holistic services plan to respond to the diverse needs of youth who may be at risk of homelessness and their families. ",
    "EligibleApplicants": [
      "00",
      "01",
      "02",
      "04",
      "05",
      "06",
      "07",
      "08",
      "11",
      "12",
      "13",
      "20",
      "25"
    ],
    "EstimatedTotalProgramFunding": "1750000",
    "ExpectedNumberOfAwards": "5",
    "FundingInstrumentType": [
      "CA"
    ],
    "GrantorContactEmail": "Christopher.Holloway@acf.hhs.gov",
    "GrantorContactEmailDescription": "Christopher.Holloway@acf.hhs.gov",
    "GrantorContactText": "Chris Holloway&lt;br/&gt;Christopher.Holloway@acf.hhs.gov",
    "LastUpdatedDate": "05302023",
    "OpportunityCategory": "D",
    "OpportunityCategoryExplanation": "",
    "OpportunityID": "345856",
    "OpportunityNumber": "HHS-2023-ACF-ACYF-CY-0070",
    "OpportunityTitle": "FY 2023 Runaway and Homeless Youth-Prevention Demonstration Program (RHY-PDP)",
    "PostDate": "05302023",
    "Version": "Synopsis 1"
  }
```

packages/server/.env.example
```diff
@@ -43,3 +43,6 @@ DATA_DIR="./data"
 
 # S3 bucket for ARPA Audit reports
 AUDIT_REPORT_BUCKET=arpa-audit-reports
+
+# SQS queue for grants-ingest events
+GRANTS_INGEST_EVENTS_QUEUE_URL="http://localhost:4566/000000000000/grant-ingest-events"
```
as1729, line 48: curious: why all the zeroes in the path? Is that just how localstack does it? `000000000000`

packages/server/src/scripts/consumeGrantModifications.js
```diff
@@ -0,0 +1,31 @@
...
+    const queueUrl = process.env.GRANTS_INGEST_EVENTS_QUEUE_URL;
+    const sqs = getSQSClient();
+    while (shutDownRequested === false) {
+        console.log(`Long-polling next SQS message batch from ${queueUrl}`);
+        // eslint-disable-next-line no-await-in-loop
+        const messages = await receiveNextMessageBatch(sqs, queueUrl);
+        if (messages.length > 0) {
+            // eslint-disable-next-line no-await-in-loop
+            await processMessages(knex, sqs, queueUrl, messages);
+        }
```
as1729, line 24: do we want to implement some type of waiting mechanism here? or is it fine to call `receiveNextMessageBatch` immediately after it returns an empty result?

as1729: I'm concerned about a situation where the queue is returning some type of error and we keep polling.

packages/server/src/lib/grants-ingest.js
```diff
@@ -0,0 +1,156 @@
...
+ * @param { import('@aws-sdk/client-sqs').SQSClient } sqs AWS SDK client used to issue commands to the SQS API.
+ * @param { string } queueUrl The URL identifying the queue to poll.
+ * @param { Array[import('@aws-sdk/client-sqs').Message] } messages Messages to process from SQS.
+ */
+async function processMessages(knex, sqs, queueUrl, messages) {
+    let grantParseErrorCount = 0;
+    let grantSaveSuccessCount = 0;
+    let grantSaveErrorCount = 0;
+
+    return Promise.all(messages.map(async (message) => {
```
as1729, line 111: Note: Not critical for this PR, but something to consider as we build on it.

I would consider using `asyncBatch` for this to ensure we do not get the same memory overload issues we have seen in the past. It looks like we already cap the messages length in line 39 so it should be OK here. Although having the limit enforced at the point where we are calling these functions might make it more explicit.

packages/server/src/lib/grants-ingest.js
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        try {
+            await deleteMessage(sqs, queueUrl, message.ReceiptHandle);
+        } catch (e) {
+            console.log(
+                `Error deleting SQS message with receipt handle ${message.ReceiptHandle} `,
+                `for grant ${grant.grant_id}`,
+            );
+            throw e;
```
as1729, line 139: When this is thrown, will other async processes be cancelled as well?

packages/server/.env.example
TylerHendrickson: Yep, this is a LocalStack convention.

as1729 (approved):

packages/server/src/scripts/consumeGrantModifications.js
TylerHendrickson: > do we want to implement some type of waiting mechanism here? or is it fine to call `receiveNextMessageBatch` immediately after it returns an empty result?

The SQS `ReceiveMessages` behavior underlying this call already implements long-polling, which will wait for up to 20 seconds before returning, so we always wait that long in scenarios where there aren't any messages available.

> I'm concerned about a situation where the queue is returning some type of error and we keep polling.

If we encounter an error from SQS when performing the `ReceiveMessages` action, the error is thrown and will cause the running process to crash, which is intentional for this very reason – in the event that there's a problem with SQS (unlikely, but possible) or how we're calling SQS (more likely; could be due to missing permissions or programming error, etc.), we want the ECS task to register an error that the container failed so that it can back off accordingly, rather than just continuing to fail forever.

A few additional pieces of context that may be worth noting:
1. In the case of an intermittent failure, where SQS is unreachable for a brief period and then comes back, a new ECS task execution will simply replace the one that crashed. Therefore, it's better for this script to "fail hard" and allow the system to react accordingly, either by self-healing automatically (by replacing the task that failed) or backing off entirely if all tasks are failing (which indicates that developer intervention/investigation is necessary).
2. The auto-scaling behavior that will be implemented by #1296 will add/remove ECS tasks running this script according to the number of messages in the SQS queue. Theoretically, the scenario in which this script starts to receive nothing but empty message batches would indicate that the queue size is getting low, which means that ECS will start shutting down (some of) the running ECS tasks as part of the configured "scale-down" behavior.

packages/server/src/lib/grants-ingest.js
TylerHendrickson: I did consider using `asyncBatch` for this, but ultimately decided it didn't seem necessary given that the `ReceiveMessage` behavior already effectively limits the pool to a concurrency level of <= 10. 

If the main loop was engaging in more of a "fire and forget" pattern where it polled for up to 10 messages, sent them to a separate pool of async workers, and then immediately started polling again (without waiting for the previous 10 messages to process), we would definitely want to make use of `asyncBatch` to enforce an upper bound on the number of workers.

packages/server/src/lib/grants-ingest.js
TylerHendrickson: Yes. I think that's okay in this case for two reasons:
1. If one of the messages fails to delete, it's pretty likely that all of the others will also (due to a bug in the code, missing permission, a connectivity error, or some other issue where the impact isn't limited to one single message).
2. In the worst-case scenario, where:
    - 10 messages were received.
    - 9 of the messages were just saved as grants to the database but haven't been deleted.
    - 1 of the loops was ahead of the others and encountered a completely one-off error when deleting its message.
    
    Then all 10 messages are returned to the queue and will eventually be consumed again. However, their data will have already been saved to the db, so this time around, nothing will actually be modified, and then they'll (presumably) be successfully deleted. A bit of unnecessary work would have occurred, but I think the per-batch processing costs are fairly negligible. 

That said, we'll want to make sure we can monitor these (and other) failure scenarios. It just occurred to me that the Datadog instrumentation recently added to the Express API parts of this repo won't actually integrate with this script, so I'll go ahead and add that now.

PR #1563, opened 2023-06-22
Search chips

### LINKED ISSUE #1394: Create filter chip component
# Description
Create the filter selection chip component. 

# Acceptance Criteria

* This component should have a label, the filter terms, and an x to remove the selection. 
* It should be able to take various primary section name and a list of terms. 
Screen Shot 2023-05-18 at 11 17 01 PM

# Parent ticket 
* https://github.com/usdigitalresponse/usdr-gost/issues/1368

jakekreider:
## Description
- Adds a new SearchFilter component which will display the chips for a current filter
- Until the state is handled in the parent, the component will remove by clicking close X or "clear all". This should probably be an emit later down the road.
- Per discussion at standup, I also split out the new search functionality into a copy of the `GrantsTable` component. The new one will display when `NODE_ENV='development'` (which should already be the case for us locally). 

Closes #1394 

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/624510/b1e5cf43-9c70-46da-95c0-30ed280ef098

## Testing
- With `NODE_ENV='development'`, confirm the filter chips display on the My Grants and Grants pages.
- Add/remove data from the `searchFilters`  variable in `GrantsTableNext.vue`. Observe display on the page.

### Automated and Unit Tests
- [ ] Added Unit tests

**I haven't added any unit tests**. I didn't find where/how these component tests are being done; would appreciate if someone could point me to that, or if I should start from scratch.

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/terraform/modules/gost_website/runtime_config.tf
+++ b/terraform/modules/gost_website/runtime_config.tf
@@ -2,7 +2,8 @@ locals {
   deployConfigContents = templatefile(
     "${path.module}/tpl/deploy-config.js",
     {
-      gost_api_domain = var.gost_api_domain,
+      gost_api_domain    = var.gost_api_domain,
+      gost_use_new_table = var.gost_use_new_table,
     }
   )
 }
--- a/terraform/modules/gost_website/tpl/deploy-config.js
+++ b/terraform/modules/gost_website/tpl/deploy-config.js
@@ -1,3 +1,3 @@
 window.APP_CONFIG = window.APP_CONFIG || {};
 window.apiURLForGOST = 'https://${gost_api_domain}/';
-window.UseNewTable = ${use_new_table};
+window.APP_CONFIG.UseNewTable = ${gost_use_new_table};
--- a/terraform/modules/gost_website/tpl/deploy-config.js
+++ b/terraform/modules/gost_website/tpl/deploy-config.js
@@ -1,2 +1,3 @@
 window.APP_CONFIG = window.APP_CONFIG || {};
 window.apiURLForGOST = 'https://${gost_api_domain}/';
+window.UseNewTable = ${use_new_table};
--- a/packages/client/.env.example
+++ b/packages/client/.env.example
@@ -1,2 +1,2 @@
-# there aren't currently any env variables needed on the client
-# so this file is empty!
+# use the in-progress grant table experience
+VUE_APP_USE_NEW_TABLE=true
--- a/packages/client/public/deploy-config.js.example
+++ b/packages/client/public/deploy-config.js.example
@@ -1,2 +1,3 @@
 window.APP_CONFIG = window.APP_CONFIG || {};
 window.APP_CONFIG.apiURLForGOST = "https://example.com/api";
+window.APP_CONFIG.UseNewTable = true;
--- a/packages/client/src/components/EmailSettingsBanner.vue
+++ b/packages/client/src/components/EmailSettingsBanner.vue
@@ -4,7 +4,7 @@
           <b-icon icon="envelope-fill" font-scale="1.5" style="width: 50px; color: #0FA958;"></b-icon>
           <div style="padding-right: 10px; color: #212529;">
           <b>The Grant ID Tool now has email notifications!</b>
-          Fine tune which email you will recieve in <u v-on:click="showProfileSettings()">Settings</u> or turn on all notifications here:
+          Fine tune which email you will receive in <u v-on:click="showProfileSettings()">Settings</u> or turn on all notifications here:
         </div >
         <b-form-checkbox v-model="checked" name="email-opt-in-switch" switch @change="onUserSubscriptionChangeSubmit"/>
       </b-alert>
--- a/packages/client/src/components/GrantsTable.vue
+++ b/packages/client/src/components/GrantsTable.vue
@@ -99,10 +99,28 @@ export default {
       perPage: 50,
       currentPage: 1,
       loading: false,
-      searchFilters: {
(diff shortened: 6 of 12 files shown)


TylerHendrickson (changes requested):
> Per discussion at standup, I also split out the new search functionality into a copy of the GrantsTable component. The new one will display when NODE_ENV='development' (which should already be the case for us locally).

This will work well for local development where the Vue dev server is running and can inject environment variables dynamically. However, relying on environment variables isn't as feasible when it comes to environments like Staging or Prod, where env vars are set at build-time. I think we'll want to be able to be able to turn this on for Staging builds in addition to local dev environments; I can see two ways of going about making that possible:
1. Get acquainted with the [way `vue-cli-service` works with environment variables](https://cli.vuejs.org/guide/mode-and-env.html) and incorporates them when building static bundles, and then update accordingly things accordingly; this would entail ensuring that the [Build GOST Website](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/.github/workflows/build-website.yml) and/or the [Build and Deploy GOST](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/.github/workflows/build-and-deploy.yml) workflows can provide the correct values on a per-target-environment basis. Note that I'm not positive that `NODE_ENV='development'` will be useable in this case – we don't set `NODE_ENV` explicitly in the CI/CD workflows, and I'm not sure the effective value is `'development'` (or at least it doesn't seem like it should be) – so that would need to be looked into as well.
2. Update the conditional `VUE_APP_USE_NEW_TABLE` check to also consider a fallback value that we can toggle on a per-environment basis, e.g. 
      ```javascript
      // NOTE: Not sure we need to check NODE_ENV anymore
      const useNewTable = process.env.VUE_APP_USE_NEW_TABLE === 'true' || 
          (window.APP_CONFIG || {}).UseNewTable === true;
      ```
      Then, configure `window.APP_CONFIG.UseNewTable` [through Terraform](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/terraform/modules/gost_website/runtime_config.tf) by passing in an input variable to the `website` module that we set in `<env>.tfvars`. 

My personal recommendation would be to go for the second option, but that's probably colored by the fact that I'm much more familiar with those mechanics than the build conventions relating to `VUE_APP_` environment variables.

jakekreider:
> This will work well for local development where the Vue dev server is running and can inject environment variables dynamically. However, relying on environment variables isn't as feasible when it comes to environments like Staging or Prod, where env vars are set at build-time.

My thought here was that it would be evaluated at build time, but didn't realize the different NODE_ENV's might not be what I need them to be. `VUE_APP_*` vars would have the same limitation, so I went with the approach with deploy-config.js. Let me know if you had something else in mind!

terraform/modules/gost_website/runtime_config.tf
```diff
@@ -2,7 +2,8 @@ locals {
   deployConfigContents = templatefile(
     "${path.module}/tpl/deploy-config.js",
     {
-      gost_api_domain = var.gost_api_domain,
+      gost_api_domain    = var.gost_api_domain,
+      gost_use_new_table = var.gost_use_new_table,
```
jakekreider, line 6: Note - I have _not_ tested TF locally. Will CI runs catch any mistakes or should I get my env setup for that?

terraform/modules/gost_website/tpl/deploy-config.js
```diff
@@ -1,2 +1,3 @@
 window.APP_CONFIG = window.APP_CONFIG || {};
 window.apiURLForGOST = 'https://${gost_api_domain}/';
+window.UseNewTable = ${use_new_table};
```
TylerHendrickson, line 3: This needs to be added to `window.APP_CONFIG` in order to be consistent with the way it's being referenced in the `.vue` files. Note that you _could_ keep this line unmodified and update those references to `window.UseNewTable`, but my personal preference would be to avoid cluttering the top-level `window` namespace and use `APP_CONFIG` for these types of things.

```suggestion
window.APP_CONFIG.UseNewTable = ${use_new_table};
```

terraform/modules/gost_website/runtime_config.tf
TylerHendrickson: CI runs will catch failures to run `terraform plan`, inspecting the plan output is always a good idea. In this specific case, I don't think that would catch a mistake involving a bad `var.gost_use_new_table` value or something like that, but I think it would be quite complicated to test this via ECS-in-LocalStack. Personally, I think it would be okay to wait for this to get deployed to Staging and just do manual testing there.

TylerHendrickson (changes requested):

terraform/modules/gost_website/runtime_config.tf
TylerHendrickson: Well what do you know! It appears that CI _did_ [catch something useful](https://github.com/usdigitalresponse/usdr-gost/actions/runs/5360920348/jobs/9726372315?pr=1563#step:10:31) after all.

Looks like the template is expecting a variable named `use_new_table`, not `gost_use_new_table`.
```suggestion
      gost_api_domain = var.gost_api_domain,
      use_new_table   = var.gost_use_new_table,
```

jakekreider: I was pretty confused by this until I realized I renamed the variable locally everywhere but forgot to stage the js template 🤦‍♂️

TylerHendrickson (approved):

Issue #1268, opened 2023-04-20
Ability to Archive/Exclude an upload from the Treasury export and Audit Report


jcomeau1:
There are a number of cases where the partners will upload an Input Template, which is valid, but then they determine that it should NOT go to Treasury. Currently there is no mechanism to Archive or Exclude a specific uploaded (validated) file from going to Treasury. This extends to the Audit Report as well.

A couple of known reasons why this could happen:

1. The Admins wanted to test something either with the Reporter itself or with the Treasury Portal
2. An Agency uploaded the same Input Template twice, but used two different Agencies

The only option partners currently have is to manually edit the output files generated and we should guard against this. Additionally, once Input is validated, it will also be included in the Audit Reports

jcomeau1:
A potential concept for solving this issue would be to give the users (admins only?) the ability to invalidate a specific upload that previously validated:


![Image](https://user-images.githubusercontent.com/78866741/236268527-5a1d407a-5302-40af-ae00-d54f6b918165.png)

mhuang-usdr:
old ticket tracking this is #705

as1729:
Goal
- It should not be in the audit report
- It should not be included in the output template

What happens to the file?

Easier from technical aspect
- Delete? Make sure the partners are 100% aware

A bit more complex from design and eng aspect
- Archive? Lets understand the risks of keeping this file. How do we display it? How long do we hold onto the file.

Who can perform this action?
- All admins?
- USDR admins?
- The file uploader and admins?

as1729:
- Lean more towards the option of "archive/hide" rather than permanently deleting the workbook. (+2 votes)
- Technically the agencies should have this saved, but having the ability to look back is always helpful.
- Highest priority is to make sure it does not end up on the treasury report.
- It will be helpful to still see it in the uploads tab with the ability to have a toggle to show/hide.
- It should also not show up on the audit report as well.

as1729:
I believe @cajinajr is working on this. Can you confirm @caitlinwinner  so @jcomeau1 is aware who to reach out to?

cajinajr:
Confirmed. I am working on this ticket.

cajinajr:
@as1729 @mhuang-usdr - Work for this ticket ready for ENG review
[...]

PR #1570, opened 2023-06-23
fix: #1268 

tzinckgraf:
## Description
This is only the backend piece. This PR adds an invalidated_at and invalidated_by to the uploads table. These can then be excluded by various reporting, such as the audit report.

## Testing
This is just the backend portion, but can be tested by manually setting the `invalidated_at` and `invalidated_by` column and running the audit report.

### Automated and Unit Tests
- [x] Added Unit tests - Partially done

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/arpa_reporter/server/db/uploads.spec.js
+++ b/packages/server/__tests__/arpa_reporter/server/db/uploads.spec.js
@@ -1,5 +1,6 @@
 const uploads = requireSrc(__filename);
 const assert = require('assert');
+const fixtures = require('../fixtures/fixtures');
 const { withTenantId } = require('../helpers/with-tenant-id');
 const { knex } = require('../mocha_init');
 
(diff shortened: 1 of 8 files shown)


packages/server/src/arpa_reporter/routes/uploads.js
```diff
@@ -155,6 +155,32 @@ router.post('/:id/validate', requireUser, async (req, res) => {
     }
 });
 
+router.post('/:id/invalidate', requireUser, async (req, res) => {
```
as1729, line 158: we should update this to `requireAdminUser` but since this is not visible to any users at the moment and we have audit-log enabled, lets update this in a follow-up PR

packages/server/__tests__/arpa_reporter/server/db/uploads.spec.js
```diff
@@ -42,6 +43,78 @@ describe('db/uploads.js', () => {
...
         });
     });
+
+    describe('usedForTreasuryExport', () => {
+        before(async () => {
+            await fixtures.seed(knex);
+        });
+
+        after(async () => {
+            await fixtures.clean(knex);
```
as1729, line 53: small improvement that can be handled in a follow-up PR.

I think adding the following code will eliminate the need for lines 70-77, 95-102, 17,19).

```js
describe('db/uploads.js', () => {
    beforeEach(async () => {
        await fixtures.clean(knex)
        await fixtures.seed(knex)
    });
    afterEach(async () => {
        await fixtures.clean(knex)
    });
}
```

as1729 (approved):
Looks good, couple of changes that we should make before the feature is live for end-users.

packages/server/src/arpa_reporter/db/uploads.js
```diff
@@ -145,6 +149,19 @@ async function markNotValidated(uploadId, trns = knex) {
...
+            invalidated_by: null,
+        })
+        .returning('*')
+        .then((rows) => rows[0]);
+}
+
+async function markInvalidated(uploadId, userId, trns = knex) {
+    return trns('uploads')
+        .where('id', uploadId)
+        .update({
```
as1729, line 162: we should also null the validated fields here-
```js
.update({
    validated_at: null,
    validated_by: null,
    ...
```

tzinckgraf: I was debating that, but would it be valuable to know if it was validated then manually invalidated?

PR #1571, opened 2023-06-23
Treasury Exports - Implements async report generation

as1729:
## Description
This PR does the following:
1. Makes treasury report generation async and sends the user a signed S3 URL via email. Note: this is behind a query-parameter feature flag for the time being.
2. Ensures that treasury report generation does not depend on having an express js request/response context. It can be called independently by passing in the needed `tenantId` or `periodId` via CloudShell or other services.
3. Ensures that we are not passing in an unlimited number promises to be executed simultaneously. Currently starting with a limit of 2.

## Screenshots / Demo Video
https://www.loom.com/share/ff1d5449c53a4545b1295e06229a5bbc

### Before
image

### After
image

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

- Go to http://localhost:8080/arpa_reports?async_treasury_report=true
- Hit the "Send Treasury Report By Email" button
- Navigate to your local SES temp storage and identify the latest sent message json
- Navigate to the URL in the email that begins with http://localhost:3000/api/exports....
- Confirm that you get redirected to a signed localstack URL and the file downloads.

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/arpa_reporter/views/Home.vue
+++ b/packages/client/src/arpa_reporter/views/Home.vue
@@ -8,6 +8,13 @@
         <DownloadButton :href="downloadTreasuryReportURL()" class="btn btn-primary btn-block">Download Treasury Report</DownloadButton>
       </div>
 
+      <div class="col" v-if="this.$route.query.async_treasury_download && isAdmin">
+        <button class="btn btn-primary btn-block" @click="sendTreasuryReport" :disabled="sending">
+          <span v-if="sending">Sending...</span>
+          <span v-else>Send Treasury Report by Email</span>
+        </button>
+      </div>
+
       <div class="col" v-if="this.$route.query.sync_audit_download && isAdmin">
         <DownloadButton :href="downloadAuditReportURL()" class="btn btn-info btn-block">Download Audit Report</DownloadButton>
       </div>
@@ -128,6 +135,34 @@ export default {
     downloadTreasuryReportURL() {
       const periodId = this.$store.getters.viewPeriod.id || 0;
       return apiURL(`/api/exports?period_id=${periodId}`);
+    },
+    async sendTreasuryReport() {
+      this.sending = true;
+
+      try {
+        const result = await getJson('/api/exports?async=true');
+
+        if (result.error) {
+          this.alert = {
+            text: 'Something went wrong. Unable to send an email containing the treasury report. Reach out to grants-helpdesk@usdigitalresponse.org if this happens again.',
+            level: 'err',
+          };
+          console.log(result.error);
+        } else {
+          this.alert = {
+            text: 'Sent. Please note, it could take up to 1 hour for this email to arrive.',
+            level: 'ok',
+          };
+        }
+      } catch (error) {
+        // we got an error from the backend, but the backend didn't send reasons
+        this.alert = {
+          text: 'Something went wrong. Unable to send an email containing the treasury report. Reach out to grants-helpdesk@usdigitalresponse.org if this happens again.',
+          level: 'err',
+        };
+      }
+
+      this.sending = false;
     },
     startUpload() {
       if (this.viewingOpenPeriod) {
--- a/packages/server/src/arpa_reporter/routes/exports.js
+++ b/packages/server/src/arpa_reporter/routes/exports.js
@@ -1,12 +1,50 @@
 const express = require('express');
 
 const router = express.Router();
+const { HeadObjectCommand, GetObjectCommand } = require('@aws-sdk/client-s3');
 const _ = require('lodash');
 
-const { requireUser } = require('../../lib/access-helpers');
+const aws = require('../../lib/gost-aws');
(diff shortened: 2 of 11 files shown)


packages/client/src/arpa_reporter/views/Home.vue
```diff
@@ -129,6 +136,34 @@ export default {
...
+
+        if (result.error) {
+          this.alert = {
+            text: 'Something went wrong. Unable to send an email containing the treasury report. Reach out to grants-helpdesk@usdigitalresponse.org if this happens again.',
+            level: 'err',
+          };
+          console.log(result.error);
+        } else {
+          this.alert = {
+            text: 'Sent. Please note, it could take up to 1 hour for this email to arrive.',
```
TylerHendrickson, line 153: Small nit, and a bit outside my wheelhouse, but I'd suggest tweaking this wording to focus on "we received/are working on your request" rather than "sent", since the latter is a bit ambiguous wrt to the email aspect.

packages/server/src/arpa_reporter/services/generate-arpa-report.js
```diff
@@ -904,9 +909,44 @@ async function generateSubRecipient(records, periodId) {
...
+async function setCSVData(data) {
+    const {
+        csvObject, admZip, records, periodId,
+    } = data;
+    const { name, func } = csvObject;
+    const csvData = await func(records, periodId);
+
+    if (!Array.isArray(csvData)) {
+        console.dir({ name, func });
+        console.dir(csvData);
```
TylerHendrickson, line 921: Since the nature of this scenario is that the structure/contents of `csvData` (and perhaps implicitly, `csvObject`) is unexpected, I'm wary of using this as a debugging step, especially in absence of other contextual information about what's being dumped to the logs. 

At the very least, it would be useful to prefix this output so that it can be easily located in a debugging scenario, but I wonder if there's a way to limit what's getting dumped here without losing relevant debugging context. Just food for thought.

packages/server/src/arpa_reporter/routes/exports.js
```diff
@@ -1,11 +1,50 @@
...
+
+    const s3 = aws.getS3Client();
+    const Key = `${user.tenant_id}/${req.params.periodId}/${req.params.filename}`;
+    const baseParams = { Bucket: process.env.AUDIT_REPORT_BUCKET, Key };
+
+    try {
+        await s3.send(new HeadObjectCommand(baseParams));
+    } catch (error) {
+        console.log(error);
+        res.redirect(encodeURI(`${process.env.WEBSITE_DOMAIN}/arpa_reporter?alert_text=The treasury report you requested has expired. Please try again by clicking the 'Send Treasury Report By Email'.&alert_level=err`));
```
tzinckgraf, line 34: Will this work? I think this will redirect, but not add the `async=True` param, which would mean the button won't show up.

packages/server/src/arpa_reporter/routes/exports.js
```diff
@@ -14,8 +53,25 @@ router.get('/', requireUser, async (req, res) => {
...
     }
+    const tenantId = useTenantId();
+
+    if (req.query.async) {
+        // Special handling for async treasury report generation and sending.
+        console.log('/api/exports?async=true GET');
+        console.log('Generating Async treasury report');
+        try {
+            const user = useUser();
+            arpa.generateAndSendEmail(user.email, periodId, tenantId);
```
tzinckgraf, line 64: We should `await` here, otherwise it could fail after the success.

as1729: Here (and similarly for the audit-report) I made a deliberate trade-off call. Lmk if we can handle this more gracefully.

The constraints are:
- The aws api-gateway request timeout is 30 seconds
- Report generation for partners with larger files can be in order of minutes

I felt it made sense to tell the client asap that their request was successful here and find an alternate means to notify the client of a failure if it occurred a few minutes into the file processing. Currently there is no "failure email" setup but that seemed like a good next-step.

wdyt?

packages/server/src/arpa_reporter/services/generate-arpa-report.js
as1729: I agree. I removed it for now. 👍

packages/server/src/arpa_reporter/routes/exports.js
tzinckgraf: That makes sense. I don't think there is an easier way to handle this more gracefully. We can discuss on Friday, but it would require more work and thought around the proper design. This looks good for now. Thanks for the explanation.

tzinckgraf (approved):

packages/client/src/arpa_reporter/views/Home.vue
as1729: @TylerHendrickson I'm going to keep this as-is for now given that it is only visible for one-partner. Will incorporate the final copy when I flip the feature-flag.

PR #1668, opened 2023-07-20
fix(server): default to zero for null values in columns for new reports

michello:
## Description
When constructing the new summaries report for records grouped by project, we originally added the values from the record without checking whether it's null or not (as this check is not done in the [original audit report generation](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/server/src/arpa_reporter/lib/audit-report.js#L86)). As a result, this causes the summation to result in NaN, which throws an error when opening the report. [See slack thread for context.](https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1689792789275749)

This PR resolves any null values to 0 if the value we're adding doesn't exist.

## Screenshots / Demo Video
[See slack thread for demo of the issue.](https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1689792789275749)

### Manual tests for Reviewer
* Before the fix - Downloaded the audit report and opened the workbook: a fixed banner indicates there's an issue with opening/parsing the workbook:
image
* After the fix - Downloaded the audit report and opened the workbook: does not result in the workbook fixed banner, and the NaN cells do not exist:
image

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -131,10 +131,10 @@ async function getProjectSummaryGroupedByProjectRow(data) {
     // set values in each column
     records.forEach(async (r) => {
         const reportingPeriodEndDate = reportingPeriods.filter((reportingPeriod) => r.upload.reporting_period_id === reportingPeriod.id)[0].end_date;
-        row[`${reportingPeriodEndDate} Total Aggregate Expenditures`] += r.content.Total_Expenditures__c;
-        row[`${reportingPeriodEndDate} Total Aggregate Obligations`] += r.content.Total_Obligations__c;
-        row[`${reportingPeriodEndDate} Total Obligations for Awards Greater or Equal to $50k`] += record.content.Award_Amount__c;
-        row[`${reportingPeriodEndDate} Total Expenditures for Awards Greater or Equal to $50k`] += record.content.Expenditure_Amount__c;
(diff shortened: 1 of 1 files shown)


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -131,10 +131,10 @@ async function getProjectSummaryGroupedByProjectRow(data) {
     // set values in each column
     records.forEach(async (r) => {
         const reportingPeriodEndDate = reportingPeriods.filter((reportingPeriod) => r.upload.reporting_period_id === reportingPeriod.id)[0].end_date;
-        row[`${reportingPeriodEndDate} Total Aggregate Expenditures`] += r.content.Total_Expenditures__c;
-        row[`${reportingPeriodEndDate} Total Aggregate Obligations`] += r.content.Total_Obligations__c;
-        row[`${reportingPeriodEndDate} Total Obligations for Awards Greater or Equal to $50k`] += record.content.Award_Amount__c;
-        row[`${reportingPeriodEndDate} Total Expenditures for Awards Greater or Equal to $50k`] += record.content.Expenditure_Amount__c;
+        row[`${reportingPeriodEndDate} Total Aggregate Expenditures`] += (r.content.Total_Expenditures__c || 0);
```
tzinckgraf, line 134: This also works for `NaN`, so that's good.

tzinckgraf (approved):

Issue #1408, opened 2023-05-22
Remove unused tabs 


Samcbehrens:
# Acceptance criteria 
Remove the Eligibility and keyword tabs 

<img width="766" alt="Screen Shot 2023-05-21 at 10 46 55 PM" src="https://github.com/usdigitalresponse/usdr-gost/assets/5168762/cbc135d1-1e4b-453c-9e2e-6a1dca1cbcc4">

PR #1679, opened 2023-07-21
Removing Keywords and Eligibility Codes tabs routes and modals

replicantSocks:
## Screenshots / Demo Video
Screenshot from 2023-07-21 14-16-38

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (changes requested):
(CC @jakekreider in case there's any additional context/info worth adding here.)
This looks good according to my tests and overall satisfies the AC listed in #1408 👍 

However, I'm wondering if the deleted files and corresponding routes need to be re-added and gated with a conditional that checks the feature flag recently implemented by @jakekreider in #1563. Here's a link to an example of it in-use: https://github.com/usdigitalresponse/usdr-gost/blob/50b45afb4eb4a04a8afd3a85a1ff5cbd665efd16/packages/client/src/views/Grants.vue#L20.

Alternatively, a new feature flag could be put in-place that more directly corresponds to the availability of Eligibility Codes and Keywords tabs (although I don't think a new feature flag is strictly necessary, since we'd presumably want to activate the changes in this PR with the others that are gated by the existing `process.env. VUE_APP_USE_NEW_TABLE` / `window.APP_CONFIG` check.

As-is, however, I think these changes would break Production if they were made live ahead of the other changes related to the Search milestone, since users need the ability to interact with the Eligibility Codes and Keywords tabs until the new advanced search features can fully replace them.

replicantSocks:
> (CC @jakekreider in case there's any additional context/info worth adding here.) This looks good according to my tests and overall satisfies the AC listed in #1408 👍
> 
> However, I'm wondering if the deleted files and corresponding routes need to be re-added and gated with a conditional that checks the feature flag recently implemented by @jakekreider in #1563. Here's a link to an example of it in-use:
> 
> https://github.com/usdigitalresponse/usdr-gost/blob/50b45afb4eb4a04a8afd3a85a1ff5cbd665efd16/packages/client/src/views/Grants.vue#L20
> .
> 
> Alternatively, a new feature flag could be put in-place that more directly corresponds to the availability of Eligibility Codes and Keywords tabs (although I don't think a new feature flag is strictly necessary, since we'd presumably want to activate the changes in this PR with the others that are gated by the existing `process.env. VUE_APP_USE_NEW_TABLE` / `window.APP_CONFIG` check.
> 
> As-is, however, I think these changes would break Production if they were made live ahead of the other changes related to the Search milestone, since users need the ability to interact with the Eligibility Codes and Keywords tabs until the new advanced search features can fully replace them.

I definitely see the issue with rolling this out as is right now. I had missed that we're switching to the new experience based on the flag, and I can send out a new change to reflect that. I assume, then we will schedule some time to purge the old components following the milestone?

jakekreider:
@replicantSocks @TylerHendrickson One thing to note: I had a typo in the original flag setup that prevented it from working right on the My Grants page. See fix in this PR https://github.com/usdigitalresponse/usdr-gost/pull/1638/files#diff-afad9e2cb3482cf68ed46dc447ca14f6c0599a5d4dfdfb810b023129ed41e080

PR #1650, opened 2023-07-16
feat(1303): creating new project summary sheets

michello:
## Description
This PR groups the report records by project and explodes the following columns by the reporting period:
- Total Aggregate Expenditures
- Total Aggregate Obligations
- Total Expenditures for Awards Greater or Equal to $50k
- Total Obligations for Awards Greater or Equal to $50k

This means, in the new project summary sheets, we are grouping records by project and creating a new column for every reporting period associated for that project.

## Screenshots / Demo Video
image

## Testing
1. Local DB has the following two records:
```
const records = [
  {
    type: 'ec5',
    subcategory: "5.19-Broadband: 'Last Mile' projects",
    upload: {
      filename: '2023-06-22-Michelle_Testing.xlsm',
      created_at: '2023-06-23T02:43:11.625Z',
      reporting_period_id: 43,
      user_id: 17,
      agency_id: 0,
      validated_at: '2023-06-23T02:43:12.277Z',
      validated_by: 17,
      ec_code: '5.19',
      tenant_id: 1,
      id: 'adf5acdb-e0c2-4554-a85c-1efc24dda9b1',
      notes: null,
      invalidated_at: null,
      invalidated_by: null,
      created_by: 'grant-admin@usdigitalresponse.org',
      agency_code: 'USDR'
    },
    content: {
      Name: 'Project Broadband',
      Project_Identification_Number__c: 'A010203',
      Completion_Status__c: 'Completed less than 50%',
      Adopted_Budget__c: 2000000,
      Total_Obligations__c: 1000000,
      Total_Expenditures__c: 1000000,
      Current_Period_Obligations__c: 1000000,
      Current_Period_Expenditures__c: 1000000,
      Project_Description__c: 'This is a description.',
      Proj_Actual_Construction_Start_Date__c: '2023-03-01T05:00:00.000Z',
      Initiation_of_Operations_Date__c: '2023-07-01T04:00:00.000Z',
      Is_project_designed_to_meet_100_mbps__c: 'Yes',
      Is_project_designed_to_exceed_100_mbps__c: 'Yes',
      Is_project_designed_provide_hh_service__c: 'Yes',
      Technology_Type_Planned__c: 'Fiber',
      Total_Miles_of_Fiber_Deployed_c: 30,
      Planned_Funded_Locations_Served__c: 30,
      Planned_Funded_Locations_25_3_Below__c: 0,
      Planned_Funded_Locations_Between_25_100__c: 0,
[...]

PR #1713, opened 2023-07-30
feat(1303): generate kpi sheets in arpa audit report

michello:
## Description
Adding logic to generate a new KPI sheets in the ARPA Audit Report. This sheet is similar to existing `Project Summaries V2` sheet [see previous PR](https://github.com/usdigitalresponse/usdr-gost/pull/1650) where we are grouping records by project except that we are displaying different columns.

In order to avoid making additional calls to generate this sheets, we have expanded the existing logic to create `Project Summaries V2` to also generate the rows for `KPI` (since both sheets are essentially using the same dataset).

## Screenshots / Demo Video
image

## Testing
1. Uploaded file and checked out exported report (screenshot above)
2. Able to see the KPI tab (and existing `Project Summaries V2`). Both tabs/sheets loaded without issues.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -150,8 +150,8 @@ async function getReportDataGroupedByProjectRow(data) {
 
         // for kpi report
         const currentPeriodExpenditure = r.content.Current_Period_Expenditures__c || 0;
-        kpiRow['Number of Subawards'] += (r.type === 'awards50k' ? 1 : 0);
-        kpiRow['Number of Expenditures'] += (currentPeriodExpenditure > 0 ? 1 : 0);
+        kpiRow['Number of Subawards'] += (r.type === 'awards50k');
+        kpiRow['Number of Expenditures'] += (currentPeriodExpenditure > 0);
         kpiRow['Evidence based total spend'] += (r.content.Spending_Allocated_Toward_Evidence_Based_Interventions || 0);
     });
 
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -102,17 +102,24 @@ async function getProjectSummaryRow(data) {
     };
 }
 
-async function getProjectSummaryGroupedByProjectRow(data) {
+async function getReportDataGroupedByProjectRow(data) {
     const { projectId, records, reportingPeriods } = data;
 
     const record = records[0];
 
     // set values for columns that are common across all records of projectId
-    const row = {
+    const projectSummaryV2Row = {
         'Project ID': projectId,
         'Project Description': record.content.Project_Description__c,
         'Project Expenditure Category Group': ec(record.type),
         'Project Expenditure Category': record.subcategory,
+    };
+
+    const kpiRow = {
+        'Project ID': projectId,
+        'Number of Subawards': 0,
+        'Number of Expenditures': 0,
+        'Evidence Based Total Spend': 0,
     };
 
     // get all reporting periods related to the project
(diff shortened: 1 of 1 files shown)


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -125,22 +132,29 @@ async function getProjectSummaryGroupedByProjectRow(data) {
...
+        // for kpi report
+        const currentPeriodExpenditure = r.content.Current_Period_Expenditures__c || 0;
+        kpiRow['Number of Subawards'] += (r.type === 'awards50k');
+        kpiRow['Number of Expenditures'] += (currentPeriodExpenditure > 0);
+        kpiRow['Evidence based total spend'] += (r.content.Spending_Allocated_Toward_Evidence_Based_Interventions || 0);
     });
 
-    return row;
+    return [projectSummaryV2Row, kpiRow];
 }
```
ajhyndman, line 158: Nit: I think this functionality makes sense, but I would suggest considering splitting it into two functions?

I think your solution here is technically more efficient, since you only have to loop over the data once.  However, it's a little confusing to read; it's one function generating data for two sheets, you have to "unzip" the data afterward, etc.  And I am very doubtful that this loop is actually a bottleneck for performance.

packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -125,22 +132,29 @@ async function getProjectSummaryGroupedByProjectRow(data) {
...
-        row['Capital Expenditure Amount'] += (r.content.Total_Cost_Capital_Expenditure__c || 0);
+        projectSummaryV2Row[`${reportingPeriodEndDate} Total Aggregate Expenditures`] += (r.content.Total_Expenditures__c || 0);
+        projectSummaryV2Row[`${reportingPeriodEndDate} Total Aggregate Obligations`] += (r.content.Total_Obligations__c || 0);
+        projectSummaryV2Row[`${reportingPeriodEndDate} Total Obligations for Awards Greater or Equal to $50k`] += (r.content.Award_Amount__c || 0);
+        projectSummaryV2Row[`${reportingPeriodEndDate} Total Expenditures for Awards Greater or Equal to $50k`] += (r.content.Expenditure_Amount__c || 0);
+        projectSummaryV2Row['Capital Expenditure Amount'] += (r.content.Total_Cost_Capital_Expenditure__c || 0);
+
+        // for kpi report
+        const currentPeriodExpenditure = r.content.Current_Period_Expenditures__c || 0;
+        kpiRow['Number of Subawards'] += (r.type === 'awards50k');
```
ajhyndman, line 152: I think you might need to tweak this slightly:

```suggestion
        kpiRow['Number of Subawards'] += (r.type === 'awards50k') ? 1 : 0;
```

packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -125,22 +132,29 @@ async function getProjectSummaryGroupedByProjectRow(data) {
...
+        kpiRow['Number of Expenditures'] += (currentPeriodExpenditure > 0);
```
ajhyndman, line 153:
```suggestion
        kpiRow['Number of Expenditures'] += Math.max(currentPeriodExpenditure, 0);
```

ajhyndman (changes requested):
Thank you for implementing this change!

I know this has been an ask from partners for many months, so I appreciate you and the team for distilling the requirements and then actually implementing it at last.

packages/server/src/arpa_reporter/lib/audit-report.js
michello: Initially I had that but thought it was redundant since `r.type === 'awards50k'` should result in 1 or 0 for true/false. For example:
```
var hello = 0;
hello += ('1' === '1'); // hello = 1
```

packages/server/src/arpa_reporter/lib/audit-report.js
michello: I don't think currentPeriodExpenditure will be either 1 or 0. This column sound like we want to add up the unique number of currentPeriodExpenditures unless we are adding up the value for that column? 

@as1729 or @jcomeau1 can you chime in on this?

packages/server/src/arpa_reporter/lib/audit-report.js
michello: I thought the bottleneck is making the db queries, within asyncBatch? I wanted to avoid making additional calls hence this approach. The two sheets are using the same data too, just surfacing different rows so I felt it made sense to group together.

If you or others feel strongly though, I can separate haha. By two functions, do we mean have one `getProjectSummaryGroupedByProjectRow` and one `getKPIDataGroupedByProjectRow` method? Or do we mean keep the logic but instead of inlining, refactor into a helper function?

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: Ohh, thank you for explaining this.  I definitely misunderstood the intent.  Again you're coercing true or false to an integer.

My change is definitely not correct then.  Here also, I would slightly prefer a ternary expression instead of a coercion, but I do not feel that strongly.

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: Oh wow, I actually didn't know javascript would coerce true and false to 1 and 0, if you add it like this!

In that case you're right.  This works as is.  I slightly prefer something more explicit, but I don't want to block the change!

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: I don't feel very strongly about this one!  It's definitely okay as it is.

---

I guess I was thinking about splitting apart `getProjectSummaryGroupedByProjectRow`, but keeping a single `createProjectSummariesGroupedByProject`.  I don't actually think `getProjectSummaryGroupedByProjectRow` needs to be async at all, so at least that asyncBatch call should be redundant?

In that case, you might end up with something like this:

```diff
 async function createReportsGroupedByProject(periodId) {
     const records = await recordsForProject(periodId);
     const recordsByProject = getRecordsByProject(records);
     const reportingPeriods = await getAllReportingPeriods();
 
-    const inputs = [];
-
-    Object.entries(recordsByProject).forEach(([projectId, r]) => {
-        inputs.push({ projectId, records: r, reportingPeriods });
-    });
-
-    const reportDataGroupedByProjectData = await asyncBatch(inputs, getReportDataGroupedByProjectRow, 2);
-    const projectSummaryGroupedByProject = reportDataGroupedByProjectData.map((row) => row[0]);
-    const KPIDataGroupedByProject = reportDataGroupedByProjectData.map((row) => row[1]);
+    const projectSummaryGroupedByProject = getProjectSummaryGroupedByProject(reportingPeriods, recordsByProject);
+    const kpiDataGroupedByProject = getKpiDataGroupedByProject(reportingPeriods, recordsByProject);
 
     return [projectSummaryGroupedByProject, KPIDataGroupedByProject];
 }
```

With all of that said, I think it makes a ton of sense to wait until we have some more concrete performance data so we better understand where our real bottlenecks are before doing any meaningful refactoring.

ajhyndman (approved):

PR #1732, opened 2023-08-04
Add datadog spans for file I/O and XLSX parsing

ajhyndman:
## Description

Uses the dd-trace library to capture custom spans for potentially expensive file I/O and XLSX parsing operations.

## Screenshots / Demo Video

This should help fill in some of the gaps in these Datadog traces.

Screen Shot 2023-08-03 at 11 53 11 pm

## Testing

- Boot up reporter locally
- Using any valid sample data, generate either the treasury report or audit reports
- Verify that behaviour remains unchanged

@TylerHendrickson I wonder if we can point development instances to the "sandbox" environment on datadog?  That might be a useful way to validate that behaviour hasn't been broken.

### Automated and Unit Tests
- ~Added Unit tests~

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


tzinckgraf (approved):

packages/server/src/arpa_reporter/services/persist-upload.js
```diff
@@ -152,46 +155,51 @@ async function ensureValidReportingPeriodId(reportingPeriodId) {
...
-
-    // Insert the upload row into the database
-    const upload = await createUpload(uploadRow);
-
-    // Persist the upload to the filesystem
-    await persistUploadToFS(upload, buffer);
-
-    // Return the upload we created
-    return upload;
+    return tracer.trace(
```
tzinckgraf, line 158: Do you need to `await` this call? I was looking at https://datadoghq.dev/dd-trace-js/ and https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/nodejs/?tab=asyncawait#creating-spans

TylerHendrickson:
@ajhyndman re
> @TylerHendrickson I wonder if we can point development instances to the "sandbox" environment on datadog? That might be a useful way to validate that behaviour hasn't been broken.

Yeah, please feel free to generate personal DD credentials and use them in your dev instance while you're debugging something. I hesitate to do this by default on all dev instances (for one, it would require hard-coding the DD creds to the repository; for another, it could easily get expensive, especially since most people probably won't be looking at Datadog during development), and would prefer to refer to the Staging environment (tagged with `env:staging`) for regression analysis.

packages/server/src/arpa_reporter/services/persist-upload.js
ajhyndman: Ahh, I guess awaiting is just a stylistic choice here.

JavaScript is actually pretty forgiving of how you use and return promises in `async` functions.  All of the following behave like functions that return a promise which resolves with the value `1`:


```
function promiseOne() {
    return Promise.resolve(1);
}

async function promiseOne() {
    return 1;
}

async function promiseOne() {
    return Promise.resolve(1);
}

async function promiseOne() {
    return await Promise.resolve(1);
}
```

Our linter is currently configured to complain about the use of a "redundant await" here if I include it, so I left that out.

image

ajhyndman:
@TylerHendrickson great, actually this makes a ton of sense.  Thanks!

packages/server/src/arpa_reporter/services/persist-upload.js
tzinckgraf: I learn something new about Javascript every day.

TylerHendrickson (approved):

tzinckgraf (approved):

packages/server/src/arpa_reporter/services/persist-upload.js
as1729: @ajhyndman I'm receiving this error in staging- I suspect it has to do with these changes potentially. 
image

Issue #1703, opened 2023-07-27
Feature flag for Bill field on Advanced Search form


ClaireValdivia:
### Why is this issue important?

We only want the Bill field to appear in the Advanced Search panel if we are able to backfill data from FFIS in this search milestone. Therefore, we want to add a feature flag that controls whether the "Bill" field will display so we can turn this on/off based on whether the backfill of data is complete when we are ready to deploy.

### Current State

na

### Expected State

na

PR #1754, opened 2023-08-07
Feat: streamline Terraform website feature flags

TylerHendrickson:
## Description

This PR is intended to reduce the number of steps involved when configuring client-side feature flags provided by `window.APP_CONFIG` in Terraform. For reference, the [opening comment](https://github.com/usdigitalresponse/usdr-gost/issues/1703#issue-1825066533) in #1703 shows the steps currently involved in adding a feature flag to `window.APP_CONFIG`. If this PR is accepted, the number of Terraform-specific steps is reduced to adding the feature flag key/value pair to the `website_feature_flags` input variable defined in the stage-specific `terraform/*.tfvars` file.

Additionally, this PR refactors the `packages/client/src/helpers/featureFlags.js` file into a directory-based module located at `packages/client/src/helpers/featureFlags/`, where the directory's `index.js` file exports lookup functions. This was done in order avoid cluttering the module-level exports with the newly-added `getFeatureFlag()` helper function that is only intended to be used within the module (and in tests, which were also added in this PR).

Changes are intended to be backwards-compatible; all existing client-facing behavior is unmodified, as well as existing imports for the `useNewGrantsTable()` lookup function.

**Edit: Quick summary of changes added after initial review**
1. Added documentation for feature flags at `docs/feature-flags.md`
2. Consolidated sourcing for all feature flags to the `window.APP_CONFIG.featureFlags` object and removed the previous `VUE_APP_USE_NEW_TABLE` env var, which is no longer needed.
3. Moved `packages/client/public/deploy-config.example.js` to `deploy-config.js` and updated `vue.config.js` to exclude it from webpack builds, which makes it available for use (for managing feature flags) in development environments.
4. Bonus / cleanup:
    1. Removed `terraform/production.tfvars` and `terraform/production.s3.tfbackend`. This should have been done long ago, and their existence is a leftover from our initial migration to AWS – we use `prod.tfvars` and `prod.s3.tfbackend`, and it was getting hard to remember which one was actually used for deployment.
    2. Deprecated `window.apiURLForGOST` in favor of a nested `window.APP_CONFIG.apiURLForGOST`. This was also a long time coming. The benefit is more consolidated usage of the `window.APP_CONFIG` object and correspondingly, slightly less pollution of the global `window` namespace.

## Testing

1. Deploy to a LocalStack/sandbox environment, providing a value for `var.website_feature_flags` input variable, e.g.
    ```shell
    $ terraform apply -var='website_feature_flags={"useNewTable": true, "someFeatureFlag": false}'
    ```
5. Using a browser debugger tool, ensure that `window.APP_CONFIG.featureFlags` evaluates to an object that corresponds to the input value provided in Step 1.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


terraform/staging.tfvars
```diff
@@ -30,6 +30,9 @@ website_managed_waf_rules = {
     metric_visibility = true
   }
 }
+website_feature_flags = {
```
as1729, line 33: this ensures that feature-flags are similar to environment variables in that they are either on/off for a given environment, correct?

If that is the case it looks like we will lose the current functionality of being able to toggle the flag for any environment which is helpful when we want to support two pieces of functionality in parallel.

lmk if I'm misunderstanding the implementation.

replicantSocks (approved):
Looks good.

I'm wondering if it's worth adding a note, possibly in development.md, or possibly elsewhere, that of where to look for the client side feature flags. My thought is that this is actively tied to feature release, and a useful pointer in getting started interacting with the team.

terraform/staging.tfvars
TylerHendrickson: @as1729 The logic that "activates" a feature flag is entirely dependent on what the relevant FE code chooses to evaluate. The primary change in this PR is really just two things:
1. Establishing a nested `window.APP_CONFIG.featureFlags` object and moving all existing (currently there's only one) feature flags into that object, rather than having them set directly on `window.APP_CONFIG`.
2. Adding this `website_feature_flags` Terraform input variable and wiring it up to populate the `window.APP_CONFIG.featureFlags` object directly, instead of requiring multiple edits throughout the Terraform code in order to set feature flags.

The use of environment variables in evaluating the "status" of a flag is not interrupted (see [this line](https://github.com/usdigitalresponse/usdr-gost/pull/1754/files#diff-31ae1b1a4f152b03716b188dbd79cda77f6fb27e8a4e1dfeae8467f5b436cd0bR7)), although it's worth noting that env vars are only effective in dev environments – Vue/webpack only evaluate env vars at build-time (and the way they get built makes them immutable), which makes them unsuitable for dynamic toggling and/or setting different values between Staging and Production. You can think of it this way:
- In dev environments, use environment variables to set and source feature flags.
- In Staging or Prod environments, use `window.APP_CONFIG.featureFlags` to set and source feature flags.
    - It's worth noting that `window.APP_CONFIG.featureFlags` _could_ be used in dev environments, but we don't currently have any mechanism to populate this object in dev environments, so it would have to be done manually in the browser console.

This effectively means that there are two potential sources for a feature flag's value: an associated env var (e.g. `VUE_APP_USE_NEW_TABLE`) and a runtime variable (e.g. `window.APP_CONFIG.featureFlags.useNewTable`), and the effective value is dependent on how the FE code chooses to evaluate them. In the case of the (currently only) `useNewTable` feature flag, the FE code considers it "active" as long as either are set to `true`.

Eventually, I think it would be worth moving entirely to `window.APP_CONFIG.featureFlags` to remove this ambiguity and establish a single way to manage feature flags, but as mentioned above, we currently lack an established method for populating `window.APP_CONFIG.featureFlags` automatically in dev environments.

TylerHendrickson:
> Looks good.
> 
> I'm wondering if it's worth adding a note, possibly in development.md, or possibly elsewhere, that of where to look for the client side feature flags. My thought is that this is actively tied to feature release, and a useful pointer in getting started interacting with the team.

@replicantSocks Good call. I'll update this PR with some documentation.

terraform/staging.tfvars
as1729: Got it, makes sense, thanks!

TylerHendrickson: @as1729 The most-recent batch of updates to this branch contradicts a few of my statements in the above comment, namely that I added tooling to consolidate sourcing feature flags from `window.APP_CONFIG.featureFlags`, meaning that env vars are no longer needed. However, part of this update enables developers to modify feature flags in development environments by making changes to the (dev-only) `packages/client/public/deploy-config.js` file.

Let me know if that raises any concerns on your end.

TylerHendrickson:
> > Looks good.
> > I'm wondering if it's worth adding a note, possibly in development.md, or possibly elsewhere, that of where to look for the client side feature flags. My thought is that this is actively tied to feature release, and a useful pointer in getting started interacting with the team.
> 
> @replicantSocks Good call. I'll update this PR with some documentation.

@replicantSocks Updated the documentation and added a bit more to this PR- can you take another pass?

packages/client/vue.config.js
```diff
@@ -6,6 +6,12 @@ module.exports = {
   configureWebpack: {
     devtool: 'source-map',
   },
+  chainWebpack: (config) => {
+    config.plugin('copy').tap(([options]) => {
+      options.patterns[0].globOptions.ignore.push('**/deploy-config.js');
+      return [options];
+    });
+  },
```
TylerHendrickson, line 14: Adapted from [this StackOverflow answer](https://stackoverflow.com/a/60166838).

Would appreciate feedback (in either direction) from someone more familiar with webpack (and its Vue configuration) than I am... which is probably most people 🙃

terraform/staging.tfvars
as1729: These changes look good. Agreed on eliminating need to set env-vars. I think it does simplify the implementation.

as1729 (approved):

packages/client/vue.config.js
as1729: Can't say that I can offer another suggestion here. Does this work as intended locally and is there a test we can do when its deployed to `staging` to confirm if its working as we expect to in hosted environments as well?

TylerHendrickson: > Does this work as intended locally
It does!

> is there a test we can do when its deployed to staging to confirm if its working as we expect to in hosted environments as well?

Yes, once deployed to staging, there are two checks that can be done:
1. The site shouldn't be broken. Since this `deploy-config.js` file sets `apiURLForGOST = "/"`, it's basically relying on the Vue dev server's proxy to the API backend. Since that proxy doesn't exist in non-dev environments and since the API in Staging/Prod is hosted on a different subdomain, this version of `deploy-config.js` would break requests to the backend.
2. I'm 99.9% confident that the above won't happen, because our CloudFront CDN explicitly redirects requests to `/deploy-config.js` to the S3 object managed by Terraform. However, I'd still want to check that this file is absent in the S3 bucket path that webpack bundles get deployed to (which would confirm that this file is properly excluded from builds).

I'm honestly less concerned about the `deploy-config.js`-specific behavior here and more interested in making sure that there aren't any unintended consequences of this new configuration that may have slipped by me. My confidence is boosted because this snipped of code seems fairly ubiquitous in StackOverflow, GitHub issues, and the Vue support forum... so at least we're in good company! 😅

PR #1771, opened 2023-08-10
FEAT: Add reusable `sqs_consumer_task` terraform module

TylerHendrickson:
## Description

This PR defines a reusable `sqs_consumer_task` Terraform module, which can be used for provisioning ECS services with task pools that scale up and down in capacity based on the number of messages available in an SQS queue (also provisioned by this module). I adapted this from the existing module at `terraform/modules/gost_consume_grants`, which was designed to behave similarly but not to be reusable for multiple use-cases.

Since nothing is actually using this new module at the moment, the changes are effectively a no-op (and will not show up in Terraform plan output), but I'm providing it now to address potential interest in moving async task execution to a standalone ECS tasks with a dedicated work queue. Although it would be useful for an actual, known use-case to vet this implementation, I think it hopefully provides a decent starting point.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [X] Added PR reviewers


terraform/modules/sqs_consumer_task/compute.tf
```diff
@@ -0,0 +1,318 @@
...
+    ignore_changes = [desired_count]
+  }
+
+  depends_on = [
+    aws_iam_role.execution,
+    aws_iam_role_policy.execution,
+  ]
+}
+
+module "consumer_task_definition" {
```
jakekreider, line 29: Should this have a [`container_depends_on`](https://github.com/cloudposse/terraform-aws-ecs-container-definition#input_container_depends_on) for the datadog container?

jakekreider (approved):
I haven't tested it but the changes all make sense, including the documentation. Happy to be an early adopter if there's a ticket you have in mind!

terraform/modules/sqs_consumer_task/compute.tf
TylerHendrickson: Good call – this PR auto-merged already, but I'll open a new one with this change.

PR #1840, opened 2023-08-28
feat(arpa): skip parsing of formulae and html

ajhyndman:
## Description

We shouldn't actually need to preserve any formula or HTML formatting when loading an upload from disk.  We can skip this parsing to save processing time, memory and disk space (when caching uploads to JSON).

This change also adds two new tracing spans, breaking down `jsonForUpload` into disk reading and JSON parsing components.  It also adds a file size tag to the `fs.readFile` span.

## Screenshots / Demo Video

image

When testing locally, cached JSON file sizes shrink by 2–300 kB with these options set.

https://app.datadoghq.com/apm/trace/755985303388168448?colorBy=service&env=sandbox&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=755985303388168448&spanViewType=metadata&timeHint=1693184401669.03

## Testing

1. Install and start up gost service locally and log in
2. Upload at least one [valid test file](https://drive.google.com/drive/folders/1B9ooTSJK5H2pCKHl5zgs4TjNeOSol1f7).
3. Generate an audit report by visiting http://localhost:8080/api/audit_report
4. Check the size of the generated JSON file (at `packages/server/data/uploads/tmp`)

### Automated and Unit Tests
- ~Added Unit tests~

### Manual tests for Reviewer

## Checklist
- ~Provided ticket and description~
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/services/persist-upload.js
+++ b/packages/server/src/arpa_reporter/services/persist-upload.js
@@ -250,7 +250,7 @@ async function jsonForUpload(upload) {
                 const { size } = await fs.stat(jsonFSName(upload));
                 span.setTag('filesize-kb', Math.round(size / (2 ** 10)));
                 span.setTag('tenant-id', upload.tenant_id);
-                span.setTag('reporting-period-id', upload.reporting_perdiod_id);
+                span.setTag('reporting-period-id', upload.reporting_period_id);
                 return f;
             });
             return tracer.trace('Cryo.parse', () => Cryo.parse(file));
--- a/packages/server/src/arpa_reporter/services/persist-upload.js
+++ b/packages/server/src/arpa_reporter/services/persist-upload.js
@@ -248,9 +248,9 @@ async function jsonForUpload(upload) {
             const file = await tracer.trace('fs.readFile', async (span) => {
                 const f = await fs.readFile(jsonFSName(upload), { encoding: 'utf-8' });
                 const { size } = await fs.stat(jsonFSName(upload));
-                span.setTag('filesize-kB', Math.round(size / (2 ** 10)));
-                span.setTag('tenantId', upload.tenant_id);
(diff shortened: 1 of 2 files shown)


packages/server/src/arpa_reporter/services/persist-upload.js
```diff
@@ -244,7 +244,15 @@ async function bufferForUpload(upload) {
 async function jsonForUpload(upload) {
     return tracer.trace(
         'jsonForUpload',
-        async () => Cryo.parse(await fs.readFile(jsonFSName(upload), { encoding: 'utf-8' })),
+        async () => {
+            const file = await tracer.trace('fs.readFile', async (span) => {
+                const f = await fs.readFile(jsonFSName(upload), { encoding: 'utf-8' });
+                const { size } = await fs.stat(jsonFSName(upload));
+                span.setTag('filesize-kB', Math.round(size / (2 ** 10)));
```
tzinckgraf, line 251: If we are adding some tags on here, could we add the tenant id / reporting period id as well? Thoughts?

ajhyndman: We can add those!  Do you want them on this span specifically, or somewhere else?

tzinckgraf: I think this span makes sense. We can always start with this and see if we want to add it elsewhere.

ajhyndman: I pushed another change and it looks like I succeeded in adding these two tags!

image

https://app.datadoghq.com/apm/trace/7692138087698454584?colorBy=service&env=sandbox&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=7692138087698454584&spanViewType=metadata&timeHint=1693189604002.007

tzinckgraf (approved):

PR #1841, opened 2023-08-28
docs: Document local workflow for datadog debugging

ajhyndman:
Sets some example env variables and documents the workflow I was able to use to test datadog tracing locally.

### DIFF
--- a/docs/setup-mac.md
+++ b/docs/setup-mac.md
@@ -97,3 +97,26 @@ See [./setup-gmail.md](./setup-gmail.md) for instructions on setting up Gmail to
     ```sh
     yarn workspace server run db:seed
     ```
+
+## Datadog
+
+It's possible to send tracing data from your local development instance to datadog.  To do so, you will need to:
+
+1. Install a [Datadog Agent](https://docs.datadoghq.com/getting_started/agent/) locally and give it credentials to send data to the USDR datadog instance.
+1. Configure the datadog tracing library with environment variables
+
+### Install the Datadog Agent
+
+Datadog provides agent installation instructions within the [datadog.hq app](https://app.datadoghq.com/account/settings/agent/latest?_gl=1%2Ancclch%2A_gcl_aw%2AR0NMLjE2OTI3NDkyNzEuQ2owS0NRand1WkduQmhEMUFSSXNBQ3hiQVZqQS13YTlHVFFORGxQaVlFWDQzb1hiWTdHYzNxTFpScXhQam84UFg3cnRqTHZxT1UxSWdNY2FBbzJRRUFMd193Y0I.%2A_gcl_au%2AMTUyNDczOTE5Mi4xNjkyMjQxMDkw%2A_ga%2AMTQ3MTcxNzMzMy4xNjkzMDIxMDQ2%2A_ga_KN80RDFSQK%2AMTY5MzE5MDg2Mi41LjEuMTY5MzE5MDkxNC4wLjAuMA..%2A_fplc%2ARnhXc1BjdFZjcmF6dFZmSmpJNDlXRDUzeXVpdVdtZWI4eDhnazhvYVdpUkI1dWRESFBwVzN1QnNIRXBDVHllJTJCdVdBZUxpJTJGJTJCWCUyRjNnUWVnQWgycWVSb3IyR2tyeWp6T0pSR01oc2YlMkZHWkN5MWd6RTRXS1lWdU80bnR3JTJGJTJCb2clM0QlM0Q.&platform=macos).
+
+Select (or create) an API key and copy the corresponding install command, which should automatically install and start the agent.
+
+### Configure tracing library
+
+Next, make sure your local `packages/server/.env` file includes the recommended datadog environment variables (specified in `packages/server/.env.example`).
+
+### Query traces from datadoghq.com
+
+Now you should be able to start your development service normally. If your environment is set correctly, the dd-trace library should print startup logs confirming the configuration you have specified.
+
+Finally, tracing data should now be available on datadog.  You can filter for your local traces by specifying the `env:sandbox` tag in your trace query.
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -46,3 +46,9 @@ AUDIT_REPORT_BUCKET=arpa-audit-reports
 
 # SQS queue for grants-ingest events
 GRANTS_INGEST_EVENTS_QUEUE_URL="http://localhost:4566/000000000000/grants-ingest-events"
+
+# Datadog tracing configuration
+DD_ENV="sandbox"
+DD_SERVICE="gost"
+DD_PROFILING_ENABLED=true
+DD_TRACE_STARTUP_LOGS=true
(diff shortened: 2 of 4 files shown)


tzinckgraf (approved):

docs/setup-mac.md
```diff
@@ -97,3 +97,26 @@ See [./setup-gmail.md](./setup-gmail.md) for instructions on setting up Gmail to
...
+## Datadog
+
+It's possible to send tracing data from your local development instance to datadog.  To do so, you will need to:
+
+1. Install a [Datadog Agent](https://docs.datadoghq.com/getting_started/agent/) locally and give it credentials to send data to the USDR datadog instance.
+1. Configure the datadog tracing library with environment variables
+
+### Install the Datadog Agent
+
+Datadog provides agent installation instructions within the [datadog.hq app](https://app.datadoghq.com/account/settings/agent/latest?_gl=1%2Ancclch%2A_gcl_aw%2AR0NMLjE2OTI3NDkyNzEuQ2owS0NRand1WkduQmhEMUFSSXNBQ3hiQVZqQS13YTlHVFFORGxQaVlFWDQzb1hiWTdHYzNxTFpScXhQam84UFg3cnRqTHZxT1UxSWdNY2FBbzJRRUFMd193Y0I.%2A_gcl_au%2AMTUyNDczOTE5Mi4xNjkyMjQxMDkw%2A_ga%2AMTQ3MTcxNzMzMy4xNjkzMDIxMDQ2%2A_ga_KN80RDFSQK%2AMTY5MzE5MDg2Mi41LjEuMTY5MzE5MDkxNC4wLjAuMA..%2A_fplc%2ARnhXc1BjdFZjcmF6dFZmSmpJNDlXRDUzeXVpdVdtZWI4eDhnazhvYVdpUkI1dWRESFBwVzN1QnNIRXBDVHllJTJCdVdBZUxpJTJGJTJCWCUyRjNnUWVnQWgycWVSb3IyR2tyeWp6T0pSR01oc2YlMkZHWkN5MWd6RTRXS1lWdU80bnR3JTJGJTJCb2clM0QlM0Q.&platform=macos).
```
TylerHendrickson, line 110: I would very much prefer we direct people to use the Docker-based agent rather than having them install a Mac or Windows system service, as it will be much easier for us to support any issues someone might run into if we can assume a Docker-based environment.

We could even provide a Dockerfile and some simple commands to start/stop the docker-based agent when needed, which I think would be really useful to future contributors.

docs/setup-mac.md
```diff
@@ -97,3 +97,26 @@ See [./setup-gmail.md](./setup-gmail.md) for instructions on setting up Gmail to
     ```sh
     yarn workspace server run db:seed
     ```
+
+## Datadog
```
TylerHendrickson, line 101: I'm not sure that basic setup steps for working on this repo should direct folks to use Datadog at all –it should rarely be necessary for local environments, and I'd like to think that's the exception (reserved for advanced debugging/investigation scenarios) rather than the norm. I think there are a few reasons why:
1. Cost. Although I don't think it's a problem for people to generate and send local runtime/APM metrics to Datadog for specific cases, I think it _should_ be reserved for intentional investigation, rather than something that's "on by default". Otherwise, we'll end up paying for a lot of throw-away development data that nobody ever uses. 
2. Related to cost, I think having these instructions geared towards a one-time setup as a system service will inspire "set-and-forget" behavior, and I would like to make sure that as much as possible, folks turn off local Datadog reporting once a specific debugging exercise or investigative effort is concluded. 
3. Development complexity, and a desire to avoid having contributors go through additional setup steps for their dev environment when they might not actually be using Datadog (since my guess is most people won't).

To turn all this into a concrete ask for this PR:
1. It's worth splitting out this Datadog section into a separate, standalone document, rather than bundling it into dev environment setup instructions.
2. Within that separate document, add some introductory text (maybe around the content that's currently on line 103) that clarifies when you might actually want/need to do this. In particular, I think it should be a generally driven by a team consensus ("yes, this is something we need to look into, and there are reasons why it would be useful for being able to experiment in local development").

packages/server/.env.example
```diff
@@ -46,3 +46,9 @@ AUDIT_REPORT_BUCKET=arpa-audit-reports
 
 # SQS queue for grants-ingest events
 GRANTS_INGEST_EVENTS_QUEUE_URL="http://localhost:4566/000000000000/grants-ingest-events"
+
+# Datadog tracing configuration
+DD_ENV="sandbox"
```
TylerHendrickson, line 51: I think it would be ideal for people to use some kind of unique modifier rather than tagging everything as `env:sandbox`. For example, I've tended to use `env:sandbox-tsh` for local development, as it helps ensure that the data I'm looking at came from my own debugging activities and not from someone else's.

docs/setup-mac.md
```diff
@@ -97,3 +97,26 @@ See [./setup-gmail.md](./setup-gmail.md) for instructions on setting up Gmail to
...
+
+### Configure tracing library
+
+Next, make sure your local `packages/server/.env` file includes the recommended datadog environment variables (specified in `packages/server/.env.example`).
+
+### Query traces from datadoghq.com
+
+Now you should be able to start your development service normally. If your environment is set correctly, the dd-trace library should print startup logs confirming the configuration you have specified.
+
+Finally, tracing data should now be available on datadog.  You can filter for your local traces by specifying the `env:sandbox` tag in your trace query.
```
TylerHendrickson, line 122: Re `env:sandbox`: see my comment on the `.env.example` file.

TylerHendrickson (changes requested):

docs/setup-mac.md
ajhyndman: I don't have a problem with us taking the stance to prefer docker workflows.  That's actually why I thought this particular documentation was a good fit for this mac-specific local development document.

I also think it makes a lot of sense to document or add configuration for the docker workflow, but that's not what I did so far.  I just thought I would offer to capture some notes from what I did work out, in case it's helpful.

If it's not, we don't have to land anything for now.

packages/server/.env.example
ajhyndman: I don't think this sample configuration precludes that workflow if there is ever any confusion, does it?  I also think it would just be _more_ confusing if I put my initials in here and someone copy-pasted the example configuration without editing.

TylerHendrickson: For sure – This comment was really just meant to contextualize my [other comment](https://github.com/usdigitalresponse/usdr-gost/pull/1841#discussion_r1307708551) related to the documentation – figured it might be worth adding a note around this in your markdown docs.

docs/setup-mac.md
TylerHendrickson: My preference would be to generally only do a documentation write-up for recommended patterns/procedures we'd like contributors to be able to repeat with the benefit of documented steps. 

For context, newer volunteers tend to refer to the `setup-mac.md` documentation when they're getting started with the project, so even though the intent for this might be "this is something that worked for me in a specific scenario", it's likely to be interpreted by others as "steps I should follow in order to fully configure my environment for local development".

PR #1847, opened 2023-08-29
fix: ensure save button disabled for new saved searches

replicantSocks:
## Description
To enable the save button we check that no values have been entered into the search form by checking looking at whether they are null or empty. The 'costSharing' field defaulted to false, which meant all searches start as save-able even before the radio button is clicked.

## Screenshots / Demo Video

Screenshot from 2023-08-29 16-43-19

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/Modals/SearchPanel.vue
+++ b/packages/client/src/components/Modals/SearchPanel.vue
@@ -171,7 +171,7 @@ const defaultCriteria = {
   fundingTypes: null,
   agency: null,
   bill: null,
-  costSharing: false,
+  costSharing: null,
   opportunityCategories: [],
   reviewStatus: [],
   postedWithin: [],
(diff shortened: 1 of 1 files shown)


as1729:
@replicantSocks I'm good with this fix given that the component was setup this way.

However, I think the ideal setup would be to have a `this.isDirty` variable in the state that holds whether the form has been manipulated.

Looks like Vuejs does give us this functionality out of the box using the `watch` api as explained in this [stackoverflow](https://stackoverflow.com/questions/52343140/how-to-implement-dirty-state-in-vuejs).

If @jakekreider and you agree on this approach it will be much appreciated if you're able to implement!

as1729 (approved):

PR #1873, opened 2023-09-08
Parallelize Audit Report generation

ajhyndman:
## Description

In January, we ran into some performance bottlenecks in ARPA report generation.  At the time, the biggest issue was that XLSX parsing was both synchronous and the primary bottleneck, which could cause the server's main thread to lock up for the duration of report generation — preventing other requests from being served.

This issue was successfully mitigated by the introduction of `asyncBatch` with #834.  This improved on the situation by forcing the server to batch XLSX parsing jobs and "come up for air" between batches so that the server had the chance to handle additional requests before continuing report generation.

Following up on this initial mitigation, we introduced a filesystem cache layer for XLSX parsing (#942).

We are still seeing performance issues, but the situation is different than it was in January in a few interesting ways.
1. Instead of parsing XLSX files, reading from the filesystem cache now appears to be our biggest bottleneck.
2. Unlike XLSX parsing, reading from disk is _not_ main thread blocking, and _can_ benefit from parallelization

Building on the above with some [insights from Datadog testing](https://usdigitalresponse.slack.com/archives/C031R1Y49KL/p1693065541547659), it looks like our usages of asyncBatch may now be unnecessarily serializing some of the disk access calls which _could_ be parallelized.  For production reports with many uploads, this effect may be further exacerbated (but this remains to be verified)!

This change attempts to reclaim the benefits of parallelizing disk access by rewriting a few of our asyncBatch usages, allowing all disk reads to once again run in parallel.

## Screenshots / Demo Video

BEFORE

image

AFTER

image

## Testing

1. Boot up reporter locally
1. Visit http://localhost:8080/api/audit_report to generate an audit report synchronously
1. Verify that behaviour remains unchanged before and after the change

NOTE: _Before_ landing this change, I would really like to have some production tracing in place: both to validate this hypothesis and also to ensure that we can get a clear image of the before and after effect of this change.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -97,18 +97,11 @@ async function getReportDataGroupedByProjectRow(data) {
         'Project Expenditure Category': record.subcategory,
     };
 
-    const kpiRow = {
-        'Project ID': projectId,
-        'Number of Subawards': 0,
-        'Number of Expenditures': 0,
-        'Evidence Based Total Spend': 0,
-    };
-
     // get all reporting periods related to the project
     const allReportingPeriods = Array.from(new Set(records.map((r) => r.upload.reporting_period_id)));
 
     // initialize the columns in the row
-    allReportingPeriods.map(async (reportingPeriodId) => {
+    allReportingPeriods.forEach(async (reportingPeriodId) => {
         const reportingPeriodEndDate = reportingPeriods.filter((reportingPeriod) => reportingPeriod.id === reportingPeriodId)[0].end_date;
         [
             `${reportingPeriodEndDate} Total Aggregate Expenditures`,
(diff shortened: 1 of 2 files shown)


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -208,17 +136,69 @@ async function createReportsGroupedByProject(periodId) {
...
+            'Project Description': record.content.Project_Description__c,
+            'Project Expenditure Category Group': ec(record.type),
+            'Project Expenditure Category': record.subcategory,
+        };
+
+        // get all reporting periods related to the project
+        const allReportingPeriods = Array.from(new Set(projectRecords.map((r) => r.upload.reporting_period_id)));
+
+        // initialize the columns in the row
+        allReportingPeriods.forEach(async (reportingPeriodId) => {
```
tzinckgraf, line 154: Does this need async?

tzinckgraf (approved):

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: It does not at all.  Thank you, great catch!

PR #1906, opened 2023-09-13
fix: broken award-ceiling ordering

as1729:
## Description
- Award-ceiling descending is fixed by putting NULLS LAST
- Export CSV button is scoped to a smaller area so user will not mistakenly trigger it if they hit the award ceiling column

### Before
image

### After
image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (changes requested):
I think this could be an acceptable change for the time-being, but in the long-term, it might make more sense to just start storing award amounts as `bigint` using a database migration. That way they only have to be cast from string on insert/update, rather than on retrieval.

If you think that's not worth doing at the moment, let me know and I can 👍 this PR.

packages/client/src/components/GrantsTableNext.vue
```diff
@@ -16,10 +16,7 @@
         <h4 class="mb-0">{{ searchTitle }}</h4>
       </b-col>
       <b-col align-self="end">
-        <a href="#" @click="exportCSV" :disabled="loading" variant="outline-primary border-0"
-          class="text-right text-nowrap">
-          <p class="mb-0">Export CSV</p>
-        </a>
+        <p class="mb-0 text-right text-nowrap"><a href="#" @click="exportCSV" :disabled="loading" variant="outline-primary border-0">Export CSV</a></p>
```
as1729, line 19: small css change that ended up revealing itself when testing award-ceiling since the entire <p> tag was a link, when I was clicking award-ceiling it would end up downloading a CSV if I mis-clicked by a hair.

packages/server/migrations/20230914014057_convert-award-ceiling-data-type.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.raw(`ALTER TABLE grants ALTER COLUMN award_ceiling TYPE bigint USING award_ceiling::bigint`);
```
as1729, line 6: Note: this does take ~ 10seconds to run locally so I wouldn't suggest to run this migration during heavy usage periods. Will need to be a bit more mindful when we push to production.

TylerHendrickson: FWIW, I think your idea of using a multi-step migration would mitigate the downtime concerns – you could adjust a bit to something like this:
1. Add new bigint column named `award_ceiling_temp` using values cast from the existing `award_ceiling` column
2. Rename columns (single transaction):
    - `award_ceiling` -> `award_ceiling_old`
    - `award_ceiling_temp` -> `award_ceiling`
3. Drop column `award_ceiling_old` (note: you might be able to do this in step 2 instead of renaming to `award_ceiling_old`)

This "add+populate+rename+drop" operation is a pretty typical pattern for "altering" existing columns on a high-traffic DB.

TylerHendrickson: Also, if we're moving `award_ceiling` to bigint, what do you think about doing the same for `award_floor` – I would imagine it would have the same sorting issues. 

A bigint might not be as necessary for storing `award_floor` values, but might be worth doing anyways for consistency and/or "just in case".

as1729: > This "add+populate+rename+drop" operation is a pretty typical pattern for "altering" existing columns on a high-traffic DB.

Agreed on this in theory. Should we do that for our current application given the usage patterns?

as1729: Resolved in a separate PR.

TylerHendrickson (approved):

PR #1907, opened 2023-09-13
fix: ignore closed and archived grants by default

as1729:
## Description
- Search panel defaults to selecting Posted and Forecasted
- Search panel consolidates Closed and Archived
- Old search defaults to Posted and Forecasted

### Old search flow defaults to filtering by posted or forecasted
image

### Can access all grants in old search flow by explicitly selecting all grant statuses
image

### New search flow defaults to loading only posted and forecasted grants
image

### New search panel
image

### New search loads all grants if explicitly selected
image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/components/GrantsTable.vue
+++ b/packages/client/src/components/GrantsTable.vue
@@ -143,7 +143,7 @@ export default {
       searchInput: null,
       debouncedSearchInput: null,
       reviewStatusFilters: [],
-      opportunityStatusFilters: [],
+      opportunityStatusFilters: ['Forecasted', 'Posted'],
       opportunityCategoryFilters: [],
       costSharingFilter: null,
       reviewStatusOptions: ['interested', 'result', 'rejected'],
--- a/packages/client/src/components/Modals/SearchPanel.vue
+++ b/packages/client/src/components/Modals/SearchPanel.vue
@@ -226,7 +226,7 @@ const defaultCriteria = {
   includeKeywords: null,
   excludeKeywords: null,
   opportunityNumber: null,
-  opportunityStatuses: ['forecasted', 'posted'],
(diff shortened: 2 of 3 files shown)


packages/client/src/components/GrantsTable.vue
```diff
@@ -143,7 +143,7 @@ export default {
       searchInput: null,
       debouncedSearchInput: null,
       reviewStatusFilters: [],
-      opportunityStatusFilters: [],
+      opportunityStatusFilters: ['Forecasted', 'Posted'],
```
TylerHendrickson, line 146: Since we're going to be dropping forecasts for the time-being, shouldn't that be removed from the defaults on the new table? 
```suggestion
      opportunityStatusFilters: ['Posted'],
```

TylerHendrickson (approved):
@as1729 FYI – I disabled auto-merge on this PR in case you wanted to modify anything related to my comment, but approving because I think this change looks good based on the PR description.

Basically, wondering if it makes sense to update this PR to exclude the `Forecasted` option entirely from the new table view. If you're planning on doing that in a separate PR, though, that works too.

packages/client/src/components/GrantsTable.vue
as1729: This is the old search experience. The only change I wanted to do here is to make it default ignore the `Closed / Archived` grants for performance improvements.

The idea here was to enable deployment to production so we can get grants-ingest without breaking the old table.

as1729: To confirm- the new experience, doesn't have forecasted grants at all. It is fully hidden from view.

PR #1920, opened 2023-09-14
fix: ensures opportunity_status is up to date

as1729:
## Description
- Ensures `archive_date` is set when we capture new grants
- Calculates `opportunity_status` at query-time based on the values of `open_date` `close_date` and `archive_date`

### After
image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/migrations/20230914192911_add-archive-date-to-grants.js
```diff
@@ -0,0 +1,21 @@
+/**
```
as1729, line 1:
```
bash-5.1# yarn db:migrate
yarn run v1.22.17
$ knex migrate:latest
Using environment: development
Batch 4 run: 1 migrations
Done in 12.76s.
bash-5.1# yarn db:rollback
yarn run v1.22.17
$ knex migrate:rollback
Using environment: development
Batch 4 rolled back: 1 migrations
Done in 13.01s.
bash-5.1# 
```

as1729: ## Before
image


## After
image

as1729:
@TylerHendrickson can you give a preliminary review on whether this is the approach you were expecting to see. Since these fields are in `text` type in the database we either need to change that or cast it to `::timestamp`

packages/server/migrations/20230914192911_add-archive-date-to-grants.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    // , ALTER COLUMN close_date TYPE timestamp USING close_date::timestamp, ADD COLUMN archive_date TYPE timestamp
+    return knex.raw(`ALTER TABLE grants ALTER COLUMN open_date TYPE timestamp USING open_date::timestamp,
+    ALTER COLUMN close_date TYPE timestamp USING close_date::timestamp,
+    ADD COLUMN archive_date timestamp`);
```
TylerHendrickson, line 9: I'd suggest using `date` as the new type for these columns:
```suggestion
    return knex.raw(`ALTER TABLE grants ALTER COLUMN open_date TYPE date USING open_date::date,
    ALTER COLUMN close_date TYPE date USING close_date::date,
    ADD COLUMN archive_date date`);
```

Note: These can (also) compare to timestamps like `now()`, so it should be perfectly compatible with our use-case, e.g. 
```sql
select '2023-09-15'::date < now() as is_before;
 is_before 
-----------
 t
(1 row)

select now()::date < now() as is_before;  -- Compared to a timestamp, date represents "midnight"
 is_before 
-----------
 t
(1 row)

select '2023-09-15'::date as some_date;
 some_date  
------------
 2023-09-15
(1 row)

select '2023-09-15'::date::timestamp as some_timestamp;  -- Easily converted to timestamp if needed
      some_date      
---------------------
 2023-09-15 00:00:00
(1 row)

select '2023-09-15'::date::text as some_text;  -- Compatible with the downward migration
 some_text  
------------
 2023-09-15
```

TylerHendrickson (changes requested):
Requesting a small change for using `date` type instead of `timestamp without time zone`, which is a small difference but I think is more intuitive.

Also, do you think it's worth doing a multi-phase migration as we discussed for the `award_ceiling` and `award_floor` migrations? Would be something like
```sql
-- Initial migration
ALTER TABLE grants
ADD COLUMN open_date_new date USING open_date::date, 
ADD COLUMN close_date_new date using close_date::date;

-- Second step
ALTER TABLE grants
RENAME COLUMN open_date TO open_date_old,
RENAME COLUMN open_date_new TO open date,
RENAME COLUMN close_date TO close_date_old,
RENAME COLUMN close_date_new TO close_date;

-- Last step
ALTER TABLE grants
DROP COLUMN open_date_old,
DROP COLUMN close_date_old;
```

packages/server/migrations/20230914192911_add-archive-date-to-grants.js
as1729: Added separately in #migra

packages/server/migrations/20230914192911_add-archive-date-to-grants.js
as1729: Done 👍

TylerHendrickson (approved):

PR #1946, opened 2023-09-20
fix: email notifications not sending the correct grants-set

as1729:
## Description
- The reason for wrong emails being sent is because the email service is passing in the raw `criteria` from the `grants_saved_search`. This PR ensures that: 
  1. The email-service is formatting the input-filters correctly prior to sending it to `getGrantsNew` function.
  2. The `getGrantsNew` function validates all input filters being passed in and sends back an appropriate error message. Note: the lack of a clear error message was the reason for a longer debug time for these issues.

## Testing
To manually test:
Sample test URL:
http://localhost:8080/api/organizations/0/users/enter-your-local-user-id/sendDigestEmail?date=2021-08-05

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -571,7 +571,7 @@ function formatSearchCriteriaToQueryFilters(criteria) {
         delete parsedCriteria.eligibility;
     }
     filters = { ...filters, ...parsedCriteria };
-    validateSearchFilters(filters);
+
     return filters;
 }
 
--- a/packages/server/src/db/index.js
+++ b/packages/server/src/db/index.js
@@ -540,6 +540,41 @@ function grantsQuery(queryBuilder, filters, agencyId, orderingParams, pagination
     }
 }
 
+function formatSearchCriteriaToQueryFilters(criteria) {
+    const parsedCriteria = JSON.parse(criteria);
+    const postedWithinOptions = {
+        'All Time': 0, 'One Week': 7, '30 Days': 30, '60 Days': 60,
+    };
+    let filters = {};
+    if (parsedCriteria.includeKeywords) {
+        filters.includeKeywords = parsedCriteria.includeKeywords.split(',').map((s) => s.trim());
+        delete parsedCriteria.includeKeywords;
+    }
+    if (parsedCriteria.excludeKeywords) {
+        filters.excludeKeywords = parsedCriteria.excludeKeywords.split(',').map((s) => s.trim());
+        delete parsedCriteria.excludeKeywords;
+    }
+    if (parsedCriteria.fundingTypes) {
+        filters.fundingTypes = parsedCriteria.fundingTypes.map((ft) => ft.code);
+        delete parsedCriteria.fundingTypes;
+    }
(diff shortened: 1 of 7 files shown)


as1729:
On nested if-statements- totally agree w/ CodeClimate that it needs to be fixed. However, given the time constraint I'm not going to make the readability changes. In order to ensure the code works, I added thorough unit-test coverage which should help in future refactoring efforts.

packages/server/src/db/index.js
```diff
@@ -540,8 +540,120 @@
     }
 }
 
+function formatSearchCriteriaToQueryFilters(criteria) {
```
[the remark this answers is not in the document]
as1729: wont fix

packages/server/src/db/index.js
```diff
@@ -540,8 +540,120 @@
...
+    if (parsedCriteria.eligibility) {
+        filters.eligibilityCodes = parsedCriteria.eligibility.map((e) => e.code);
+        delete parsedCriteria.eligibility;
+    }
+    filters = { ...filters, ...parsedCriteria };
+    validateSearchFilters(filters);
+    return filters;
+}
+
+function validateSearchFilters(filters) {
```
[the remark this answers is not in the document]
as1729: wont fix

packages/server/src/db/index.js
```diff
@@ -540,8 +540,120 @@ function grantsQuery(queryBuilder, filters, agencyId, orderingParams, pagination
...
+    if (parsedCriteria.postedWithin) {
+        filters.postedWithinDays = postedWithinOptions[parsedCriteria.postedWithin] || 0;
+        delete parsedCriteria.postedWithin;
+    }
+    if (parsedCriteria.eligibility) {
+        filters.eligibilityCodes = parsedCriteria.eligibility.map((e) => e.code);
+        delete parsedCriteria.eligibility;
+    }
+    filters = { ...filters, ...parsedCriteria };
+    validateSearchFilters(filters);
```
replicantSocks, line 574: Is this worth calling here? We're not looking at the return and it's called again when we get the grants

packages/server/src/db/index.js
```diff
@@ -540,8 +540,120 @@ function grantsQuery(queryBuilder, filters, agencyId, orderingParams, pagination
     }
 }
 
+function formatSearchCriteriaToQueryFilters(criteria) {
```
replicantSocks, line 543: Honest question: Would it be better to convert the search calls to fit the serialized db representation rather than change the db representation to fit the search query? I'm trying to think about when input rules and fields change, which is likely on the client, but might be an afterthought for the notifications.

packages/server/src/db/index.js
as1729: That's a good point. I'll remove it. When I put this in, I mistakenly thought the error was being thrown.

as1729: Fixed 👍

replicantSocks (approved):
Seems ok for now

PR #1979, opened 2023-09-27
Hide dashboard stats, skip separate query for total count

jakekreider:
## Description
1. Dashboard stats are hidden with the same feature flag that enables new table, let me know if it should be removed altogether.
2. I went to remove the total count and page totals, but I couldn't not set `total-rows` and still have the next & previous buttons work. Instead, I took a stab at adding total to the main query so the second one doesn't need to run. It works in my testing and `EXPLAIN ANALYZE` on the query shows the same cost as before. But let me know if I'm missing something fundamental here & I'll look for a different approach. 

### Before - Paginator - shows 5224 results
image

### After - Paginator - shows 5224 results (expected)
image

### Before - Dashboard - Notice the `/dashboard` request
image

### After - Dashboard - Notice no `/dashboard` request
image

## Testing
1. Confirm dashboard stats don't display when new Grants table is on
2. Try different filters on grants search, total count should be the same as before. 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/client/src/views/Dashboard.vue
```diff
@@ -104,7 +104,7 @@
     </b-row>
     <br v-if="!showStats" />
 
-    <b-card title="Total Interested Grants by Agencies">
+    <b-card title="Total Interested Grants by Agencies" v-if="showStats" >
```
as1729, line 107: Does hiding this here also eliminate making a request to the backend on page load?

as1729: @jakekreider I updated the code a bit to ensure we do not load this data from the backend as well.

jakekreider: Good catch, thanks! And I appreciate the screen shots as well.

as1729:
@jakekreider I was able to run this code locally and there are no issues. I am getting information as expected. I QA'd both the paginator and the old/new dashboard and both are working correctly.

Since I have also committed to this PR, I'll wait for @TylerHendrickson 's approval before merging

TylerHendrickson (approved):

PR #1992, opened 2023-09-28
Add task queue for ARPA audit reports

TylerHendrickson:
## Description

This PR implements an SQS-based task queue for generating and sending ARPA audit reports. Similar to the way we consume messages from `grants-ingest`, the mechanics are:
- A dedicated SQS queue exists for storing messages that represent a request for audit report generation
- An existing route handler for `GET /api/audit-report` requests publishes to the SQS queue when the endpoint receives a request with `?queue=true`. 
- When the queue contains messages, ECS Fargate tasks are created according to the number of messages in the queue.
- When started, each ECS task performs the following activities:
    1. Receive up to 1 message from the queue
    2. Decode the message's JSON body, and call `generateAndSendEmail()` based on its contents
    3. If `generateAndSendEmail()` throws an exception (or if the entire process crashes), nothing happens – after a timeout period configured on the SQS queue, the message will be returned to the queue for another receive-and-process attempt.
    4. If `generateAndsendEmail()` returns without throwing an exception, the message is considered "done" and is deleted from the queue.
    5. If a `SIGINT` or `SIGTERM` signal was received while during handling of the previous message, the task shuts down. Otherwise, it goes back to Step 1 to potentially handle additional messages in the queue.
- A second queue is provisioned to serve as a dead-letter queue (DLQ). Messages that were received twice without being deleted are sent to the DLQ for inspection/developer intervention.

### Refactor Call-Out

One particular commit in this branch, a492c06af77b415216d0129a1765668b1a6d04af, touches a fair number of files and functions throughout the `packages/server/arpa_reporter` code. The motivation behind these changes is to ensure that the `packages/server/arpa_reporter/lib/audit-report.js` module (in particular, the call-path starting at `generateAndSendEmail()`) is usable in contexts that are not triggered by an (Express) request.

All of the functions modified in this commit previously invoked at least one of the `useRequest()`, `useUser()`, `useTenantId()` functions provided by `packages/server/src/arpa_reporter/use-request.js`, either directly (i.e. the now-modified function called one of those somewhere in its body) or indirectly (i.e. it called a function that (called a function that...) called one of those). Since the `use-request.js` module expects the availability of data specific to an HTTP request, this causes any usages to be incompatible with any runtime scenario where this data is unavailable. The problem with relying on these helpers is that it makes the code less portable/reusable, especially in cases like these. 

In order to maximize compatibility with existing code and to keep the modifications as minimal as possible, I followed these patterns while refactoring the `generateAndSendEmail()` call-path:
1. Any function which directly called a `use-request.js` function now makes that call as a default function parameter. For example:
    ```js
    // Before:
    function myFunction(something, trns = knex) {
        const tenantId = useTenantId();
        // code that relies on tenantId follows...
    }
    // After:
    function myFunction(something, trns = knex, tenantId = useTenantId()) {
        // code that relies on tenantId follows...
    }
    ```
    This allows unmodified, previously-compatible function calls to omit the new argument without any changes to behavior, while allowing new, non-request-based calls to provide the value explicitly.
2. Any function which called one of the functions affected by the "Round 1" modifications, but did not directly call a `use-request.js` function itself, now accepts a corresponding (optional) parameter. For example:
    ```js
    // Before:
    function myFunction(something, trns = knex) {
        const result = someFunctionThatNeedsATenantId();
        // code that relies on result follows...
    }
    // After:
    function myFunction(something, trns = knex, tenantId) {
        const result = someFunctionThatNeedsATenantId(tenantId);
        // code that relies on result follows...
    }
    ```
3. After applying the "Round 2" modifications, rinse and repeat for all functions in the call-path.

Altogether, a simplified representation of the changes could be:
- Before:
    ```js
    function getTenant(trns = knex) {
        const tenantId = useTenantId();
        return await trns('tenants').where({ tenant_id: tenantId }).first();
    }
    
    function tenantDisplayName() {
        const tenant = await getTenant();
        return tenant.display_name;
    }
    
    function greetUser() {
        const user = useUser();
        const tenantName = tenantDisplayName();
        return `Hi, ${user.name}! You are a member of ${tenantName}.`
    }
    
    router.get('/greeting', requireUser, async (req, res) => {
        const greeting = greetUser();
        res.json({ greeting });
        return;
    });
    
    function notInRequest(userId) {
        try {
            const greeting = greetUser();
            console.log('${greeting} 😁');
        } catch(e) {
            console.log("Cannot greet ${userId} because I don't know who they are 😢");
        }
    }
    ```
    - Requests to `GET /greeting` with an authenticated user will succeed.
    - Calling `notInRequest()` will always result in the `"Cannot greet ... 😢"` message.
- After:
    ```js
    function getTenant(trns = knex, tenantId = useTenantId()) {
        return await trns('tenants').where({ tenant_id: tenantId }).first();
    }
    
    function tenantDisplayName(tenantId) {
        const tenant = await getTenant(undefined, tenantId);
        return tenant.display_name;
    }
    
    function greetUser(user = useUser(), tenantId) {
        const tenantName = tenantDisplayName(tenantId);
        return `Hi, ${user.name}! You are a member of ${tenantName}.`
    }
    
    router.get('/greeting', requireUser, async (req, res) => {
        const greeting = greetUser();
        res.json({ greeting });
        return;
    });
    
    function notInRequest(userId) {
        try {
            const user = getUser(userId);
            const greeting = greetUser(user);
            console.log('${greeting} 😁');
        } catch(e) {
            console.log("Cannot greet ${userId} because I don't know who they are 😢");
        }
    }
    ```
    - Although the handler is unmodified, authenticated requests to `GET /greeting` will behave the same as before.
    - As long as `notInRequest()` is able to call `getUser(userId)`, it is now able to log the `"Hi, ... 😁"` message.

### Monitoring

This PR includes configuration for a Datadog monitor that alerts after two failed attempts to generate an audit report. Along with the DLQ, this monitor will be especially useful in two ways:
1. We will be able to know when an audit report has failed, instead of needing partners to tell when they haven't received an email with the report.
2. After investigation and making code changes, we'll have the option to [redrive](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-dead-letter-queue-redrive.html) failed messages in order to retry report generation without needing to ask partners to try again.

### Auto-scaling behavior
Initial auto-scaling behaviors have been configured based on the assumption that audit report requests are somewhat infrequent, but it is possible that multiple requests could be received at around the same time. With that in mind, the number of running tasks correspond to the queue's message count as follows:
| **Min Messages** | **Max Messages** | **Task Count** |
|---|---|---|
| 0 | 0 | _Not Running_ |
| 1 | 1 | 1 |
| 2 | 3 | 2 |
| 4 | 5 | 3 |
| 6 | 10 | 4 |
| 11 | 20 | 5 |
| 21 | 50 | 6 |
| 51 | _Infinite_ | 7 |

## Testing

Note: functional testing requires LocalStack

1. Create an SQS queue for testing
    ```cli
    $ awslocal sqs create-queue --queue-name arpa-audit-reports
    {
        "QueueUrl": "http://localhost:4566/000000000000/arpa-audit-reports"
    }
    ```
2. Send a message to the queue (replace the example user ID with a fixture user whose tenant can generate a report)
    ```cli
    $ awslocal sqs send-message \
        --queue-url 'http://localhost:4566/000000000000/arpa-audit-reports'
        --message-body '{"userId": 1234}'
    ```
3. Export the following environment variables required to run the script, changing values as-needed to suit your environment:
    ```
    ENABLE_GRANTS_DIGEST=false 
    ENABLE_GRANTS_SCRAPER=false 
    DATA_DIR=/path/to/data/volume/mount 
    NOTIFICATIONS_EMAIL="foo@example.org"
    WEBSITE_URL="http://example.org"
    AWS_REGION=us-west-2
    AWS_DEFAULT_REGION=us-west-2 
    AWS_ACCESS_KEY_ID=test 
    AWS_SECRET_ACCESS_KEY=test
    TASK_QUEUE_URL=http://localhost:4566/000000000000/arpa-audit-reports
    ```
5. Run the script to receive the message and generate the report (note: this may need adjusting to be run in/with `docker compose`)
    ```cli
    node packages/server/scripts/arpaAuditReport.js
    ```

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Provided adequate test coverage for all new code – given urgency and potential need for iteration, tests will be added in a follow-up PR
- [x] Added PR reviewers


packages/client/src/arpa_reporter/views/Home.vue
```diff
@@ -111,7 +111,7 @@ export default {
       this.sending = true;
 
       try {
-        const result = await getJson('/api/audit_report?async=true');
+        const result = await getJson('/api/audit_report?queue=true');
```
TylerHendrickson, line 114: Note: The button is still disabled.

as1729 (approved):

PR #2023, opened 2023-10-03
feat: adds infra required to support digest emails

as1729:
## Description
Implements the infrastructure pieces within this design. The business logic migration will come separately.
image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  # Task resource configuration
+  # TODO: Tune these values after observing usage in different environments.
+  #       See also: --max_old_space_size in NODE_OPTIONS env var.
+  consumer_task_size = {
+    cpu    = 1024 # 1 vCPU
+    memory = 4096 # 4 GB
+  }
+
+  # Messaging
+  autoscaling_message_thresholds = [1, 3, 5, 10, 20, 50]
```
TylerHendrickson, line 440: Unlike the `arpa_audit_report` task, where simultaneous requests for reports are rare and we need to generate them in an hour or less, we can probably assume this task's source queue will generally have a high starting message count (whenever it isn't empty) and doesn't need to work through the queue as fast as possible. These thresholds set the number of tasks based on the number of messages in the source queue, so low values like `1, 3, 5` aren't worth setting in this particular use-case. Individual jobs (per-message) will also probably finish pretty swiftly, so I would tune this more along the values we expect to see (i.e. the number of unique search criteria).

Something like this is probably a decent starting point:
```suggestion
  autoscaling_message_thresholds = [200, 500, 1000, 2000, 5000, 10000]
```

NB: That :point_up: is 1 task at 0-200 messages, working up to 7 tasks at anything > 10k messages.

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+    NOTIFICATIONS_EMAIL = "grants-notifications@${var.website_domain_name}"
+    WEBSITE_DOMAIN      = "https://${var.website_domain_name}"
+  }
+  consumer_task_efs_volume_mounts = [{
+    name            = "data"
+    container_path  = "/var/data"
+    read_only       = false
+    file_system_id  = module.api.efs_data_volume_id
+    access_point_id = module.api.efs_data_volume_access_point_id
+  }]
```
TylerHendrickson, line 430: I don't believe this task (or any of the others) requires access to any files in the `/var/data` directory, so this can be removed.
```suggestion
```

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  # Task resource configuration
+  # TODO: Tune these values after observing usage in different environments.
+  #       See also: --max_old_space_size in NODE_OPTIONS env var.
+  consumer_task_size = {
+    cpu    = 1024 # 1 vCPU
+    memory = 4096 # 4 GB
+  }
```
TylerHendrickson, line 437: (See my note above on line 420)

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+    NOTIFICATIONS_EMAIL = "grants-notifications@${var.website_domain_name}"
+    WEBSITE_DOMAIN      = "https://${var.website_domain_name}"
+  }
+  consumer_task_efs_volume_mounts = [{
+    name            = "data"
+    container_path  = "/var/data"
+    read_only       = false
+    file_system_id  = module.api.efs_data_volume_id
+    access_point_id = module.api.efs_data_volume_access_point_id
+  }]
```
TylerHendrickson, line 505: Same as with the previous task – I think this can be removed.
```suggestion
```

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  # Task resource configuration
+  # TODO: Tune these values after observing usage in different environments.
+  #       See also: --max_old_space_size in NODE_OPTIONS env var.
+  consumer_task_size = {
+    cpu    = 1024 # 1 vCPU
+    memory = 4096 # 4 GB
+  }
+
+  # Messaging
+  autoscaling_message_thresholds = [1, 3, 5, 10, 20, 50]
```
TylerHendrickson, line 519: This ultimately depends on anticipated number of individual emails that will get queued up; try this as a starting point:

```suggestion
  autoscaling_message_thresholds = [100, 200, 400, 600, 800, 1000, 5000, 10000]
```

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  sqs_publisher = {
+    principal_type       = "Service"
+    principal_identifier = "ecs-tasks.amazonaws.com"
+    source_arn           = module.api.ecs_service_arn
+  }
+  sqs_max_receive_count             = 2
```
TylerHendrickson, line 525: Might be worth increasing this depending on the max messages that are fetched at once, and how likely any of those jobs would be to fail.

Bear in mind that it will be really important that a message's job not fail at any point _after_ an email gets sent (otherwise a retried message will send a duplicate). Initially, it might be best to set this value to `1` (i.e. no retries / DLQ after the first failure) on the basis of:
1. A single missed email isn't likely to be a huge problem.
2. There's a DLQ, so any failures could always be retried.
3. Assuming it's better for someone to _not_ receive an email than to potentially receive duplicate emails.
4. At this point, there's little evidence to suggest that a failure would be ephemeral and that an initially-failed message would be successful upon retry.

terraform/main.tf
```diff
@@ -301,6 +332,213 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+    send-emails = module.api.send_emails_policy_json
+  }
+
+  # Task resource configuration
+  # TODO: Tune these values after observing usage in different environments.
+  #       See also: --max_old_space_size in NODE_OPTIONS env var.
+  consumer_task_size = {
+    cpu    = 1024 # 1 vCPU
+    memory = 4096 # 4 GB
+  }
```
TylerHendrickson, line 516: I'm less confident about where these values will ultimately land given that we know that `mustache` doesn't exactly have a low resource footprint. That said, we've been sending messages successfully up to this point on API tasks with far less allocated resources. Unless you're planning on building and sending multiple emails in-parallel within a single task (e.g. pull down 5 messages from the queue and process them in an async batch), I'm not sure there's evidence that starting with more resources will make individual emails get sent out any faster.

Suggestion:
```suggestion
  consumer_task_size = {
    cpu    = 256 # .25 vCPU
    memory = 512 # MB
  }
```

Correspondingly, the `--max_old_space_size` value should be adjusted to something like `450`.

TylerHendrickson (changes requested):

terraform/main.tf
as1729: I initially removed it, but ran into a terraform validation error as it seemed to be expecting this volume mount to exist. Let me try again.

https://github.com/usdigitalresponse/usdr-gost/actions/runs/6397702168/job/17366227447

```
Error: Missing required argument

  on main.tf line 471, in module "digest_email_send":
 471: module "digest_email_send" {

The argument "consumer_task_efs_volume_mounts" is required, but no definition
was found.
Error: Terraform exited with code 1.
Error: Process completed with exit code 1.
```

TylerHendrickson: Ah, good find. We discussed this in our call today, but just to state for the record: you're running into this because of the [variable definition in the module](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/terraform/modules/sqs_consumer_task/variables.tf#L38-L47), here:
```tf
variable "consumer_task_efs_volume_mounts" {
  description = "EFS volumes to mount in ECS task consumer containers."
  type = list(object({
    name            = string
    container_path  = string
    read_only       = bool
    file_system_id  = string
    access_point_id = string
  }))
}
```

The module uses this variable in two separate `for` loops to configure a separate volume mount for each list item provided to this variable when "calling" the module. Moreover, Terraform sees this particular variable as a required input because it doesn't have a `default` value assignment. 

So, there are two ways to fix:
1. When calling the module, provide an empty list as the value for this input variable, e.g.
    ```tf
    consumer_task_efs_volume_mounts = []
    ```
2. Set a `default` value of an empty list in the variable definition lined above:
    ```tf
    variable "consumer_task_efs_volume_mounts" {
      description = "EFS volumes to mount in ECS task consumer containers."
      type = list(object({
        name            = string
        container_path  = string
        read_only       = bool
        file_system_id  = string
        access_point_id = string
      }))
      default = []
    }
    ```

Given that an empty list works just fine, I don't think it should be a requirement in the first place, so I'd advocate for the second option.

dysmento:
@as1729 asked if I could help out on this one. I implemented as many of @TylerHendrickson's suggestions as I could, but there's a little bit still not working. `terraform plan` has two complaints about `ecs_task_role_name` not being a real attribute.

```
│ Error: Unsupported attribute
│ 
│   on main.tf line 423, in resource "aws_iam_role_policy" "digest_email_get_grants-publish_to_digest_email_send_queue":
│  423:   role        = module.digest_email_get_grants.ecs_task_role_name
│     ├────────────────
│     │ module.digest_email_get_grants is a object
│ 
│ This object does not have an attribute named "ecs_task_role_name".
╵
╷
│ Error: Unsupported attribute
│ 
│   on modules/gost_api/outputs.tf line 56, in output "digest_email_kickoff_cron_ecs_task_role_name":
│   56:   value = module.digest_email_kickoff_cron.ecs_task_role_name
│     ├────────────────
│     │ module.digest_email_kickoff_cron is a object
│ 
│ This object does not have an attribute named "ecs_task_role_name".
```

And the other thing is I didn't know which scripts to move out of `packages/server/lib/digest`

terraform/main.tf
```diff
@@ -320,6 +341,148 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  statement {
+    sid       = "AllowPublishToQueue"
+    actions   = ["sqs:SendMessage"]
+    resources = [module.digest_email_send.sqs_queue_arn]
+  }
+}
+
+resource "aws_iam_role_policy" "digest_email_get_grants-publish_to_digest_email_send_queue" {
+  name_prefix = "send-digest-email-send-requests"
+  role        = module.digest_email_get_grants.ecs_task_role_name
```
TylerHendrickson, line 423:
```suggestion
  role        = module.digest_email_get_grants.iam_task_role_name
```

terraform/main.tf
```diff
@@ -320,6 +341,148 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_audit_report_queue" {
...
+  statement {
+    sid       = "AllowPublishToQueue"
+    actions   = ["sqs:SendMessage"]
+    resources = [module.digest_email_get_grants.sqs_queue_arn]
+  }
+}
+
+resource "aws_iam_role_policy" "digest_email_kickoff_cron-publish_to_digest_email_get_grants_queue" {
+  name_prefix = "send-digest-email-get-grants-requests"
+  role        = module.api.digest_email_kickoff_cron_ecs_task_role_name
```
TylerHendrickson, line 354: Per item 2b in our [Slack discussion](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1699579161642309?thread_ts=1699568729.822169&cid=C0324KDQSCR):

```suggestion
  role        = module.api.ecs_task_role_name
```

terraform/modules/gost_api/outputs.tf
```diff
@@ -47,3 +47,11 @@ output "efs_data_volume_access_point_id" {
...
   value = module.send_emails_policy.json
 }
+
+output "digest_email_kickoff_cron_security_group_id" {
+  value = module.digest_email_kickoff_cron_security_group.id
+}
+
+output "digest_email_kickoff_cron_ecs_task_role_name" {
+  value = module.digest_email_kickoff_cron.ecs_task_role_name
+}
```
TylerHendrickson, line 57: Per item 2a in our [Slack discussion](https://usdigitalresponse.slack.com/archives/C0324KDQSCR/p1699579161642309?thread_ts=1699568729.822169&cid=C0324KDQSCR):
```suggestion
output "digest_email_kickoff_cron_ecs_task_role_name" {
  value = join("", aws_iam_role.task.*.name)
}
```

Or, per item 2b (which I'm leaning toward, personally):
```suggestion
```

terraform/modules/sqs_consumer_task/variables.tf
```diff
@@ -44,6 +44,7 @@ variable "consumer_task_efs_volume_mounts" {
     file_system_id  = string
     access_point_id = string
   }))
+  default = []
```
TylerHendrickson, line 47: 🙏

TylerHendrickson (commented):
@dysmento Dropped a few suggestions in based on today's Slack thread.

as1729:
Closing as @jeffsmohan is already working on resolving the issue in separate PRs.

PR #2033, opened 2023-10-05
Revert "fix: prevent changing search when results are loading"

as1729:
Reverts usdigitalresponse/usdr-gost#2025


as1729:
@replicantSocks the reason for revert is-
- "Edit" and "Clear" button does not work when a saved search is selected. User can click it but nothing ends up happening.

TylerHendrickson (approved):

PR #2060, opened 2023-10-11
ARPA audit report tuning: Round 2

TylerHendrickson:
## Description

This PR provides further performance/reliability optimizations (and a few other tweaks) for the ARPA audit report. 

Summary of changes:
- Refactor the `createProjectSummaries()` function to only use a single Postgres query in order to build the report.
  - Although the number of queries emitted over building this spreadsheet were not contributing to performance issues on the Postgres side, the number of queries executed in-parallel were maxing out the available number of (local) connections maintained by Knex. 
  - The previous query patterns were to execute several deeply-nested queries:
    - To find the current reporting period
    - To find all previous (non-current) reporting periods
    - For each identified reporting period, at least one query to find all uploads associated with the reporting period
  - The unbounded (especially when concurrent) querying behavior was causing certain generate-and-send processes to fail in Production. Now, only a single query is issued when generating this spreadsheet.
  - **Note:** the number of rows potentially fetched by this query is (still) unbounded. While not currently an issue, we can expect this number to grow per-tenant over time. Further work is warranted to pre-compute the results of this query at the point when a new file contributing to the audit report is uploaded.
  - This query is similar to that which is produced by the `usedForTreasuryExport()` helper function, although it contains a few optimizations for limiting both the number of returned rows and for limiting the virtual table size for inner-join operations.
  - In addition to reducing the number of Postgres queries, the new query pattern yields additional memory optimizations for the `Project Summaries` sheet generation process. 
    - Although other recent PRs have introduced concurrency limits to control sudden memory usage spikes, the total amount of memory required by the `createProjectSummaries()` was still dependent on being able to load all records from all uploads in all reporting periods.
    - With these changes, we only load records for a single upload at a time, and the only record-derived data retained across the sheet-building process are the values used in the final rows. Therefore, the amount of memory required at any given moment during the function call is more-closely aligned to the aggregate row data and/or the number of records in a single upload. This is still unbounded (and the size of a single upload and/or the total amount of project data in a`Project Summaries` sheet may very well still cause problems down the line), but this buys us some runway by avoiding the need to hold multiple uploaded spreadsheets in memory at once.
- Update the behavior of sending SES emails through the AWS transport to throw errors instead of just logging them. This is necessary to ensure that workflows which cannot be considered successful (such as the ARPA report-generation process) unless an email is sent. The previous "fire-and-forget" async behavior caused the generate-and-send process to "finish" before the `sendEmail()` call had returned.
- Updated various unit tests with hard-coded `localhost` expectations in the ARPA suite to instead use the configured `ARPA_REPORTER_BASE_URL` value.

**Next steps**
1. Once these changes have been tested and confirmed as working successfully, we should be able to re-enable the ability for partners to request ARPA audit report emails.
2. The optimizations in this PR should be applicable to other sheet-generating functions in the ARPA audit report workflow. We should prioritize which of those warrant similar optimization efforts and follow up accordingly.
3. As mentioned above, a better long-term solution for limiting resource requirements (and relatedly, keeping email-sending times within the 1-hour limit) lies in a strategy that maintains precomputed, running aggregates values for these spreadsheets. We should investigate and discuss whether it's worth exploring those sooner than we might be able to deliver a modular reporting tool.

## Testing

- Follow the testing path outlined in the QA script for uploading files and requesting an ARPA audit report.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/lib/grants-ingest.test.js
+++ b/packages/server/__tests__/lib/grants-ingest.test.js
@@ -12,9 +12,12 @@ describe('processMessages', async () => {
 
     const serlializeGrantEvent = (newData = {}, prevData = {}) => {
         const eventData = { detail: { versions: { new: newData, previous: prevData } } };
-        if (newData !== {} && prevData !== {}) {
+        const newDataJSON = JSON.stringify(newData);
+        const prevDataJSON = JSON.stringify(prevData);
+        const emptyObjectJSON = JSON.stringify({});
+        if (newDataJSON !== emptyObjectJSON && prevDataJSON !== emptyObjectJSON) {
             eventData.detail.type = 'update';
-        } else if (newData !== {}) {
+        } else if (newDataJSON !== emptyObjectJSON) {
             eventData.detail.type = 'create';
         } else {
             eventData.detail.type = 'delete';
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -8,14 +8,16 @@ const { log } = require('../../lib/logging');
 const aws = require('../../lib/gost-aws');
 const { ec } = require('./format');
 
-const { getPreviousReportingPeriods, getReportingPeriod, getAllReportingPeriods } = require('../db/reporting-periods');
+const { getPreviousReportingPeriods, getAllReportingPeriods } = require('../db/reporting-periods');
 const { getCurrentReportingPeriodID } = require('../db/settings');
-const { recordsForProject, recordsForReportingPeriod, mostRecentProjectRecords } = require('../services/records');
+const {
+    recordsForProject, recordsForReportingPeriod, recordsForUpload, EC_SHEET_TYPES,
+} = require('../services/records');
 const { usedForTreasuryExport } = require('../db/uploads');
 const { ARPA_REPORTER_BASE_URL } = require('../environment');
 const email = require('../../lib/email');
 const { useTenantId } = require('../use-request');
-const { getUser } = require('../../db');
+const { getUser, knex } = require('../../db');
 
 const REPORTING_DATE_FORMAT = 'MM-DD-yyyy';
 const REPORTING_DATE_REGEX = /^(\d{2}-\d{2}-\d{4}) /;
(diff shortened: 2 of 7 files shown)


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -139,47 +141,93 @@ async function createObligationSheet(periodId, domain, tenantId, logger = log) {
...
-                },
-                project: {
-                    id: record.content.Project_Identification_Number__c,
-                },
-            },
-        });
-        recordLogger.debug('populating row from record');
-        const reportingPeriod = await getReportingPeriod(
-            record.upload.reporting_period_id, undefined, tenantId,
+    const uploads = await knex('uploads')
```
tzinckgraf, line 144: Can we move this to a function inside of `db`? Not sure if we will use it anywhere else, but it does help keep the functions in here smaller.

packages/server/__tests__/lib/grants-ingest.test.js
```diff
@@ -12,9 +12,12 @@ describe('processMessages', async () => {
 
     const serlializeGrantEvent = (newData = {}, prevData = {}) => {
         const eventData = { detail: { versions: { new: newData, previous: prevData } } };
-        if (newData !== {} && prevData !== {}) {
+        const newDataJSON = JSON.stringify(newData);
```
tzinckgraf, line 15: I know this is a minor change, but we might want to leave it out of an ARPA specific PR.

packages/server/src/lib/gost-aws.js
```diff
@@ -85,9 +85,13 @@ function sendEmail(message) {
         },
     };
     const command = new SendEmailCommand(params);
-    transport.send(command)
-        .then((data) => console.log('Success sending SES email:', JSON.stringify(data)))
-        .catch((err) => console.error('Error sending SES email:', err, err.stack));
+    try {
+        await transport.send(command)
+            .then((data) => console.log('Success sending SES email:', JSON.stringify(data)));
```
tzinckgraf, line 90: This code will work, but is there a reason to mix the `await` and `.then` paradigm for async?

packages/server/src/arpa_reporter/lib/audit-report.js
TylerHendrickson: For the moment, my intent was to explicitly avoid query reuse (at the expense of larger LoC in the function) until we can see how applicable this query pattern is for the other sheets in the audit report workbook (or similar workflows).

I have no issues ultimately moving/repurposing/refactoring this if it ends up being more broadly usable, but one of the issues I'd found was contributing to technical debt around this problem was the prevalence of query functions being reused and expanded over time (as opposed to, say, writing a more performant, but single-use query) and I'd like to avoid that until we have a more holistic strategy for getting what we need from the DB in these workflows.

tzinckgraf: Makes sense. Thank you for the explanation.

packages/server/__tests__/lib/grants-ingest.test.js
TylerHendrickson: Agree in premise with avoiding scope-creep in PRs, but I'm generally fine with opportunistic unit test improvements as a result of someone being "in the neighborhood".

I'd have more of an issue if:
1. An out-of-scope change ends up generating a lot of discussion that holds up the delivery of everything else – certainly worth breaking out the change into a separate PR in that case.
2. There's an explosion of scope related to actual functional (runtime) code that makes it difficult to evaluate and/or test changes because so much is going on.

If there's an issue with these particular test updates that you think warrants more discussion  to discuss and/or think would benefit from scrutiny in a more dedicated PR, I can remove and submit separately, but if not I think I'm okay leaving as-is.

packages/server/src/lib/gost-aws.js
TylerHendrickson: Fixed.

as1729 (approved):

PR #2062, opened 2023-10-11
Fix: Set consistent arch for postgres to avoid Docker issue on Apple Silicon Macs

adele-usdr:
Fix: Workaround for issue where Docker can't find the right image for an Apple Silicon Mac. Making this change will provide consistency for all contributors, ensuring everyone is working against the same emulated environment.


TylerHendrickson (approved):
💯

Issue #1913, opened 2023-09-13
My Profile Page - Backend Notification Support


kevin-kasunic:

thelmaboamah:
Requirements:
- Migration to support addition notification type "GRANT_FINDER_UPDATES" (reasonable name for Occasional updates?)
- Include GRANT_FINDER_UPDATES in notificationType array in email/constants.js, it's "SUBSCRIBED" by default.
- API put request already iterates through above array to insert values so I don't think there are any changes to make there, but not sure yet
- Is a migration needed give existing users this notification type?
- Update users and db tests

@as1729 @TylerHendrickson could I please get some feedback on this loose plan? I'm pretty new to database migrations so I'll probably need to reach out for some help.

as1729:
> Migration to support addition notification type "GRANT_FINDER_UPDATES" (reasonable name for Occasional updates?)
Include GRANT_FINDER_UPDATES in notificationType array in email/constants.js, it's "SUBSCRIBED" by default.

This name sounds good to me 👍 

> API put request already iterates through above array to insert values so I don't think there are any changes to make there, but not sure yet

There might be some unit tests that need to be updated but otherwise that should work right off the bat.

> Is a migration needed give existing users this notification type?

Currently we treat a `null` value as subscribed so we can keep that precedence. That means only for new user-creation we need to set this value as `subscribed` for existing users we can leave them as `null`. On second thought- it might be better to write a script that will update all users and add this value. Once the script is added to the `/scripts` directly either Ty or I can run it when it is shipped.

thelmaboamah:
got it! thanks @as1729!

ClaireValdivia:
@thelmaboamah did want to check in on this one - we would like the default for "Occasional Updates" to be "on" for all current and new users - would that be helpful to wrap into this issue or do you think a separate issue would be best?

ClaireValdivia:
@as1729 thoughts on how to test this on in staging?

as1729:
@ClaireValdivia seems to be working for me in staging in the my-profile url: https://staging.grants.usdr.dev/#/my-profile

<img width="592" alt="image" src="https://github.com/usdigitalresponse/usdr-gost/assets/6497366/e0ea4f91-5eca-4a59-86a7-430e672a2417">

ClaireValdivia:
[...]

Issue #1911, opened 2023-09-13
My Profile Page - Editable Name


kevin-kasunic:
Make user full name editable. Create Edit Profile modal. Add edit button to trigger modal. Support /users/:userId partial update on the backend. 

Depends on issues #1908 and #1910

PR #2070, opened 2023-10-11
Feature: Edit name on My Profile page

thelmaboamah:
## Description
Gives a user the ability to update their name from the My Profile page. Clicking or tabbing to the edit button opens a modal with a field for updating their name. This PR implements the modal, and a new users routes, and connects the two via the Vuex store.

I added the markup for the Email Notifications section of the page to get a jump on ticket #1913.

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/17172664/0af6eafa-23d5-4b49-b78e-f47618f29693

#### With Email Notification Markup
Screenshot 2023-10-14 at 11 03 01 AM

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/client/src/components/Modals/EditUser.vue
```diff
@@ -0,0 +1,106 @@
...
+
+export default {
+  data() {
+    return {
+      formData: {
+        name: null,
+      },
+    };
+  },
+  validations: {
```
thelmaboamah, line 46: I grabbed these validation values from the AddUser modal

packages/client/src/store/modules/users.js
```diff
@@ -61,6 +61,11 @@ export default {
       await fetchApi.post('/api/organizations/:organizationId/users', user);
       await dispatch('fetchUsers');
     },
+    async updateUser({ commit }, user) {
+      await fetchApi.put(`/api/organizations/:organizationId/users/${user.id}`, user);
+      const data = await fetchApi.get('api/sessions');
```
thelmaboamah, line 66: I'm not sure about this... I am calling api/sessions since the user object the put is returning at the moment is incomplete. But I can resolve that within the put route by calling the same getUsers function /sessions uses so there's only one network request?

packages/server/src/db/index.js
```diff
@@ -85,6 +85,16 @@ async function createUser(user) {
     };
 }
 
+async function updateUser(user) {
+    const { id, name } = user;
```
thelmaboamah, line 89: only supporting the name update currently

packages/server/src/routes/users.js
```diff
@@ -60,6 +60,19 @@ router.post('/', requireAdminUser, async (req, res, next) => {
     }
 });
 
+router.put('/:userId', requireUser, async (req, res) => {
+    const user = req.body;
```
thelmaboamah, line 64: Should I be doing any checks here?

TylerHendrickson (changes requested):
First, the new functionality looks solid and I like the overall implementation 👍 

Requesting changes based on some of your comments + a few decisions I think are worth hashing out regarding the new `PUT` route on the API side...

In general, when I see a `PUT /<resourceType>/<identifier>` route, I think the indication would be to completely overwrite the identified resource with whatever data is included in the request, assuming the request data can be successfully validated against the `<resourceType>` schema (NB: there's a fair amount of stylistic opinion as to whether nested data/sub-objects of a different resource type should be updated in the same operation, which I don't think we need to get into at this point). Given all that, along with (as you noted) the fact that `db.updateUser()` function only supports updating the `name` field (which is perfectly acceptable, given the scope of this feature), I don't think it's robust enough to warrant a `PUT /user/<id>` route, and may lead to confusion (or at least be somewhat non-obvious) if implemented as-is*. 

A few thoughts on how to move forward:
1. If the route currently at `PUT /api/organizations/:organizationId/users/:userId` was instead nested down a level to `PUT /api/organizations/:organizationId/users/:userId/name`, I think everything else in this PR would be perfectly acceptable from a RESTful design perspective.
2. Changing the route from `PUT` to `PATCH` would also resolve API design concerns, since `PATCH` generally implies "partial update" behavior and should only consider the data provided in the request when making an update. Furthermore, a `PATCH` endpoint that performs a partial update and then responds with the full representation of the updated object also makes sense to me.
3. Expanding the scope of the endpoint (for example, by way of modifying the `db.UpdateUser()` function) to update all (or at least all top-level) attributes of a user would also make the behavior of the endpoint more obvious. However, I'm wary of bogging down this PR with changes that are certainly outside the scope of changes requested by #1911.
4. Acknowledging that there are plenty of non-obvious API design decisions and similarly-limited endpoints already implemented all throughout this codebase, and that there's not a lot of precedent at the moment for "intuitive API design" at the moment, I think it's also arguable that this code is perfectly fine as-is. My only reluctance on this point is that any future changes that may attempt to make this new `PUT` route useful beyond a `name` field update would also require FE updates to avoid causing issues.

Altogether, I think Option 2 would be best for this PR, for the following reasons:
- Doesn't contribute to non-obvious API design issues affecting much of this codebase.
- Is perfectly idiomatic from a RESTful API design perspective.
- Shouldn't hinder any efforts to improve API design in the future.
- Resolves the [issue you noted](https://github.com/usdigitalresponse/usdr-gost/pull/2070/files#r1355376104) regarding multiple network requests, since the `PATCH` endpoint could (should) respond with the full user object.
- Hopefully only involves minimal changes to this PR.

packages/server/src/routes/users.js
TylerHendrickson: Per my broader feedback, if this route was changed to a `PATCH`, I think it would be good to validate that only a `name` attribute is provided in the request body. 

Something that amounts to the following would be great:
```js
const allowedFields = new Set([ 'name' ]);
for (cons key of Object.keys(req.body)) {
    if (!allowedFields.has(key)) {
        res.status(400).json({ message: `Request body contains unsupported field: ${key}` });
        return;
    }
}

const user = req.body;  // Note: no destructuring, as we expect req.body to represent a user object
try {  // proceed as-written...
    const result = await db.updateUser(user); 
```

TylerHendrickson (approved):
Looks great! 🚀

Issue #949, opened 2023-02-13
Recent activity should display scrolling list


chrisxkeith:
**Current State**
The query that backs this table returns every grant that has had any status set on it or has been assigned by any member of the user's team or any sub-team of the user's team. The date displayed is the last time the status was updated or an assignment was made. Despite the "Recent Activity" label, the query actually retrieves all such interactions that occurred since the beginning of time.

The pagination on this page is confusing and it is not clear what it reflects, but it does not work the same as the pagination on My Grants or Browse Grants tab.

**Expected State**
- Pagination on Recent Activity page should reflect how pagination works on Browse Grants tab
![image](https://github.com/usdigitalresponse/usdr-gost/assets/136839582/57ea1677-ba2e-4c54-9000-ab2ac13361ef)

chrisxkeith:
???
https://github.com/usdigitalresponse/usdr-gost/blob/bd0ab6f8f53749c7590c1fd2a61ec3c99a24399c/packages/client/src/views/RecentActivity.vue#L125
https://github.com/usdigitalresponse/usdr-gost/blob/bd0ab6f8f53749c7590c1fd2a61ec3c99a24399c/packages/client/src/store/modules/dashboard.js#L78

ClaireValdivia:
in QA party, still seeing this not working on staging and prod

adele-usdr:
@ClaireValdivia just to make sure I'm looking at the the right issue, does anyone who saw this again have a screenshot to drop in?

ClaireValdivia:
this is from staging, seeing similar behavior on prod
![Screenshot 2023-11-07 at 9 36 12 AM](https://github.com/usdigitalresponse/usdr-gost/assets/136839582/ac4fe066-2e8f-4c25-8809-53ae4221910b)

adele-usdr:
@ClaireValdivia what do you expect each of those numbers to represent?  Right now, the first number is the total interested grants (which probably has more counted in it than shown on the screen), and the second is the total # shown.  I also noticed "x of y"  is a button. Is clicking on it supposed to do something?

ClaireValdivia:
@adele-usdr I think we just want it flipped - i.e. "10 of 30", but if we are going to show it, I think there also needs to be a way to see all 30 - right now there is only the one page with 10 results. "30 of 10" should not be a button - good catch!

[...]

PR #2097, opened 2023-10-16
fix: #949 update interested grants count labels

adele-usdr:
Update totalRows and label to use counts that are filtered or not filtered properly.

#949 Unusual display in "Recent Activity" grid

### DIFF
--- a/packages/client/src/views/RecentActivity.vue
+++ b/packages/client/src/views/RecentActivity.vue
@@ -53,7 +53,7 @@
       <b-pagination class="m-0" v-model="currentPage" :total-rows="totalRows" :per-page="perPage" first-number
         last-number first-text="First" prev-text="Prev" next-text="Next" last-text="Last"
         aria-controls="grants-table" />
-      <b-button class="ml-2" variant="outline-primary disabled">{{ grantsInterested.length }} of {{ totalRows }}</b-button>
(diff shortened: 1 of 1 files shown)


adele-usdr:
I got some local test failures so not sure yet if this is right.

adele-usdr:
I'd also like to figure out how to add tests for this!

TylerHendrickson:
> I'd also like to figure out how to add tests for this!

You and me both! 😆 

Unfortunately, the main `Finder` client-side code doesn't have a lot of coverage in its test suite; at present, the only tests that exist for the client-side are only covering the following parts of the `packages/client` path:
- `src/helpers/fetchApi`
- `src/helpers/featureFlags`
- `src/arpa_reporter`

I don't have a lot of background in maintaining this type of code, so I've been a bit reluctant to dive in on adding unit tests for client-side code without 1) someone on the team who can advise setting up a test suite, or 2) enough capacity to research and set things up accordingly. Given that this is a longstanding shortcoming of this part of the codebase, I don't think adding tests needs to block this PR, but it would be good to get to a point where fixes like these can have accompanying test cases to prevent future regressions.

TylerHendrickson (approved):
These changes look good to me – counts look correct when testing locally 🎉 
I also don't see any failures when running tests on my end.

Issue #1396, opened 2023-05-19
BUG: Sorting grants by "Viewed by" or "Interested Agencies" fails


TylerHendrickson:
**Description:** Users browsing grants on either the "Browse Grants" or "My Grants" tabs in the ID tool have the option of sorting the results table by certain columns. However, when attempting to sort by "Viewed by" or "Interested agencies", requests to `GET /api/organizations/:organizationId/grants` fail with a 500 status code because of an error thrown on the backend due to a malformed Postgres query:
> `select distinct "grants".* from "grants" left join "grants_interested" on "grants"."grant_id" = "grants_interested"."grant_id" where ("eligibility_codes" ~ $1 and "description" ~* $2) order by "grants_interested"."grant_id" asc, "grants"."grant_id" asc limit $3 - for SELECT DISTINCT, ORDER BY expressions must appear in select list`

**Steps to reproduce:**
1. Login to the ID Tool (role doesn't matter)
2. Navigate to the "Browse Grants" tab.
3. Ensure that the browser's developer console is recording requests, and attempt to sort the results based on the "Viewed by" column. Observe that the request fails with a 500 response code and that the order of results in the UI does not change.
4. Attempt to sort the results based on the "Interested Agencies" column. Observe that the request fails with a 500 response code and that the order of results in the UI does not change.
5. Optionally, navigate to the "My Grants" tab and assuming there is a non-empty list of results, repeat steps 4 and 5 on that tab. The observed results should be the same.

**Impacts:**
- The sort feature for both columns is broken for users, although no error is presented.

**Desired behavior:**
- ~~Users are able to sort grants based on the "Viewed by" column.~~
- ~~Users are able to sort grants based on the "Interested agencies" column.~~
- Users don't have the option of sorting by the "Viewed by" and "Interested agencies" fields (i.e. remove this feature).

**Implementation details:**
- UI: Remove `sortable: true` option for the columns keyed by `viewed_by` and `interested_agencies` from the [`packages/client/src/components/GrantsTable.vue` component](https://github.com/usdigitalresponse/usdr-gost/blob/_staging/packages/client/src/components/GrantsTable.vue#LL101C1-L108C11).
- API: 
[...]

PR #2158, opened 2023-10-27
fix: remove sorting grants, Viewed By and Interested Agencies #1396

dysmento:
## Description
Resolves the problem that sorting grants by 'Viewed By' and 'Interested Agencies' throws API errors.

On the frontend, the Viewed By and Interested Agencies columns are no longer sortable.

On the server, tightened up the input validation such that any values for `orderBy` other than `award_ceiling|open_date|close_date` are ignored completely.

## Screenshots / Demo Video
CleanShot 2023-10-27 at 09 37 15@2x

## Testing
Navigate to My Grants and Browse Grants and confirm that sorting is unavailable for Viewed By and Interested Agencies columns.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/components/GrantsTable.vue
+++ b/packages/client/src/components/GrantsTable.vue
@@ -102,11 +102,9 @@ export default {
         },
         {
           key: 'viewed_by',
-          sortable: true,
         },
         {
           key: 'interested_agencies',
-          sortable: true,
         },
         {
           // opportunity_status
--- a/packages/client/src/components/GrantsTableNext.vue
+++ b/packages/client/src/components/GrantsTableNext.vue
@@ -110,11 +110,9 @@ export default {
         },
         {
           key: 'viewed_by',
-          sortable: true,
         },
         {
           key: 'interested_agencies',
-          sortable: true,
         },
         {
           // opportunity_status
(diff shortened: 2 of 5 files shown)


TylerHendrickson (commented):
Overall, I like the changes here – in particular, preventing the inclusion of arbitrary values from the `ordering` request input is a much tighter approach that limits usage scenarios in a way that is less error-prone as well as less of a risk for malicious (or otherwise unintended) use.

Although outside the immediate scope of #1396, I think it might still be justifiable to move the result-ordering parsing and validation logic from the `grantsQuery()` function into `buildOrderingParams()` instead. Currently, the code implies some sensible expectations about result-ordering behavior:
- Only a single ordering criteria can be applied.
- Results are returned in descending order by default, but ascending order may be requested.
- Results may be sorted according to full-text search relevance (`rank`) or one of the `award_ceiling`, `open_date`, `close_date` column values.

Additionally, the `getGrants()`, `grantsQuery()`, and `getGrantsNew()` functions all seem to expect an `orderingParams` argument as an object structured as:
- `orderBy`: one of `rank`, `award_ceiling`, `open_date`,  or `close_date`
- `orderDesc`: boolean, as a string literal (where only a value of `"true"` affects the behavior).

However, a request that doesn't conform to these simply proceeds along the path of an unspecified default behavior – an `ordering` querystring containing unsupported input yields a response with arbitrarily-ordered results rather than an error, which makes problems harder to track down. I think it would be good to instead enforce the expectations described above within the `buildOrderingParams()` function, which can be responsible for either throwing an error (which should result in a response with a `400` status code when called by a route handler) and otherwise reliably provide a object that is actually structured in an expected manner.

@dysmento As mentioned, the changes I'm advocating for here larger than what's required for #1396 (which this PR currently satisfies) although still support these changes. If you agree with the proposal and don't think it would be too much of a lift to implement, I think it would be fine to include them here. Otherwise, I can create a separate ticket and mark this PR as approved. Thoughts?

dysmento:
I'll go ahead with your suggestions. Thanks!

dysmento:
@TylerHendrickson I have factored the validation of sort params into the `buildOrderingParams` function, and now the routes that use it will return a 400 error if it fails validation.

TylerHendrickson (approved):

Issue #1808, opened 2023-08-19
[Bug]: Actually expire non-current S3 objects


TylerHendrickson:
### Why is this issue important?

There are three S3 buckets provisioned within the GOST service, all of which have object versioning enabled, which helps mitigate the possibility of data loss (e.g. during accidental (soft-)deletion) and generally allows us to restore S3 data if/when necessary as part of a maintenance activity.

Two of these buckets (although all three should) have lifecycle rules enabled, which facilitate automatic cleanup of S3 objects that are no longer in-use. We do this for many reasons, both in the interest of cost as well as (in the case of stored partner data) to mitigate risk by ensuring that data is not retained after it no longer provides tangible value to us or our partners.

### Current State

Currently, all three buckets are not properly configured to ensure that **non-current** object versions are fully deleted. 

For example, the ARPA reports bucket defined via the `module.arpa_audit_reports_bucket` resource in `terraform/modules/gost_api/storage.tf` is configured to expire objects after 14 days. However, its `noncurrent_version_expiration` configuration is misconfigured, which is resulting in indefinite retention of outdated S3 objects which have either been expired or replaced (i.e. uploaded over) by a newer object version. Therefore, when the lifecycle rules expire an object after 14 days, it is simply marked as an expired/non-current version (i.e. soft-deleted), but no follow-up action exists to fully remove the object.

What we have is this:
  - `Current version --(   14 days later  )--> non-current version`
  - `Current version --(replaced by upload)--> non-current version`

But what we actually want is: 
  - `Current version --(   14 days later  )--> non-current version --(7 days later)--> deleted`
  - `Current version --(replaced by upload)--> non-current version --(7 days later)--> deleted`

In two cases, this seems to be the result of a typo in Terraform. For example, we currently have `noncurrent_version_expiration = { days = 7 }` defined for the ARPA reports bucket, which mimics the structure of the `expiration = { days = 14 }` input variable. However, `noncurrent_version_expiration` actually expects a key of `noncurrent_days` rather than `days`, as demonstrated in the examples [here](https://github.com/cloudposse/terraform-aws-s3-bucket/blob/main/README.md). Annoyingly, this misconfiguration doesn't seem to result in a terraform validation error.

### Expected State

Lifecycle rules for the three S3 buckets are updated to fully delete non-current object versions according to the following rules:
- ARPA reports bucket:
  - Deletes ("expires") non-current versions after 7 days.
- Website logs bucket:
[...]

PR #2170, opened 2023-11-01
Add and fix configuration to expire non-current s3 objects

ifranch:
Fixes issue #1808 
## Description

- Update the terraform for S3 buckets arpa_audit_reports_bucket and logs_bucket to use the correct noncurrent_days parameter rather than days.
- Add new lifecycle rules for the origin S3 bucket to expire noncurrent versions after 7 days, and not expire or transition objects otherwise.


TylerHendrickson:
Closing in favor of #2179

Issue #2202, opened 2023-11-09
partner can't delete users


ClaireValdivia:
### Why is this issue important?

Mel Evans from City of Sparks reached out trying to delete two users and she's not able to. Logged in and tried to do so as well and was not able to - delete button does nothing

### Current State

Delete button on User page does not remove user. Aditya confirmed that any users who have a saved search can't be deleted.

The error trace can be found [here in Datadog](https://app.datadoghq.com/apm/trace/1488372201979358490?colorBy=service&env=production&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=1488372201979358490&spanViewType=errors&timeHint=1699541585263.001&view=spans).

```
error: delete from "users" where "id" = $1 - update or delete on table "users" violates foreign key constraint "grants_saved_searches_created_by_foreign" on table "grants_saved_searches"
    at Parser.parseErrorMessage (/app/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/app/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/app/node_modules/pg-protocol/dist/parser.js:39:38)
    at TLSSocket.<anonymous> (/app/node_modules/pg-protocol/dist/index.js:11:42)
    at TLSSocket.emit (node:events:520:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at TLSSocket.Readable.push (node:internal/streams/readable:228:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23)
    at TLSWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

As shown above, the cause of this issue is the existence of a foreign key constraint, `grants_saved_searches_created_by_foreign` on table `grants_saved_searches`, which links the `grants_saved_searches.created_by` column to the `users.id` column. Therefore, when user with one or more saved searches is being deleted, the attempt fails (due to a violation of the foreign key constraint) because it would "orphan" the `grants_saved_searches` rows associated with that user.

### Expected State

- Delete button should remove user regardless of whether they have a saved search.
- When the user being deleted has created saved searches, those saved searches are also deleted.

### Implementation Plan

[...]

PR #2207, opened 2023-11-10
fix: add migration for cascading delete on users -> saved searches #2202

dysmento:
## Description
Adds the necessary migration to update the foreign key constraint on `grants_saved_searches.created_by`

This will allow deleting of users with saved searches.

I use four transactions within migration for each of the four steps described in the ticket.


as1729 (approved):
🚀

PR #2241, opened 2023-11-21
feat: add profile header quick menu

ifranch:
## Description

Update the header dropdown menu options when the profile feature flag is enabled. Changes are:

- Replacing "Settings" with "My Profile" and directing to /my-profile page
- Adding icons for each options
- Updating the width of the dropdown menu
- Adding a margin between the icons and the text
- Making the text lowercase as seen in the mocks
- Updating the font size to match the mocks

## Screenshots / Demo Video

Before/with flag not enabled:
Screenshot 2023-11-20 at 9 54 16 PM

With flag enabled:
Screenshot 2023-11-20 at 9 53 25 PM


dysmento (approved):
Cool!

PR #2248, opened 2023-11-21
feat: digest email design update

SDBowen:
## Description

This PR implements the design changes to the digest email. The design can be found [here in Figma](https://www.figma.com/file/HIINnuujheHbdbDesH8mKE/Notifications?type=design&node-id=17-1160&mode=design&t=TXQTPbVujxdboPKP-0)

Some additional data (the saved search name and date) was made available to the `buildDigestBody()` function as they are needed in the updated email template.

Inside the `getGrantDetail()` function, some logic was added to strip `<p>` tags from the description that gets used in the `_grant_detail.html` mail template. Removing these tags allowed for the `View on Grants.gov` to appear on the same line as the description. 

I am not certain if the above process to remove these tags is necessary. Our seed data contains Grant descriptions enclosed with `<p>` tags. But, when I look at the Grant data in production, I do not see Grant descriptions with these `<p>` tags. So, unclear if these `<p>` tag enclosed descriptions still exist.

## Screenshots / Demo Video

**Gmail**
Screenshot from 2023-11-21 11-36-00

**Outlook  (desktop)**
Screenshot 2023-11-21 132806
Screenshot 2023-11-21 132838

## Testing

Enable `New Grants Digest` in your profile and create a Saved Search to receive a grant digest email when new grants are processed.

### Automated and Unit Tests
- [x] Updated Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/server/src/lib/email.js
```diff
@@ -117,10 +117,14 @@ function sendWelcomeEmail(email, httpOrigin) {
 
 function getGrantDetail(grant, emailNotificationType) {
     const grantDetailTemplate = fileSystem.readFileSync(path.join(__dirname, '../static/email_templates/_grant_detail.html'));
+
+    const shortDescription = grant.description?.length > 380 ? grant.description.substring(0, 380) : grant.description;
+    const description = shortDescription.replace(/<\/?p[^>]*>/g, '');
```
TylerHendrickson, line 122: @SDBowen Re 
> I am not certain if the above process to remove these tags is necessary. Our seed data contains Grant descriptions enclosed with `<p>` tags. But, when I look at the Grant data in production, I do not see Grant descriptions with these `<p>` tags. So, unclear if these `<p>` tag enclosed descriptions still exist.

I don't believe that the grant description text should ever include HTML; my suspicion is that the fixtures were based on grant data that was scraped (prior to the `grants-ingest` service), when that may have been more of a possibility. That said, it might be worth keeping, since I don't think there's any guarantee from the grants.gov data source that markup will never be present in a text value.

In fact, it might be worth taking this one step further to remove _any_ HTML (within reason) present in the original `grant.description` text, and then truncate the result to the desired length.

```suggestion
    const description = grant.description.replace(/(<([^>]+)>)/ig, '').substring(0, 380);
```

TylerHendrickson (changes requested):
@SDBowen Changes look good! Two small comments for you to consider.

packages/server/src/lib/email.js
SDBowen: Makes sense and cleaner :+1:

TylerHendrickson (approved):

PR #2274, opened 2023-11-28
fix: switch to use vue-select instead of Multiselect

adele-usdr:
## Update new search and edit search panel controls

Switch to use vue-select control instead of Multiselect. Follow-up bug coming to add more description text.

## Screenshots / Demo Video
https://github.com/usdigitalresponse/usdr-gost/assets/146878468/07d47d68-1e2c-47fb-8de9-ff85ff4283b8

## Testing
Select different options in the search panel. Make sure clicking out to close menu still works. Verify no extra x shows up in single selection fields.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [x ] Added steps to test feature/functionality manually

## Checklist
- [x ] Provided ticket and description
- [x ] Provided screenshots/demo
- [x ] Provided testing information
- [x ] Added PR reviewers


packages/client/src/components/Modals/SearchPanel.vue
Vikariusu, line 1: Is the code on line 472 (css for `.search-panel .search-fields-radio-group`) still relevant? 
Seems like it was a workaround when using a Multiselect component, removing this styling doesn't seem to cause radio buttons to overlap with any dropdowns.

adele-usdr: good call! Just pushed a new commit to remove this.

Vikariusu:
The functionality looks great! 
One question I have is how closely are we following the Figma file? If we want to follow the design, the selected options would have a background of $raw-gray-200, no border etc.

Nit: is the "Eligibility" dropdown the only one that can be typed in? This means the arrow on the right that expands and collapses the dropdown doesn't get the cursor: pointer treatment.

adele-usdr:
Good call- making a few more tweaks to match the figma design.  @caitlinwinner how does this look?

Screenshot 2023-11-30 at 9 56 11 AM

adele-usdr:
Confirmed with Caitlin that we don't need to use a custom style for these drop downs at this time.  Also fixed an issue to stop Enter from submitting the form for dropdowns that are supposed to stay open after options are selected.  This also avoids a Safari-specific bug where the Eligibility dropdown was getting stuck open after hitting Enter.

Vikariusu (approved):

Issue #2218, opened 2023-11-14
Remove old Grants table implementation


adele-usdr:
### Why is this issue important?

Discussed in Grants Finder meeting that we are ready to remove the old grants table implementation. We should rename the new one, so there's only one grants table when folks go looking for it.

### Current State

The in-use grants table is GrantsTableNext.vue and the old one is GrantsTable.vue

### Expected State

Current grants table implementation should be at GrantsTable.vue.

### Implementation Plan

Move current implementation to GrantsTable.vue, remove feature flag.

PR #2324, opened 2023-12-08
fix: #2218 Additional cleanup from removing useNewTable feature flag

adele-usdr:
## Description
Remove last use of the useNewTable feature flag in production


adele-usdr:
Pushed an additional change to actually remove GrantsTableNext.vue

TylerHendrickson (approved):

PR #2342, opened 2023-12-12
feat: switch grant finder to use history API for client-side routing

sanason:
## Description
This PR should satisfy the first requirement in the associated ticket:

> GA4 page_view events fire on state changes that generate a unique URL

The second requirement (adding an "agency_name" parameter) will be handled in a separate PR.

Nothing in this PR refers directly to Google Analytics. The idea is that switching the router to use the HTML5 history API will automatically fix our Google Analytics problem, for reasons explained in [this PR](https://github.com/usdigitalresponse/usdr-gost/pull/2323). An additional benefit is that Grant Finder will have nicer URLs - instead of `https://grants.usdigitalresponse.org/#/my-grants`, the URL will be `https://grants.usdigitalresponse.org/my-grants`. To avoid breaking any existing bookmarks or emailed links, URL's with hashes will be redirected to the non-hash version.

## Screenshots / Demo Video
Here is evidence that we're getting page view events for all different paths. I think that all of the data for paths other than `/` comes from me navigating in my dev environment - it appears that production, staging and dev are all feeding data into the same data stream.

Screenshot 2023-12-18 at 11 49 23 AM

## Testing
No feature flags to set, just the usual `docker compose up`.

I tested the following things manually:

- For all grant finder routes, the non-hash URL can be loaded directly (type into the address bar and hit enter).
- For all grant finder routes, the old hash URL can be loaded directly and redirects to the non-hash equivalent.
- All internal links works.
- The login flow still works, including post-authentication redirects.
- URL's in emails sent by grant finder use the new non-hash style.
- Unknown routes display a 404 page.
- Arpa reporter routes still work.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/vue.config.js
+++ b/packages/client/vue.config.js
@@ -11,6 +11,16 @@ module.exports = {
       options.patterns[0].globOptions.ignore.push('**/deploy-config.js');
       return [options];
     });
+    // Added as workaround for bug in @vue/cli-plugin-unit-mocha.
+    // https://github.com/vuejs/vue-cli/issues/4053
+    // Unit tests report webpack compilation failures without this workaround.
+    if (process.env.NODE_ENV === 'test') {
+      const scssRule = config.module.rule('scss');
+      scssRule.uses.clear();
+      scssRule
+        .use('null-loader')
+        .loader('null-loader');
+    }
   },
   pages: {
     main: {
--- a/docs/development.md
+++ b/docs/development.md
@@ -60,7 +60,7 @@ See [here](../docker/README.md) for more information about commands to use when 
 
    ![AWS SES Error](./img/error-aws-ses.png)
 
-1. Visit `client_url/login` (e.g <http://localhost:8080/#/login>) and login w/ user `grant-admin@usdigitalresponse.org`. You'll see a confirmation message on the screen. Check your logs to find the generated session link.
+1. Visit `client_url/login` (e.g <http://localhost:8080/login>) and login w/ user `grant-admin@usdigitalresponse.org`. You'll see a confirmation message on the screen. Check your logs to find the generated session link.
 
    **NOTE:** if you only see a blank screen then ensure you've set up the `packages/client/.env`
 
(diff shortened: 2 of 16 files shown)


packages/client/vue.config.js
```diff
@@ -11,6 +11,16 @@ module.exports = {
       options.patterns[0].globOptions.ignore.push('**/deploy-config.js');
       return [options];
     });
```
sanason, line 13: Adding the new router unit test revealed the following error when tests were run:

```
WEBPACK  Failed to compile with 1 error(s)

Error in ./src/components/SearchFilter.vue?vue&type=style&index=0&id=7c053370&lang=scss&

  Module build failed (from ../../node_modules/sass-loader/dist/cjs.js):
  SassError: Can't find stylesheet to import.
    ╷
  2 │ @import '../../scss/colors-semantic-tokens.scss';
    │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵
    Users/nason/Jobs/USDR/Repositories/usdr-gost/packages/client/src/components/SearchFilter.vue 2:9  root stylesheet
```

I think this appeared now because the router necessarily imports a lot of components so the router test was the first test to (indirectly) reference `SearchFilter`.

TylerHendrickson:
> I think that all of the data for paths other than `/` comes from me navigating in my dev environment - it appears that production, staging and dev are all feeding data into the same data stream.

@sanason Not necessarily in-scope for this issue, but does GA have a concept of environments/stages that can be used to filter out non-Production activity? Or would we need to create separate properties for each stage and update our code with environment-specific tracking codes?

TylerHendrickson (approved):

sanason:
> Or would we need to create separate properties for each stage and update our code with environment-specific tracking codes?

This, I think. I created a ticket to do so [here](https://github.com/usdigitalresponse/usdr-gost/issues/2395).

PR #2394, opened 2023-12-19
feat: remove "Occasional Updates" toggle from profile page

sanason:
## Description
I removed the toggle from the "My Profile" page (see screenshot). I didn't delete any of the other code added to support "Occasional Updates" because my impression is that we want to restore this feature once we get email lists working.

I also fixed an issue I discovered during testing - toggling any of email settings on "My Profile" was causing a 403 error in the background. I don't think this was affecting the user experience but it could cause noise in Datadog. See in-code comments for details.

## Screenshots / Demo Video
Screenshot 2023-12-20 at 3 06 50 PM

## Testing
Load the app and check there is no toggle on the profile page. No relevant feature flags.

I manually tested the following and saw no errors logged in the browser console:
- Load the "My Profile" page
- Toggle on/off the remaining 2 toggles (New Grant Digest, Grants Assignment)

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/client/src/store/modules/users.js
```diff
@@ -79,15 +79,14 @@ export default {
       }
       await dispatch('fetchUsers');
     },
-    async updateEmailSubscriptionPreferences({ dispatch, commit }, { userId, preferences }) {
```
sanason, line 82: All existing uses of this method were being called on the logged in user, plus line 89 below clearly assumes that the method is only called on the logged in user, so I updated the method to only apply to the logged in user.

packages/client/src/store/modules/users.js
```diff
@@ -79,15 +79,14 @@ export default {
...
+    async updateEmailSubscriptionPreferencesForLoggedInUser({ commit, getters }, { preferences }) {
       const data = await fetchApi.put(
-        `/api/organizations/:organizationId/users/${userId}/email_subscription`,
+        `/api/organizations/:organizationId/users/${getters.loggedInUser.id}/email_subscription`,
         {
           preferences,
         },
       );
       commit('SET_LOGGED_IN_USER', data.user);
-      dispatch('fetchUsers');
```
sanason, line 90: This API call was returning '403 - Forbidden' with my test user. I think only admin users are permitted to fetch users. But it's safe to remove this call since the previous line updates the store for the modified user.

packages/client/src/components/EmailSettingsBanner.vue
sanason, line 1: I started to fix this component to use `updateEmailSubscriptionPreferencesForLoggedInUser` but then I noticed that it isn't referenced anywhere so I think it's safe to delete instead.

TylerHendrickson (approved):

PR #2430, opened 2024-01-04
fix: ensure we ignore microsoft safe links requests

as1729:
## Description
This PR adds the following:
1. Increases the number of clicks of the passcode link that will allow the user to login.
2. Ignores any requests originating from Microsoft Safe Links

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/routes/sessions.js
+++ b/packages/server/src/routes/sessions.js
@@ -3,6 +3,7 @@ const _ = require('lodash-checkit');
 const path = require('path');
 const { sendPassCode } = require('../lib/email');
 const { validatePostLoginRedirectPath } = require('../lib/redirect_validation');
+const { isMicrosoftSafeLinksRequest } = require('../lib/access-helpers');
 
 const router = express.Router({ mergeParams: true });
 const {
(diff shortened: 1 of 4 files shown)


packages/server/src/routes/sessions.js
```diff
@@ -16,12 +16,20 @@ const { isUSDRSuperAdmin } = require('../lib/access-helpers');
...
 // links. Now, we send login links through a clientside redirect instead so this should not be necessary.
-const MAX_ACCESS_TOKEN_USES = 1;
+// This has been updated to accommodate Microsoft Safe Links which will prefetch the link twice.
+const MAX_ACCESS_TOKEN_USES = 4;
 
 // the validation URL is sent in the authentication email:
 //     http://localhost:8080/api/sessions/?passcode=97fa7091-77ae-4905-b62e-97a7b4699abd
 //
 router.get('/', async (req, res) => {
+    const userAgent = req.headers['user-agent'] || '';
```
tzinckgraf, line 26: I would recommend moving this out to a different function. It would be easy enough to test, and then we could add to it if we see this same issue with another service.

tzinckgraf (approved):

tzinckgraf (approved):

packages/server/src/routes/sessions.js
```diff
@@ -14,14 +15,14 @@ const {
 } = require('../db');
 const { isUSDRSuperAdmin } = require('../lib/access-helpers');
 
-// NOTE(mbroussard): previously we allowed 2 uses to accommodate automated email systems that prefetch
-// links. Now, we send login links through a clientside redirect instead so this should not be necessary.
-const MAX_ACCESS_TOKEN_USES = 1;
+// Increasing the max-uses to ensure users are able to log-in even if their email client/security provider has clicked on the link already.
+// Specifically the issue was identified with Microsoft Safe Links, which clicks on the link to check if it is safe.
+const MAX_ACCESS_TOKEN_USES = 4;
```
TylerHendrickson, line 20: @as1729 I'm a bit late to the party, but I'm curious why this is necessary alongside the `isMicrosoftSafeLinksRequest` middleware. If it's not, I think a follow-up PR is warranted to turn this back down.

My reading of the middleware is that it will always emit a 200 status response and stop handling the request before the actual router handler function is called, and therefore the token won't have actually be used to authenticate a session. If that's accurate, then it doesn't seem like increasing the `MAX_ACCESS_TOKEN_USES` limit is necessary. Preventing token reuse (for actually authenticating a session) would be much preferred from a security standpoint.

In other words, it seems like the route should be accessible without any limit as long as `isMicrosoftSafeLinksRequest` returns early, because we're not actually establishing a new session for the user. But once a magic link token has authenticated a user, it should not be able to do so a second time.

PR #2441, opened 2024-01-09
metrics: build website with different Google tag IDs for different en…

sanason:
…vironments

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (changes requested):
@sanason Although it's not the most important thing, having same-content builds* affords some advantages, such as opportunities for reducing build/deploy time in our pipelines with caching. IMO it would be ideal to avoid different build inputs in different environments, as long as we can reasonably do so. Thus far, we've been able to use the Terraform-managed `deploy-config.js` artifact to provide values for environment-specific variables at runtime via the `window.APP_CONFIG` global object.

Rather than using `--mode` here, the alternate Terraform route would consist of something like this:
- Most changes would happen in the `terraform/usdr-gost/terraform/modules/gost_website` module: 
  - `tpl/deploy-config.js` would be updated to add a (nullable) Google tag ID as a property on the `window.APP_CONFIG` object.
  - `variables.tf`: Add an optional (`default = null`) variable for passing in a Google tag ID.
  - `runtime_config.tf`: Responsible for passing in the Google tag ID variable to the `templatefile()` call that renders `tpl/deploy-config.js`.
- Supporting changes to the top-level Terraform project in the `terraform/` directory:
  - `terraform/variables.tf`: Basically a duplicate of the change to the other `variables.tf` file mentioned above, allowing a Google tag ID to be provided as an optional variable when Terraform runs.
  - `terraform/prod.tfvars`: Hard-code the Google tag ID so that it gets automatically included when Terraform runs against our Production environment.
  - `terraform/main.tf`: Responsible for passing the new variable provided at the top-level Terraform project down into the `website` module.
- Finally, the client package changes for `packages/client/public/index.html` would need minor refactoring so that the Google tag ID is pulled from the `window.APP_CONFIG` object rather than `process.env`.

*in lieu of immutable builds, since true immutability is difficult to achieve with webpack.

PS- Happy to chat about this further if the changes above can be made clearer and/or if you want to discuss the merits of this approach in more detail, particularly given that the process for implementing new `window.APP_CONFIG` values is admittedly a bit convoluted.

sanason:
> Although it's not the most important thing, having same-content builds* affords some advantages, such as opportunities for reducing build/deploy time in our pipelines with caching. IMO it would be ideal to avoid different build inputs in different environments, as long as we can reasonably do so. 

@TylerHendrickson Makes sense. Updated to use the Terraform approach.

TylerHendrickson (approved):
Looks great!

PR #2442, opened 2024-01-09
chore: add logging around safe links requests

as1729:
## Description
This PR adds logging to capture request headers if we identify a request originating from Microsoft. The goal is to get an understanding of the volume of these requests and see if we are able to accurately intercept these requests purely based on the data present in the headers. Once we confirm that we are logging these requests appropriately, we can reduce the number of `MAX_ACCESS_TOKEN_USES` back down to 1.

## Screenshots / Demo Video
N/A

## Testing
N/A

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/lib/access-helpers.js
+++ b/packages/server/src/lib/access-helpers.js
@@ -129,7 +129,11 @@ async function isMicrosoftSafeLinksRequest(req, res, next) {
     const userAgent = req.headers['user-agent'] || '';
     const nativeHost = req.headers['x-native-host'] || '';
     if (userAgent.toLowerCase().includes('oneoutlook') || nativeHost.toLowerCase().includes('oneoutlook')) {
-        log.info({ headers: req.headers }, 'Microsoft Safe Links request');
+        log.info({
+            'user-agent': userAgent,
+            'native-host': nativeHost,
+            headers: Object.keys(req.headers),
+        }, 'Microsoft Safe Links request');
         res.json({ message: 'Success' });
         return;
     }
--- a/packages/server/src/lib/access-helpers.js
+++ b/packages/server/src/lib/access-helpers.js
@@ -1,4 +1,5 @@
 const { getUser, inTenant } = require('../db');
+const { log } = require('./logging');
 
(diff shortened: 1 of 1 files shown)


packages/server/src/lib/access-helpers.js
```diff
@@ -128,6 +129,7 @@ async function isMicrosoftSafeLinksRequest(req, res, next) {
     const userAgent = req.headers['user-agent'] || '';
     const nativeHost = req.headers['x-native-host'] || '';
     if (userAgent.toLowerCase().includes('oneoutlook') || nativeHost.toLowerCase().includes('oneoutlook')) {
+        log.info({ headers: req.headers }, 'Microsoft Safe Links request');
```
as1729, line 132: Note: if we release this change to production and not see any evidence of this log then it is clear that we have not truly identified what a request from safe-links looks like. If that is the case I'll go ahead and add additional logging around the whole request to get better data.

TylerHendrickson, line 132: Overabundance of caution, but because `req.headers` _might_ contain values that we'd want to remain out of logs, I'd recommend constraining this a bit so that only the `user-agent` header value is logged:
```suggestion
        log.info({ 'user-agent': req.headers['user-agent'] }, 'Microsoft Safe Links request');
```

or, for a bit more insight into the complete set of headers:
```suggestion
        log.info({ 'user-agent': req.headers['user-agent'], headers: Object.keys(req.headers) }, 'Microsoft Safe Links request');
```

TylerHendrickson (changes requested):

packages/server/src/lib/access-helpers.js
as1729: Good idea 👍  I took your suggestion and added an additional key `nativeHost`

TylerHendrickson: Ah, good idea adding that. Will approve!

TylerHendrickson (approved):

PR #2478, opened 2024-01-19
Add .gitattributes file for this repo

TylerHendrickson:
This PR simply defines a [`.gitattributes` file](https://git-scm.com/docs/gitattributes) for the `usdr-gost` repository. This change is primarily intended to ease contributor burden when it comes to developing on various operating systems, each with their own opinions on how newlines should be handled (which this PR standardizes as unix-style `LF` (or `\n`) endings). It also includes a few filetype-specific overrides.

NB: This is a direct copy of the [same file used in the cpf-reporter repository](https://github.com/usdigitalresponse/cpf-reporter/blob/main/.gitattributes).


sanason (commented):
LGTM

sanason (approved):

PR #2507, opened 2024-01-25
Add pdf to list of binary file types in .gitattributes.

sanason:
`packages/server/src/static/forms/nevada_spoc.pdf` keeps showing up as modified, even though I haven't touched it, and I've gotten this warning:

> warning: in the working copy of 'packages/server/src/static/forms/nevada_spoc.pdf', CRLF will be replaced by LF the next time Git touches it


TylerHendrickson (approved):
Good catch – approved!

PR #2515, opened 2024-01-26
Exclude .xlsx and .xlsm files from line ending autofixes

ajhyndman:
## Description

Without this change, switching between branches with different versions of t.xlsx or .xlsm files in this repo will cause git to get confused.

Git will attempt to auto-convert anything that looks like a CLRF line ending inside of one of these excel files which could corrupt the file and causes a new uncommitted change to appear unexpectedly.

This change explicitly marks excel files as binary so that they are exempt from this kind of autoconversion.

## Checklist
- ~Provided testing information~
- [x] Added PR reviewers


as1729 (approved):

packages/.gitattributes
```diff
@@ -20,3 +20,5 @@
 *.ttf               binary
 *.woff              binary
 *.woff2             binary
+*.xlsm              binary
+*.xlsx              binary
```
TylerHendrickson, line 24: Just for completeness...
```suggestion
*.xls               binary
*.xlsm              binary
*.xlsx              binary
```

TylerHendrickson (changes requested):
@ajhyndman Thanks for this fix! 

One additional request: would you mind moving the `.gitattributes` to the repository root? I'd meant to add it there in the first place; nesting it under `packages/` was an oversight on my part.

ajhyndman:
Sure, good ideas!  Updated

jeffsmohan:
👋 Just dropping in to say I ran into this with a fresh development environment, and I'm excited for the fix to land! (A fresh checkout of the git repo before I've even touched anything shows 21 files as dirty, which was disconcerting...)

TylerHendrickson (approved):

PR #2524, opened 2024-01-29
Add user script to make email login easier in development

jeffsmohan:
## Description

While getting my development environment working for the first time, I struggled a bit with setting up email-based login. This PR updates the docker development docs to walk new developers through setting up dev email sending and creating a new user. 

This relies on a new `user_creation` helper script (based heavily on the [tenant_creation.js](https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/server/src/db/tenant_creation.js) script). 

👋 I'm totally new to USDR and the Grants project, so if there was already an easier/better way to get email logins working in dev, please let me know! I'm happy to just close out the PR and call it a good learning exercise. Also, please nudge me in the right direction if anything about my workflow is off base here... thanks! 

## Screenshots / Demo Video
Screenshot 2024-01-29 at 11 11 13 AM
image

## Testing
- Go through the steps outlined in the docker documentation file
- Ensure you can now login via emailed link for the created user

### Automated and Unit Tests
- [ ] Added Unit tests
(I'm not sure if tests are normative for a script like this in the codebase... it's intended for internal dev use only, and the tenant_creation.js script wasn't tested as far as I could tell. But I'm happy to write some up if this code seems worth merging!)

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -22,8 +22,11 @@ steps to prepare your environment:
 4. Run `docker compose up -d` to start the services in the background (the `-d` flag).
 5. Install application dependencies via yarn: `docker compose exec app yarn`.
 6. Visit http://localhost:8080 to confirm you see 'Grants Identification Tool' with a prompt to enter an email address.
-If you see a blank screen, [review the logs](#cookbook-logs), you may need to [run a db migrate](#cookbook-db-migrate) and restart the app container.
-
+  - If you see a blank screen, [review the logs](#cookbook-logs), you may need to [run a db migrate](#cookbook-db-migrate) and restart the app container.
+7. To enable email-based login, you'll need email sending and a user:
+  - Seed the database (which automatically runs outstanding migrations): `docker compose exec app yarn db:seed`
+  - Set up [email sending via Gmail](../docs/setup-gmail.md)
+  - Add a user with your email: `docker compose exec app yarn create_user`
 
 **Note:** Some systems may have Docker Compose installed as an integrated plugin for Docker,
 while others may have it installed as a standalone executable (e.g. via `pip install`).
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "build": "yarn workspace client run build",
     "coverage": "nyc --reporter=lcov --reporter=text-summary yarn workspaces run test",
     "create_tenant": "yarn workspace server create_tenant",
+    "create_user": "yarn workspace server create_user",
     "db:migrate": "yarn workspace server run db:migrate",
     "db:seed": "yarn workspace server run db:seed",
     "debug:server": "yarn workspace server debug",
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -14,6 +14,7 @@
     "consume-grants": "node src/scripts/consumeGrantModifications.js",
     "coverage": "nyc yarn test",
     "create_tenant": "node src/db/tenant_creation.js",
+    "create_user": "node src/db/user_creation.js",
     "migrate_saved_searches": "node src/db/saved_search_migration.js",
     "db:migrate": "knex migrate:latest",
     "db:rollback": "knex migrate:rollback",
--- a/packages/server/src/db/user_creation.js
+++ b/packages/server/src/db/user_creation.js
@@ -1,0 +1,127 @@
+require('dotenv').config();
+
+// This script creates a new staff user under a given tenant and agency.
+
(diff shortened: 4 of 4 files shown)


docker/README.md
```diff
@@ -22,8 +22,11 @@ steps to prepare your environment:
 4. Run `docker compose up -d` to start the services in the background (the `-d` flag).
 5. Install application dependencies via yarn: `docker compose exec app yarn`.
 6. Visit http://localhost:8080 to confirm you see 'Grants Identification Tool' with a prompt to enter an email address.
-If you see a blank screen, [review the logs](#cookbook-logs), you may need to [run a db migrate](#cookbook-db-migrate) and restart the app container.
-
+  - If you see a blank screen, [review the logs](#cookbook-logs), you may need to [run a db migrate](#cookbook-db-migrate) and restart the app container.
+7. To enable email-based login, you'll need email sending and a user:
+  - Seed the database (which automatically runs outstanding migrations): `docker compose exec app yarn db:seed`
+  - Set up [email sending via Gmail](../docs/setup-gmail.md)
```
TylerHendrickson, line 28: I would like to avoid encouraging use of the `setup-gmail.md` instructions, for these reasons:

1. While in certain cases, it may be absolutely necessary for a developer to be able to send email from their development environment (like when working on a ticket directly related to email templating/sending behavior), email is generally not necessary for local development.
2. These instructions assume developers use Gmail, which may not always be the case.
3. Although there are controls that help secure this setup, it still involves generating static credentials that have the ability to send email, which should be avoided if possible. This is especially true given that a lower-friction option is to set up a "Allow Less Secure Apps", which I like even less.
4. It's a workflow that's difficult to provide support for if/when someone runs into issues.
5. (Not specific to Gmail integration, but related) Anytime a user nukes their Docker DB, they need to recreate any custom users.

I think as much as possible, it would instead be preferable to lean into supporting developer workflows that can sidestep email requirements. A few things that seem worthy of mention...
1. For login scenarios, the backend (`app` container) server does log generated auth tokens in development environments.
2. When in-use, the LocalStack container [serves an endpoint](https://docs.localstack.cloud/user-guide/aws/ses/#sent-emails) that makes it possible to retrieve emails "sent" to the LocalStack-mocked SES service.

One idea I can offer is that we could update the `db:seed` script to create users that are less representative of specific people and more representative of certain personas that developers can use for testing in local environments. I'm also interested in hearing others' thoughts and ideas around how to optimize the developer experience surrounding this task, since it's obviously one of (if not the single) most common developer workflow.

TylerHendrickson: NB: I don't think any of the above needs to hold up this specific PR, since it's seems useful under the current status-quo, and I don't want to make perfect be the enemy of good here. Given that this PR is still in "Draft", though, I figured it would be worth starting the discussion here, even if we ultimately move any of its resulting action items elsewhere.

jeffsmohan: Thanks for the additional context, that's super helpful! (As I was setting up my dev environment based on the docker setup instructions, I got briefly stuck at logging in. I did find the login URL in the logs, but I wasn't sure if using that was sidestepping some part of a fully set up dev stack.) 

It sounds like a good developer workflow is achievable with pretty minimal changes:

- Run the DB seed for a new installation, which installs some generic accounts (`admin@usdigitalresponse.org` and `staff@usdigitalresponse.org`)
- Tail the docker logs (`docker compose logs app -f`)
- Use one of the generic accounts at `http://localhost:8080/login`, then follow the login link output in the logs

In order to make this workflow a little more seamless, I'd propose: 

1. Update the DB seed code to generate the generic admin and staff user accounts
2. Update the `Login link generated` console log to be slightly more visible (e.g., add `console.log('--------')` above and below that line, or similar)
3. Document the intended developer login flow at the end of the Docker setup instructions
4. Add a callout box at the top of the Gmail setup documentation page, noting it shouldn't be required except in special circumstances (e.g., working on emails themselves)

@TylerHendrickson if the above makes sense to you, I can close out this draft PR and work up a new one implementing the changes I laid out!

TylerHendrickson: @jeffsmohan That idea sounds great- thanks for the help!

jeffsmohan:
Per discussion [in the above thread](https://github.com/usdigitalresponse/usdr-gost/pull/2524#discussion_r1470332892) I'll close this PR out and follow up with a new one taking a different approach to easier dev logins

PR #2531, opened 2024-01-30
Re-commit bulk upload templates

ajhyndman:
Re-commit previously corrupted bulk upload templates from #2514 after landing .gitattributes fixes in #2515.


TylerHendrickson (approved):

as1729 (changes requested):
I see a bit of a discrepancy with #2530

PR #2539, opened 2024-01-30
Upgrade GHA usages of actions/upload-artifact and actions/download-artifact

jeffsmohan:
## Description
Upgrades Github workflows to use v4 of `actions/upload-artifact` and `actions/download-artifact`. These have a few performance improvements, but primarily have longer-term support since they're running on node v20. 

I've configured all artifact uploads with `overwrite: true` since v4 makes artifacts immutable. In general use, I don't think this would be an issue for our setup (all artifacts have the git commit hash in their filename), but I believe it would break if you manually re-ran a workflow, which would cause the same filename to get uploaded again. In this case, I believe the overwrite behavior of removing the file, then re-uploading it, should work for us. 

## Testing
I've so far been unable to test these changes. Maybe someone with a deeper understanding of GH workflows can help point me in the right direction?  

image

[This stack overflow](https://stackoverflow.com/questions/67707244/how-to-test-github-workflow-without-merging-into-master-main-branch) points to some options, but I wasn't able to get any working well here: 

- Push your code and create a PR. (I see folks suggesting this, but so far on this draft PR I'm not able to get it working. It seems to be running the un-updated workflows from `main` branch. I did see some mentions that this is for security reasons, since it would allow arbitrary code execution...)
- Trigger the workflow via the GitHub CLI. (I haven't been able to get this to work either.)
- Use [nektos/act](https://github.com/nektos/act) to run the workflows locally — this seems real heavy-handed and I'd like to avoid the setup

### Automated and Unit Tests
Not tested

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,7 +210,7 @@ jobs:
           INSPECT_NAME: ${{ inputs.server-image-registry }}/${{ inputs.server-image-name }}@${{ steps.build-push.outputs.digest }}
       - name: Upload attestations
         if: steps.store-attestations.outcome == 'success'
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
         with:
           name: ${{ env.ATTESTATION_ARTIFACTS_KEY }}
           path: ${{ steps.store-attestations.outputs.path }}
@@ -275,7 +275,7 @@ jobs:
           ENDOFREPORT
           cat "$REPORT_FILE" >> $GITHUB_STEP_SUMMARY
       - name: Upload build artifact
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
         with:
           name: ${{ env.ARTIFACTS_KEY }}
           path: ${{ env.ARTIFACTS_PATH }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -216,6 +216,7 @@ jobs:
           path: ${{ steps.store-attestations.outputs.path }}
           if-no-files-found: error
           retention-days: ${{ inputs.server-image-artifacts-retention-days }}
+          overwrite: true
 
   website-bundle:
     name: Build website deployment artifact
@@ -281,3 +282,4 @@ jobs:
           path: ${{ env.ARTIFACTS_PATH }}
           if-no-files-found: error
           retention-days: ${{ inputs.website-artifact-retention-days }}
+          overwrite: true
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -171,12 +171,12 @@ jobs:
             github.com:443
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: Download website build artifacts
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: actions/download-artifact@6b208ae046db98c579e8a3aa621ab581ff575935 # v4.1.1
         with:
           name: ${{ needs.build.outputs.website-artifacts-key }}
           path: ${{ needs.build.outputs.website-artifacts-path }}
       - name: Download docker build attestation artifacts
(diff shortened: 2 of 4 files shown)


TylerHendrickson:
@jeffsmohan To address the questions in the OP:

> - Push your code and create a PR. (I see folks suggesting this, but so far on this draft PR I'm not able to get it working. It seems to be running the un-updated workflows from main branch. I did see some mentions that this is for security reasons, since it would allow arbitrary code execution...)

Yes, we require that GHA workflows run against the base branch, which is necessary to support fork-based contributor workflows while maintaining good security posture.

> - Use [nektos/act](https://github.com/nektos/act) to run the workflows locally — this seems real heavy-handed and I'd like to avoid the setup

Agreed. I do have act configured and it is cumbersome to the point of not being terribly useful, especially given the amount of mocking that is necessary to test certain workflows.

----

FWIW, I think the most straightforward way to test this would be to first merge this into the `main` branch on the (i.e. your) forked repository, and then create a second, test branch and PR on the upstream repository. The test PR should then execute `pull_request` and `pull_request_target` workflows that run against the forked repo's base branch (i.e. your `main`). I can't 100% guarantee this will work, but I _think_ it should.

PS- Once finished, you may want/need to rebase your `main` branch from the upstream `usdigitalresponse/usdr-gost` repo in order to get your fork back in-sync.

jeffsmohan:
Ah yeah, fork the repo, great idea! Honestly forget that even existed on GitHub. I'll work on getting this tested in a forked repo and report back.

jeffsmohan:
I've managed to get some of the basic workflows running and passing on my forked repo. But @TylerHendrickson I'd love your opinion on how to fully test the rest. 

The issue I run into is my fork doesn't have AWS secrets set up, so it can't run downstream actions like terraform plan or terraform apply. I might be tempted to say it's alright, as long as we've tested at least some representative actions with artifact uploads/downloads, that might give us enough confidence to land this PR. But I'm not actually able to run any of the workflows that use the `download-artifact` action, which I'd really like to find a way to do. 

The only workflows that utilize the `download-artifact` action are: 

- `terraform-plan.yml`
- `terraform-apply.yml`
- `deploy-production.yml`

I assume we don't want to set up my fork to really deploy to production, but I'd love to be able to test terraform-plan and terraform-apply. You've got a lot more context than me... would it be safe to run terraform-plan/apply using our real AWS secrets? Would it be relatively easy/safe to use some staging or dev-specific AWS secrets we have lying around? Are there other easy workarounds to still be able to test `download-artifact`?

TylerHendrickson:
> I've managed to get some of the basic workflows running and passing on my forked repo. (See e.g. the checks on this PR: [jeffsmohan#2](https://github.com/jeffsmohan/usdr-gost/pull/2)) But @TylerHendrickson I'd love your opinion on how to fully test the rest.
> 
> The issue I run into is my fork doesn't have AWS secrets set up, so it can't run downstream actions like terraform plan or terraform apply. I might be tempted to say it's alright, as long as we've tested at least some representative actions with artifact uploads/downloads, that might give us enough confidence to land this PR. But I'm not actually able to run any of the workflows that use the `download-artifact` action, which I'd really like to find a way to do.
> 
> The only workflows that utilize the `download-artifact` action are:
> 
> * `terraform-plan.yml`
> * `terraform-apply.yml`
> * `deploy-production.yml`
> 
> I assume we don't want to set up my fork to really deploy to production, but I'd love to be able to test terraform-plan and terraform-apply. You've got a lot more context than me... would it be safe to run terraform-plan/apply using our real AWS secrets? Would it be relatively easy/safe to use some staging or dev-specific AWS secrets we have lying around? Are there other easy workarounds to still be able to test `download-artifact`?

@jeffsmohan I should've seen this snag – It didn't occur to me that the only time we actually download an artifact is when ship to Prod and then add the artifacts to the release following deployment.

Regarding this PR:
- Given that the `upload-artifact` changes [appear to be working in your test](https://github.com/jeffsmohan/usdr-gost/actions/runs/7745863659/job/21124271300), I feel reasonably confident in these changes.
- We'll be able to confirm these changes are working pretty quickly after merging when the CD pipeline deploys to Staging. If something _does_ fail during that run, it should be straightforward enough to roll back.

With that in mind, I think this PR should be good to merge and will approve.

More long-term: we don't currently have any AWS dev environments that we can share out credentials for, but this is a good use-case that highlights of some of the limitations in testing our workflows with limited access to secrets. I've considered adding throw-away deployments to disposable environments as a CI step before, because it would be useful to test that Terraform changes can actually run rather than just relying on the ability to generate a plan as we do now. The caveat here is that the time it takes to provision the entire stack from-scratch seems like it would be a burden to developers. There are probably some ways we can address some of the slower aspects of this process though. I'll do some thinking on this and see if I can come up with some useful ideas around this problem.

TylerHendrickson (approved):

PR #2540, opened 2024-01-30
Clarify local development login flow

jeffsmohan:
## Description
Improves the new developer experience my making the standard login flow a bit clearer. It does this by: 

1. Creating generic admin and staff users in the DB seed script
2. Making the logs for the generated login URL stand out more
3. Update the docker setup documentation to walk through standard login flow in dev
4. Update the gmail setup documentation with a warning that you generally shouldn't need it

(This PR is based on the discussion here: https://github.com/usdigitalresponse/usdr-gost/pull/2524/files#r1470332892)

## Screenshots / Demo Video

Screenshot 2024-01-30 at 11 48 21 AM

Screenshot 2024-01-30 at 12 03 32 PM

## Testing
Manually ran the DB seed script and login flow for the admin and staff users

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/api/users.test.js
+++ b/packages/server/__tests__/api/users.test.js
@@ -158,7 +158,7 @@ describe('`/api/users` endpoint', () => {
                 const response = await fetchApi(`/users`, agencies.own, fetchOptions.admin);
                 expect(response.statusText).to.equal('OK');
                 const json = await response.json();
-                expect(json.length).to.equal(12);
+                expect(json.length).to.equal(14);
             });
             it('lists users for an agency outside this user\'s hierarchy but in the same tenant', async () => {
                 const response = await fetchApi(`/users`, agencies.offLimits, fetchOptions.admin);
--- a/packages/server/__tests__/api/users.test.js
+++ b/packages/server/__tests__/api/users.test.js
@@ -1,6 +1,6 @@
 const { expect } = require('chai');
 const sinon = require('sinon');
-const { getSessionCookie, makeTestServer } = require('./utils');
+const { getSessionCookie, makeTestServer, knex } = require('./utils');
 const email = require('../../src/lib/email');
 const emailConstants = require('../../src/lib/email/constants');
 
@@ -158,7 +158,11 @@ describe('`/api/users` endpoint', () => {
                 const response = await fetchApi(`/users`, agencies.own, fetchOptions.admin);
                 expect(response.statusText).to.equal('OK');
                 const json = await response.json();
-                expect(json.length).to.equal(14);
+                const ownAgency = await knex('agencies').where({ id: agencies.own }).first();
+                const expectedUserCount = (
+                    await knex('users').where({ tenant_id: ownAgency.tenant_id }).count().first()
+                ).count;
+                expect(json.length).to.equal(parseInt(expectedUserCount, 10));
             });
             it('lists users for an agency outside this user\'s hierarchy but in the same tenant', async () => {
                 const response = await fetchApi(`/users`, agencies.offLimits, fetchOptions.admin);
--- a/packages/server/src/lib/email.js
+++ b/packages/server/src/lib/email.js
@@ -77,7 +77,11 @@ function sendPassCode(email, passcode, httpOrigin, redirectTo) {
     );
 
     if (process.env.DEV_LOGIN_LINK && process.env.NODE_ENV === 'development') {
-        console.log(`Login link generated: \x1b[32m${href}`);
+        const BLUE = '\x1b[34m';
+        const message = `| Login link generated: ${href} |`;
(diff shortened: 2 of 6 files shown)


packages/server/src/lib/email.js
```diff
@@ -77,7 +77,11 @@ function sendPassCode(email, passcode, httpOrigin, redirectTo) {
     );
 
     if (process.env.DEV_LOGIN_LINK && process.env.NODE_ENV === 'development') {
-        console.log(`Login link generated: \x1b[32m${href}`);
+        const BLUE = '\x1b[34m';
```
jeffsmohan, line 80: I wasn't sure if the preference was for something like this vs using a library like chalk. We seem to have chalk installed but only as a dev dependency, so I didn't want to rely on it here, but I'm happy to update if that feels cleaner.

jeffsmohan:
I've updated the seed admin and staff emails to use `example.com` for domain so we don't accidentally email all USDR staff. (Whoops! Since I had email sending via gmail set up still, it seems I actually emailed a login email to a real USDR google group address.)

packages/server/src/lib/email.js
TylerHendrickson: 🤔 Given that this execution path is specific to dev workflows, is there a way to use chalk without declaring it as a production dependency? 

In general, my preference for logging is to slowly migrate over to structured logging (we currently use `bunyan` for this, with a preconfigured base logger [exported by `lib/logging.js`](https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/server/src/lib/logging.js#L16-L21)). However, given that this is specifically intended to be human-readable and not something we would try to query with a tool like Datadog, it may be best to leave the colorization as-is for now (assuming there isn't a way to use chalk here without bringing it in as a prod dependency).

packages/server/__tests__/api/users.test.js
```diff
@@ -158,7 +158,7 @@ describe('`/api/users` endpoint', () => {
                 const response = await fetchApi(`/users`, agencies.own, fetchOptions.admin);
                 expect(response.statusText).to.equal('OK');
                 const json = await response.json();
-                expect(json.length).to.equal(12);
+                expect(json.length).to.equal(14);
```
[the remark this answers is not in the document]
TylerHendrickson: Agreed. Maybe something like this instead?
```suggestion
                const ownAgency = await knex('agencies').where({ id: agencies.own }).first();
                const expectedUserCount = (
                    await knex('users').where({ tenant_id: ownAgency.tenant_id }).count().first()
                ).count;
                expect(json.length).to.equal(parseInt(expectedUserCount, 10));
```

Note: To make the above work, you'll also have to modify line 3 to import `knex` from `./utils`, e.g.
```javascript
const { getSessionCookie, makeTestServer, knex } = require('./utils');
```

packages/server/src/lib/email.js
jeffsmohan: It's possible to use chalk here from devDependencies, but I think it ends up more messy than worth it. You run into a few issues you have to skirt: 

1. The airbnb base eslint config we use disallows devDependencies being used outside test and config files (https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L71-L95), so we'd have to either change our eslint config or ignore the rule for this line of code. 
2. We're using chalk v5, which moved to ESM, meaning we can't `require('chalk')` without downgrading to v4 or using a pretty hacky-feeling dynamic import workaround. 

Here's basically what that code would look like:

```node
    if (process.env.DEV_LOGIN_LINK && process.env.NODE_ENV === 'development') {
        // eslint-disable-next-line import/no-extraneous-dependencies
        import('chalk').then((chalkModule) => {
            const chalk = chalkModule.default;
            const message = `| Login link generated: ${href} |`;
            console.log(chalk.blue('-'.repeat(message.length)));
            console.log(chalk.blue(message));
            console.log(chalk.blue('-'.repeat(message.length)));
        });
    }
```

Personally, I don't think it's a particular improvement. It also has the small danger of someone copy-pasting code that relies on chalk outside of the `NODE_ENV === 'development'` check and possibly breaking staging/prod. 

If this is the only instance of wanting color output for now, I'd be inclined to leave it as-is. If we start needing this a few places, I'd probably consider promoting chalk to a full dependency. Since this is server-side code, we shouldn't be as worried about bundle size, etc.

Thoughts?

jeffsmohan:
Addressed comments, and ready for re-review!

packages/server/src/lib/email.js
TylerHendrickson: Agree with the above – let's leave as-is unless/until there are any compelling use-cases for production.

TylerHendrickson (approved):

PR #2579, opened 2024-02-08
Fix up Category of Funding Activity label and add colons to table

adele-usdr:
Fix up Category of Funding Activity label and add colons to table

## Description
Follow up to fix truncated Category of Funding Activity label and add colons to the table

## Screenshots / Demo Video
Screenshot 2024-02-07 at 7 09 55 PM

## Testing
Turn on Grant Details Page feature flag and look at table to see colons in first column, and the complete Category of Funding Activity label

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [x ] Added steps to test feature/functionality manually

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/client/src/views/GrantDetails.vue
+++ b/packages/client/src/views/GrantDetails.vue
@@ -402,7 +402,4 @@ export default {
   white-space: nowrap;
   width: 1%;
 }
-.grants-details-table-fit-content:after {
-  content: ": "
-}
 </style>
--- a/packages/client/src/views/GrantDetails.vue
+++ b/packages/client/src/views/GrantDetails.vue
@@ -236,7 +236,7 @@ export default {
         name: 'Award Ceiling',
         value: this.selectedGrant.award_ceiling,
       }, {
-        name: 'Category of',
+        name: 'Category of Funding Activity',
         value: this.selectedGrant.funding_activity_categories?.join(', '),
       }, {
         name: 'Opportunity Category',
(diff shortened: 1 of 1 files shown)


packages/client/src/views/GrantDetails.vue
```diff
@@ -402,4 +402,7 @@ export default {
   white-space: nowrap;
   width: 1%;
 }
+.grants-details-table-fit-content:after {
```
jeffsmohan, line 405: Whoops, my bad, this also falls into the category of "things I tweaked from the design intentionally but forgot to call out in my review." 

I would argue the colons are extraneous and thus slightly hinder reading of the table. The point of the tabular layout is to make clear that the item in the left column describes the column in the right. That said, I don't feel strongly, so feel free to add the colons in. 

One very minor nit: If we decide we want the colons, they should probably be part of the text proper, rather than added in CSS. I haven't tested this, but I wouldn't be surprised if accessibility devices treat the CSS content differently than "real" content in the html of the page.

jeffsmohan (changes requested):
Good catch on the category label! One very minor comment on the colons, and then looks great.

packages/client/src/views/GrantDetails.vue
adele-usdr: I think you're right on that - which further supports your argument that they're extraneous :) I'll remove the colons for now!

jeffsmohan (approved):
lgtm!

PR #2608, opened 2024-02-14
Add ability to easily create/manage arpa treasury report queue locally

inaki:
## Description
The task achieved was to add two env variable in two different files that will make easier to create/manage arpa treasure report queue locally. 

## Screenshots / Demo Video
N/A

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/localstack/entrypoint/init-aws.sh
+++ b/localstack/entrypoint/init-aws.sh
@@ -18,3 +18,5 @@ awslocal s3api create-bucket --bucket arpa-audit-reports --region us-west-2 --cr
 
 awslocal sqs create-queue --queue-name grants-ingest-events
 awslocal sqs create-queue --queue-name arpa-queue
+
+awslocal sqs create-queue --queue-name arpa-treasury-queue --region us-west-2
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -49,6 +49,7 @@ AUDIT_REPORT_BUCKET=arpa-audit-reports
 # SQS queue for grants-ingest events and arpa events
 GRANTS_INGEST_EVENTS_QUEUE_URL="http://localhost:4566/000000000000/grants-ingest-events"
 ARPA_AUDIT_REPORT_SQS_QUEUE_URL="http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/arpa-queue"
+ARPA_TREASURY_REPORT_SQS_QUEUE_URL="http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/arpa-treasury-queue"
 
 # for setting up the worker processes to run locally over localstack
 TASK_QUEUE_URL="http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/arpa-queue"


inaki:
I'll make sure to follow the branch naming convention for next PRs :)

tzinckgraf:
@inaki this looks great so far, but we might have to update this a little bit more.

How do you plan on running the workers?

Looking at https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/server/src/scripts/arpaTreasuryReport.js, it looks like it wants this `process.env.TASK_QUEUE_URL`, which as of now is set to the audit report.

I also added some runners in `package.json` to make running the process locally a little easier. It may be worth adding that for the treasury report workers as well.

TylerHendrickson:
@inaki @tzinckgraf Adding a few thoughts on the path forward for this type of setup...

First, maybe `TASK_QUEUE_URL` doesn't need to be set in `packages/server/.env` since it's a common directive that each async task worker expects to point to whatever queue it should be polling to find new tasks; it could just as well be passed to a task runner as a CLI flag like `--task-queue-url=http://sqs.region.localstack-hostname:4566/000000000000/queue`. For that matter, I'm actually not aware of anything that ensures env vars from `packages/server/.env` are sourced when running any of the async task workers that initialize from `packages/server/src/scripts` – I don't see any usages of `dotenv` in `arpaAuditReport.js` or `arpaTreasuryReport.js`, for example. (Note: this is different with `docker compose` setups – see below). So there's a general point here that currently, there are no environment variables (including `TASK_QUEUE_URL`) inherently deriving from `packages/server/.env` when a script is run.

Second, although setting the queue URL via a common environment variable name is useful for instrumentation and deployment, it doesn't have to be the (only) means by which the queue URL is configured for a specific task worker. Again, we could use a CLI flag, or we could have a "higher weight" environment variable that we use in dev environments, e.g.
```js
// async function main() in packages/server/src/scripts/arpaAuditReport.js ...
const queueUrl = process.env.AUDIT_REPORT_TASK_QUEUE_URL || process.env.TASK_QUEUE_URL;
```
and expect that `AUDIT_REPORT_TASK_QUEUE_URL` is only used in non-dev environments.

Third, we could lean into what @tzinckgraf configured for the `start:arpa-audit-worker` and `debug:arpa-audit-worker` scripts in `packages/server/package.json` and rely on that mechanism to ensure that those scripts are properly configured when run in a development environment.

Finally, as mentioned earlier, there are `docker compose` workflows to consider and potentially build a solution around. Our `docker-compose.yml` file configures the `app` service with `env_file: ['./packages/server/.env']`, which does mean that anything run in the `app` service container loads environment variables from that file (regardless of whether a tool like `dotenv` is in use). Another (and not mutually-exclusive) approach here would be to define task workers as new services in `docker-compose.yml`. 

----

To offer something concrete: from my perspective, we should have an easy way for developers to run these workers with necessary configuration options (including/especially environment variables) by default. I feel like this would work best by continuing with the `package.json` scripts convention that @tzinckgraf mentioned earlier, but adding a bit more instrumentation to ensure that, at minimum, default environment variables are sourced from `packages/server/.env`, and that `TASK_QUEUE_URL` was set explicitly per-worker when executing the script. I could see that working like this:
1. In `packages/server/.env.example`, set per-worker task queue URLs, e.g. `AUDIT_REPORT_TASK_QUEUE_URL=https://...`
2. In `packages/server/package.json` modify `start:arpa-audit-worker` to do the following
    1. Something (maybe using `dotenv`?) that sources environment variables from `.env`
    2. Runs `TASK_QUEUE_URL=$AUDIT_REPORT_TASK_QUEUE_URL node src/scripts/arpaAuditReport.js`
3. Repeat step 2 for `debug:arpa-audit-worker`
4. Add `start:arpa-treasury-

TylerHendrickson:
PS- I clicked the wrong button; sorry for closing & reopening the PR!

as1729:
Closing as the requirements have changed.

PR #2612, opened 2024-02-15
updating left and right margins and right column width

adele-usdr:
## Description
updating left and right margins and right column width

## Screenshots / Demo Video
Before:
Screenshot 2024-02-14 at 7 17 32 PM

After:
Screenshot 2024-02-14 at 7 17 03 PM

### DIFF
--- a/packages/client/src/views/GrantDetails.vue
+++ b/packages/client/src/views/GrantDetails.vue
@@ -11,7 +11,7 @@
         <b-row>
 
           <!-- Left page column: title, table data, and grant description -->
-          <b-col>
+          <b-col class="ml-5">
             <h2 class="mb-5">{{ selectedGrant.title }}</h2>
             <b-table
               class="grant-details-table mb-5"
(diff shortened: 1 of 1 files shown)


packages/client/src/views/GrantDetails.vue
```diff
@@ -11,7 +11,7 @@
         <b-row>
 
           <!-- Left page column: title, table data, and grant description -->
-          <b-col>
+          <b-col class="ml-5">
```
jeffsmohan, line 14: I think these margins are a bit cleaner to add/maintain if you add them as a `mx-5` on the top-level `section` tag in the component. (And you can remove the `container-fluid` class while you're there... I should have removed it when I did the design touch-ups, as it's redundant. If you're feeling extra, I'm realizing you could also remove the extraneous `div` on L3, since it's purely vestigial.)

adele-usdr: good call! pushed a new commit for these suggestions. I ended up doing m-5 since having a bottom margin looks nice too.

jeffsmohan (approved):
nice!

PR #2677, opened 2024-02-28
Fix tflint warnings

jeffsmohan:
## Description
Github PRs show outstanding check failures on every single PR's diff, even though the failures are global and unrelated to the change. This is added noise that devs shouldn't have to think about. This PR attempts to fix all the outstanding `tflint` issues to eliminate the noise, and then enforce warning-level linting compliance in CI. 

image

Note: I was hoping to add a `pre-commit` hook (e.g., https://github.com/antonbabenko/pre-commit-terraform) to run `tflint` on any changed `.tf` files. (This would improve DX when editing those files, as you wouldn't have to wait for CI to hear about your issues.) However, it turns out what I thought was [`pre-commit`](https://pre-commit.com/) is actually just a package.json alias to running eslint, so this seems out of scope for now. 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


terraform/modules/gost_website/variables.tf
```diff
@@ -14,12 +14,6 @@ variable "tags" {
   default = {}
 }
 
-variable "permissions_boundary_arn" {
-  description = "ARN of the managed policy to set as the permissions boundary for all roles."
-  type        = string
-  default     = null
-}
```
TylerHendrickson, line 21: This deletion is the reason that `terraform validate` is failing. The [error message](https://github.com/usdigitalresponse/usdr-gost/actions/runs/8075730432/job/22063163214?pr=2677#step:14:19),
> ```
> Error: Unsupported argument
>   on main.tf line 62, in module "website":
>   62:   permissions_boundary_arn = local.permissions_boundary_arn
>
> An argument named "permissions_boundary_arn" is not expected here.
> ```

refers to the presence of a `permissions_boundary_arn` in [`main.tf`'s configuration of the `website` module](https://github.com/usdigitalresponse/usdr-gost/blob/57240edfa5a5fb3a79d74b2f54c3d9edf82336ec/terraform/main.tf#L58-L85), but the module's source directory (`terraform/modules/gost_website`) has no `variable` block that that declares an input variable named `permissions_boundary_arn`.

I'm guessing that this got removed because the `terraform/modules/gost_website` was defining a variable (`permissions_boundary_arn`) that wasn't actually used anywhere. Since that's the case, the simple fix for this error would be to delete [Line 62 in `terraform/main.tf`](https://github.com/usdigitalresponse/usdr-gost/blob/57240edfa5a5fb3a79d74b2f54c3d9edf82336ec/terraform/main.tf#L62):
```tf
  permissions_boundary_arn = local.permissions_boundary_arn
```

**However,** a few things to consider:
- I tend to declare this variable as a convention when creating modules, because our security controls in AWS enforce that all IAM roles defined by continuous deployment (that is, any IAM role defined in this repo's `terraform/` directory) be configured with a standard permissions boundary, which explicitly limits the scope of the role's permissions.
  - This is useful for open-source projects like this one because it acts as a safeguard against code changes that could otherwise introduce an overly-permissive role, which not every reviewer might be able to recognize.
- The `terraform/modules/gost_website` module is the _only_ one that doesn't define at least one IAM role. 
- If this module was ever updated to define a new IAM role, we'd need to re-add this variable so that it could be passed to the role definition. Otherwise, Terraform would fail because AWS would reject the role as noncompliant with our security posture.
- Because this particular convention is so strict, I've typically taken the stance that it's better to include an unused (but standard) variable so that if/when it ends up being required, the solution is (hopefully) more obvious as a standard convention.

As the designer of this solution (and someone with a lot of Terraform experience), I don't feel like I'm in the best position to assume what will be (non-)obvious to less-experienced contributors. **If the points listed above don't feel very compelling to people who aren't me, then keeping the unused variable definition around is indeed just code bloat** – meaning we should delete this code block (as it is currently in this PR) and the offending line in `terraform/main.tf` mentioned above. If we think it's worthing keep this convention in-place, the alternative would be to re-add this block and add a `tflint-ignore` pragma for the unused variable, which would look like this:
```tf
# tflint-ignore: terraform_unused_declarations
variable "permissions_boundary_arn" {
  description = "ARN of the managed policy to set as the permissions boundary for all roles."
  type        = string
  default     = null
}
```

jeffsmohan: Aha, of course, I should've realized! The reason I couldn't grep for it in my editor is because I had deleted it for the linter. 

Anyway, as a very new terraform user, everything in terraform is non-obvious. :P That said, I'm on board with your reasoning for keeping the variable in place even where it's unused, to give us the best chance of keeping things properly secured. I'm on board with the ignore pragma, combined with a little explainer comment summarizing the reasoning. I'll put a commit together. 

Thanks for bringing me up to speed on the context!

jeffsmohan: Btw, maybe I'm misunderstanding what happened here, but it feels concerning that our linter told me to remove an unused variable that apparently WAS actually used elsewhere. As a general rule, I really don't want my autofix-able lint rules to be able to introduce bugs and break my systems. Any idea why that seems to be happening here?

jeffsmohan:
@TylerHendrickson ok, I think this is in good shape for a proper review now, thanks!

jeffsmohan:
@TylerHendrickson ok, I think this is in good shape for a proper review now, thanks!

terraform/modules/gost_website/variables.tf
TylerHendrickson: It seems that the linter inspected the `gost-website` module and found a variable declaration that wasn't used anywhere in the module, and therefore (in absence of any override) should be removed.

However, on the _caller_ side, in `terraform/main.tf`, the input for that variable was still being provided. It seems that tflint isn't smart enough to remove input variables at the same time (I believe this is because it inspects modules independently of their upstream usages).

A bit of a pseudocode example in case it's helpful:
```js
function addTwo(a, b, c) {  // <-- Linter detects that variable `c` is unused in this function
  return a + b; 
}

function myCallingFunction() {
    const sum = addTwo(1, 2, 3);  // <-- For some reason, linter will not fix this invalid usage
    console.log(`The sum is ${sum}`);
}
```

I'll agree that it's vexing that `tflint` would not complain about there being an input for an undeclared variable – I just tested by removing the variable declaration on my local environment. However, it might be worth noting that the [`hashicorp.terraform` VSCode extension](https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform) was able to catch the issue:
image

Since `hashicorp.terraform` is being added to the list of recommended extensions in #2703, that should at least make situations like these a bit less painful and more obvious to developers in the future.

TylerHendrickson (approved):
Thanks for your help on this- looks great!

terraform/modules/gost_website/variables.tf
jeffsmohan: Ahh, thanks for the explainer, I see the issue now. Yeah, I think if we've got coverage of the issue in VSCode's autoformatting/linting, plus it's obviously getting caught in CI by validate, that should cover the bases here.

Issue #2572, opened 2024-02-06
[spike] Investigate options to mitigate v-html vulnerability


jeffsmohan:
### Why is this issue important?

We currently display grant description text on the site using `v-html`, which inserts the contents as raw html. This can open us up to [XSS attacks](https://en.wikipedia.org/wiki/Cross-site_scripting) unless we properly sanitize the text first. Level of concern is not super high, since we should be able to trust the source of the data, but it's still well worth fixing the security hole. 

### Current State

We use [`v-html`](https://vuejs.org/guide/essentials/template-syntax.html#raw-html) in a the [legacy and new grant details modal/page](https://github.com/search?q=repo%3Ausdigitalresponse%2Fusdr-gost%20v-html&type=code) to display the details. 

The description text we get from our data sources is not especially consistent/clean, so it sometimes contains HTML entities (e.g., `&amp;`) or tags (e.g., `<p>` or `<br>`) that are useful in displaying the data properly. However, they can't be counted on and are sometimes double-encoded (e.g., `&amp;#8226;` for a bullet). 

### Expected State

We should find a better way to display the content with good readability and without any security vulnerability. 

### Implementation Plan

This ticket is a spike to investigate possible approaches. An initial non-exhaustive set of ideas to explore based on my conversation with @TylerHendrickson: 

- Use a library to sanitize the description text (either allow-list or strip out any HTML) and then continue to use `v-html` on this "safe" content
- Convert the description text to something like markdown in the data pipeline, then render that on the page
- Integrate with an LLM to help translate the messy description text input into a well-formatted output

jeffsmohan:
Note that an analogous vulnerability exists in our email rendering where we render un-escaped grant description text: https://github.com/usdigitalresponse/usdr-gost/blob/9e903c3923a01bff87cc7d49a0f69236a1ff57a0/packages/server/src/static/email_templates/_grant_detail.html#L34

PR #2705, opened 2024-03-01
Update emails to point to new Grant Details page

jeffsmohan:
## Description
Updates the digest and assigned emails to link to our new Grant Details page when the flag is on, instead of directly to grants.gov. 

## Screenshots / Demo Video
Screenshot 2024-03-01 at 12 01 12 PM
Screenshot 2024-03-01 at 12 04 17 PM

## Testing
Testing this in dev is a lot of pretty fiddly effort, but here are the steps: 

1. Set up so emails actually get sent/captured in dev
   1. Create an ethereal email to work with: https://ethereal.email/ (nice, one-click creation of an ephemeral email account)
   2. Update the `NODEMAILER_*` environment variables in your `packages/server/.env` file from the values listed on your ethereal account
   3. Comment out `secure: true` in our nodemailer setup (Gmail uses a secure connection, but ethereal doesn't): https://github.com/usdigitalresponse/usdr-gost/blob/05f688e75df1e3fbf6698f88b49281dd117eb280/packages/server/src/lib/email/email-nodemailer.js#L36
   4. Rebuild your docker container so it picks up the new env variables (`docker compose down app` and `docker compose up -d`)
2. Trigger the emails manually via a node REPL
   - Digest email
     1. `docker compose exec app yarn node`
     2. ```node
           const emailService = require('./packages/server/src/lib/email');
           const knex = require('./packages/server/src/db/connection')
           const grants = await knex('grants').limit(3)
           emailService.sendGrantDigest({name: 'Test agency', matchedGrants: grants, recipients: ['test@example.com'], openDate: '2024-01-01'})
           ```
   - Assignment email
     1. Comment out the `assignee` variable in `sendGrantAssignedNotficationForAgency` and replace it with `const assignees = [{ email: 'test@example.com' }];`: https://github.com/usdigitalresponse/usdr-gost/blob/05f688e75df1e3fbf6698f88b49281dd117eb280/packages/server/src/lib/email.js#L221
     2. `docker compose exec app yarn node`
     3. ```node
           const emailService = require('./packages/server/src/lib/email');
           const knex = require('./packages/server/src/db/connection')
           const grant = await knex('grants').first();
           emailService.sendGrantAssignedEmail({ grantId: grant.grant_id, agencyIds: [1], userId: 1 });
           ```
3. Once the test email is sent, go to your "Messages" tab in your Ethereal mail account, and you should see the sent email listed there for viewing

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


inaki (approved):
In order test the rebuild was not enough, but it works if you delete docker instance and setup a new instance with the env variable and the testing changes.

inaki:
Also looks good in mobile.

Screenshot 2024-03-05 at 4 00 59 PM

jeffsmohan:
@TylerHendrickson Even though this has already been reviewed and approved, I'd appreciate your eyes on the code too, particularly the staging/prod flags.

packages/server/src/lib/email.js
```diff
@@ -218,10 +236,10 @@ async function sendGrantAssignedNotficationForAgency(assignee_agency, grantDetai
     // TODO: add plain text version of the email
     const emailPlain = emailHTML.replace(/<[^>]+>/g, '');
     const emailSubject = `Grant Assigned to ${assignee_agency.name}`;
-    const assginees = await db.getSubscribersForNotification(assignee_agency.id, notificationType.grantAssignment);
+    const assignees = await db.getSubscribersForNotification(assignee_agency.id, notificationType.grantAssignment);
 
     const inputs = [];
-    assginees.forEach((assignee) => inputs.push(
+    assignees.forEach((assignee) => inputs.push(
```
TylerHendrickson, line 242: Good catch 🎩

packages/server/src/static/email_templates/_grant_detail.html
```diff
@@ -31,7 +31,7 @@
                                                 style="Margin:0;padding-bottom:16px;">
                                                 <p
                                                     style="Margin:0;-webkit-text-size-adjust:none;-ms-text-size-adjust:none;mso-line-height-rule:exactly;font-family:arial, 'helvetica neue', helvetica, sans-serif;line-height:24px;color:#000000;font-size:16px">
-                                                    {{{description}}}<span>... View on <a style="color:#0068D6;text-decoration:none;" href="{{link_url}}">Grants.gov</a></span>
+                                                    {{{description}}}<span>... View on <a style="color:#0068D6;text-decoration:none;" href="{{link_url}}">{{ link_description }}</a></span>
```
TylerHendrickson, line 34: Noticed that the `href` values are being URL-encoded...
```suggestion
                                                    {{{description}}}<span>... View on <a style="color:#0068D6;text-decoration:none;" href="{{& link_url }}">{{ link_description }}</a></span>
```
Would you mind updating this file to include `&` on any other `href` values provided by mustache?

TylerHendrickson (changes requested):
Requesting a quick tweak, otherwise looks good!

packages/server/src/static/email_templates/_grant_detail.html
jeffsmohan: Good catch! 

I'm trying to think of a good way to prevent opening ourselves up to an injection attack here. (I think email clients by and large have a lot of guardrails to prevent stray scripts anyway, but I'd still prefer to avoid the issue on our end.) I think for now I'll adopt the convention of appending `_safe` to a variable name to indicate it's known to be safe, and escape the value from the DB when building the template variables. But if you have another preference here, let me know!

I'm noting we have the same tougher-to-solve issue with rendering the grant description un-escaped here as well, in addition to our use of `v-html` in frontend code. I've added a note to the ticket tracking that issue: https://github.com/usdigitalresponse/usdr-gost/issues/2572#issuecomment-1986096350

jeffsmohan:
@TylerHendrickson Thanks for the review! I've addressed the URL escaping issue, and I also fixed another bug I found with an incorrect utm_campaign value.

packages/server/src/static/email_templates/_grant_detail.html
TylerHendrickson: This solution looks good to me!

TylerHendrickson (approved):

PR #2710, opened 2024-03-04
[issue-2369] Back or Browse Grants Link from Grant Details

inaki:
## Description
This PR implements navigation links in the Grant Details page as per the provided design specifications and user navigation scenarios.

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/485547/b6c3c124-d544-4dca-984c-1416a3112f48

### Testing
1. Verify Navigation from Grants Table to Grant Details:
    - Click on a grant item in the Grants Table component.
    - Ensure that you are redirected to the Grant Details page.
    - Confirm that the URL includes the grant ID as a parameter.

2. Check Back Link Behavior:
    - Visit the Grant Details page from various parts of the application (browse grants and my grant).
    - Click on the "Back" link above the grant title.
    - Verify that you are redirected to the previous page you were on.
    - If the previous page was saved search results, confirm that you return to the specific saved search results.
    - If the previous page was a tab within the My Grants page, ensure that you return to the same tab.
    - Compare the behavior with clicking the browser's back button.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/client/src/components/GrantsTable.vue
```diff
@@ -440,7 +440,7 @@ export default {
       if (!newGrantsDetailPageEnabled()) {
         return;
       }
-      this.$router.push({ name: 'grantDetail', params: { id: item.grant_id } });
+      this.$router.push({ name: 'grantDetail', params: { id: item.grant_id }, query: { fromApp: true } });
```
jeffsmohan, line 443: We shouldn't need to surface this "from app" state in the URL where it's extra user-visible clutter. Instead, I would set up a route guard in the GrantDetails component to check the `from` route. So it can check if there's a `from` parameter, and if so, set a flag in the component data appropriately. See the docs here: https://router.vuejs.org/guide/advanced/navigation-guards.html#In-Component-Guards

packages/client/src/views/GrantDetails.vue
```diff
@@ -1,5 +1,11 @@
 <template>
   <section>
+      <div class="grant-details-back-link" style="display: flex; justify-content: start" v-if="showBackLink">
+          <a @click="goBack" class="back-link">Back</a>
```
jeffsmohan, line 4: Unless I'm missing it, this class is not used anywhere, so let's remove it.

packages/client/src/views/GrantDetails.vue
```diff
@@ -434,11 +443,19 @@ export default {
     selectableOption(option) {
       return option.status_code !== HEADER;
     },
+    goBack() {
+      this.$router.go(-1);
```
jeffsmohan, line 447: This is a bit of personal preference, so feel free to ignore if you prefer this setup, but I tend to prefer inlining really simple one-liners instead of adding the indirection of a method like this. So like `@click="$router.back()` instead of `@click="goBack"`.

jeffsmohan (changes requested):
Just a few minor fixes, and it should be ready to go!

packages/client/src/views/GrantDetails.vue
```diff
@@ -434,11 +443,19 @@ export default {
...
   },
 };
 </script>
 
 <style lang="css">
+.grant-details-back-link {
+position: absolute;
+left: 11.5rem;
+top: 112px;
+}
```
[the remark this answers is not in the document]
inaki: I changed this to use similar code as the .grant-deatils-container. Although a better solution would be to have layout classes everywhere for this cases.

packages/client/src/components/GrantsTable.vue
inaki: For some reason the route guard is not working. I tried the from argument and try to passed this to the `next({ source: <from code> })` but it did not worked.

packages/client/src/views/GrantDetails.vue
```diff
@@ -1,5 +1,11 @@
 <template>
   <section>
+      <div class="grant-details-back-link" style="display: flex; justify-content: start" v-if="showBackLink">
```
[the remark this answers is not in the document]
inaki: Agreed this is removed.

packages/client/src/views/GrantDetails.vue
inaki: yes, that was use but not anymore, is romoved!

packages/client/src/components/GrantsTable.vue
jeffsmohan: I had expected there to be a cleaner way to achieve this (i.e., `this.$router` surfacing the actual page history as a list), but alas, doesn't seem like it. I was able to get it working with something like this, though: 

```diff
diff --git a/packages/client/src/views/GrantDetails.vue b/packages/client/src/views/GrantDetails.vue
index 5b4f3100..d62c8bb5 100644
--- a/packages/client/src/views/GrantDetails.vue
+++ b/packages/client/src/views/GrantDetails.vue
@@ -173,6 +173,7 @@ export default {
   },
   data() {
     return {
+      isFirstPageLoad: false,
       showDialog: false,
       orderBy: '',
       assignedAgenciesFields: [
@@ -200,6 +201,14 @@ export default {
       copyUrlSuccessTimeout: null,
     };
   },
+  beforeRouteEnter(to, from, next) {
+    const isFirstPageLoad = from.name === null && from.path === '/';
+    next((vm) => {
+      if (isFirstPageLoad) {
+        vm.setFirstPageLoad();
+      }
+    });
+  },
   created() {
     // watch the params of the route to fetch the data again
     this.$watch(
@@ -349,6 +358,9 @@ export default {
     debounceSearchInput: debounce(function bounce(newVal) {
       this.debouncedSearchInput = newVal;
     }, 500),
+    setFirstPageLoad() {
+      this.isFirstPageLoad = true;
+    },
     async markGrantAsViewed() {
       await this.markGrantAsViewedAction({ grantId: this.selectedGrant.grant_id, agencyId: this.selectedAgencyId });
     },
```

packages/client/src/views/GrantDetails.vue
```diff
@@ -1,5 +1,13 @@
 <template>
   <section>
+    <div class="container-fluid">
```
jeffsmohan, line 3: Sorry, I don't know if I was clear enough in my last comment. I think the HTML and CSS can be much cleaner if we utilize the existing CSS grid setup on the [`<div class="grant-details-container">` div](https://github.com/usdigitalresponse/usdr-gost/pull/2710/files#diff-456776fc1d1394d51d55789b5ae9f202b288850b0586d36f58f3c4330bd44a9eR18). 

Specifically: 

1. We should no longer need the `<div class="container-fluid">` div on this line. 
2. The two `<div class="grant-details-back-link">` elements would move inside the `<div class="grant-details-container">`. 
3. The CSS edits you have right now (e.g., messing with the padding) can be removed. 
4. Instead, you'd just need to update the grid with something like this: 

```diff
diff --git a/packages/client/src/views/GrantDetails.vue b/packages/client/src/views/GrantDetails.vue
index 5b4f3100..71009e9c 100644
--- a/packages/client/src/views/GrantDetails.vue
+++ b/packages/client/src/views/GrantDetails.vue
@@ -445,11 +445,15 @@ export default {
   grid-template-columns: 1fr 437px;
   grid-template-rows: auto;
   grid-template-areas:
+    "back-link back-link"
     "headline  main-actions"
     "content   secondary-actions";
   column-gap: 90px;
   row-gap: 48px;
 }
+.grant-details-back-link {
+  grid-area: back-link;
+}
 .grant-details-headline {
   grid-area: headline;
   align-self: end;
```

Hopefully that makes it clearer what I had in mind, but let me know if it'd be helpful to discuss live!

jeffsmohan (changes requested):
Thanks for the fixes! Couple more quick comments.

packages/client/src/views/GrantDetails.vue
inaki: Include the back inside the container and the grid. I was thinking is there any situation where the details content doesn't render we won't have a back button. What you think?

jeffsmohan (approved):
This lgtm, you just need to fix the outstanding linting error and this should be good to go

jeffsmohan (approved):

PR #2725, opened 2024-03-06
Fix for grant assign email with missing description

jeffsmohan:
## Description
When sending the grant assignment email, issues sending the email can bubble up and cause the node server to crash. Instead, we want to gracefully catch these errors and return a 500 response. Also fixes the proximate cause of the error here, which was a missing description for the grant. Specifically, this PR: 

1. Fixes the grant detail rendering so it renders the grant even when the description is missing
2. Email utils for the grant assign email now catch errors building/sending the email and logs the error
3. API endpoint for sending the grant assign email now catches errors and gracefully returns a 500

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/src/routes/grants.js
```diff
@@ -348,7 +348,12 @@ router.put('/:grantId/assign/agencies', requireUser, async (req, res) => {
     }
 
     await db.assignGrantsToAgencies({ grantId, agencyIds, userId: user.id });
-    email.sendGrantAssignedEmail({ grantId, agencyIds, userId: user.id });
+    try {
+        email.sendGrantAssignedEmail({ grantId, agencyIds, userId: user.id });
```
as1729, line 352: Should this be `await email.sendGrantAssignedEmail` to ensure that it waits for the function to complete to catch the error?

jeffsmohan: Yes, good catch! And that probably also explains why the second test I was trying to write wasn't working... let me see if I can get a test for the 500 response as well

jeffsmohan: Ugh, I think the test is unfortunately not worth the effort (unless you have a quick read on how to get it working). I was trying to add something like this to the `PUT /api/grants/:grantId/assign/agencies` tests in `__tests__/api/grants.test.js`: 

```node
            it('gracefully returns 500 response for unexpected email errors', async () => {
                const emailSpy = sandbox.spy(email, 'sendGrantAssignedNotficationForAgency');
                sandbox.stub(email, 'buildGrantDetail').throws();
                const response = await fetchApi(`/grants/${assignEndpoint}`, agencies.own, {
                    ...fetchOptions.admin,
                    method: 'put',
                    body: JSON.stringify({ agencyIds: [agencies.own] }),
                });
                expect(response.statusText).to.equal('Internal Server Error');
                expect(emailSpy.notCalled).to.equal(true);
            });
```

I believe the issue is that our existing tools for stubbing and throwing don't have access to stub out the method that's getting called, since it's really more of an integration test than a unit test here. (Otherwise, we could use `rewire` or `sinon` to mock out the function the code is calling.) Anyway, I'm going to move on, but if you have the answer here, I'd love to hear it!

TylerHendrickson (approved):
Chiming in to say I think this is okay to merge without testing the 500 scenario explicitly 🙂  

_Side note..._
Just my 2¢, but I'm not entirely convinced the user/client needs to be made aware of the error in this case. If we had a paradigm where this type of thing rolled back the DB change (which would be nice, but definitely out-of-scope), it might be justified in giving the user a prompt to retry the failed operation, but given that the assignment has already occurred, I'm not sure what recourse the user really has when presented with "500 internal server error". 

If there's a way to avoid swallowing the error without making it the user's problem, that feels better to me in the context of this particular situation. Better yet, it would be great if there was a more generic way to configure Express so that a request/response loop's unhandled error didn't bubble up to the point of crashing the process. 

To be clear, I think this PR is okay to ship as-is; just take this as food-for-thought.

jeffsmohan:
> If there's a way to avoid swallowing the error without making it the user's problem, that feels better to me in the context of this particular situation. Better yet, it would be great if there was a more generic way to configure Express so that a request/response loop's unhandled error didn't bubble up to the point of crashing the process.

Yeah totally! In general I like to push email sending into a task queue outside of the main API's request-response flow. Emails can be flaky or slow to send, so this gives us retries and keeps the send time from affecting API response latency. In general, if retries don't solve the issue, then there's probably nothing the user can do anyway, so it's not that helpful to surface the error. 

I could see a broader refactoring of the email machinery to make it easier to catch errors at the right places, attempt retries, and handle it all on a message queue. But obviously that's beyond the scope for this little PR. :)

jeffsmohan:
@TylerHendrickson had to fix a merge conflict. Please re-approve when you get a chance!

PR #2735, opened 2024-03-07
Prevent cursor from remaining in v-select component after selection

jeffsmohan:
## Description
Once a user has selected a option in either of the v-select boxes on the Grant Details page, we want to ensure we blur the input so the cursor doesn't continue blinking confusingly. Instead, this PR focuses the submit button, since that's what the user is likely to want to do next and enables quick and easy keyboard navigation. 

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/126250/3d347047-1a63-40b8-92db-adf83d0b5669

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


inaki:
@jeffsmohan I test keyboard navigation and interaction and it works fine. Although, I can not delete or change the team and team status. Is this out of the scope of this PR?

jeffsmohan:
> @jeffsmohan I test keyboard navigation and interaction and it works fine. Although, I can not delete or change the team and team status. Is this out of the scope of this PR?

@inaki Can you clarify the problem you're seeing? Maybe steps to repro or a video demo? I'm able to remove or update the team status just fine. Maybe you're logged in as a user without the proper permissions to update that field?

inaki:
Uploading rev.mov…

inaki:
If I delete the selected team then I can't choose again another one.

inaki:
I think it did not upload before

https://github.com/usdigitalresponse/usdr-gost/assets/485547/9c4b36a8-1f23-4e36-88d7-98ca10b8289a

jeffsmohan:
@inaki Thanks for recording the video, that's helpful! When I follow the same set of steps you did, the site works correctly for me, both on main and this feature branch. 

I wonder if your app process is crashing at some point in that flow because you've got different nodemailer setup in `.env` from working on emails? (There's a bugfix PR'd right now to make the app resilient to an error of this type: https://github.com/usdigitalresponse/usdr-gost/pull/2725)

Anything from the logs to help explain? 

Regardless, I'm pretty sure this is unrelated to the change in this PR, so I don't think it needs to be blocking here. Though, if we're able to reproduce your issues elsewhere, then those are definitely worth ticketing up separately.

PR #2736, opened 2024-03-08
Change back link text from `Back to Results` to `Back`

inaki:
## Description
- [x] Change `Back to Results` text to `Back` on the GrantDetails view. 
- [x] Move scss directory inside `src` in order to be able to import using relative path `@/scss/...` instead of `../../scss`
- [x] Corrected `App.vue` && `SearchFilter.vue` files to use the relative path.
- [x] Add semantic token called `link-back`  

## Screenshots / Demo Video
Screenshot 2024-03-08 at 1 18 48 AM

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/client/src/App.vue
```diff
@@ -14,7 +14,7 @@ export default {
 </script>
 
 <style lang="scss">
-@import '../scss/custom.scss';
+@import '@/scss/custom.scss';
```
jeffsmohan, line 17: We can now use the `@` alias for imports within the scss directory itself as well, right? Like: 

```diff
diff --git a/packages/client/src/scss/colors-semantic-tokens.scss b/packages/client/src/scss/colors-semantic-tokens.scss
index 7dd8603c..2f4d58dd 100644
--- a/packages/client/src/scss/colors-semantic-tokens.scss
+++ b/packages/client/src/scss/colors-semantic-tokens.scss
@@ -7,7 +7,7 @@
  *   https://www.figma.com/file/HWu4iIjcoX2txuN543qzIB/USDR-Design-System?type=design&node-id=8%3A2&mode=design&t=8mml86X91f9FzVAE-1
  */

-@import './colors-base-tokens.scss';
+@import '@/scss/colors-base-tokens.scss';

 // USDR BRAND
 $brand-primary: $usdr-brand-primary;
diff --git a/packages/client/src/scss/custom.scss b/packages/client/src/scss/custom.scss
index d1d5a96c..a57b128f 100644
--- a/packages/client/src/scss/custom.scss
+++ b/packages/client/src/scss/custom.scss
@@ -12,7 +12,7 @@
  */

 // IMPORT USDR DESIGN SYSTEM SEMANTIC TOKENS
-@import './colors-semantic-tokens.scss';
+@import '@/scss/colors-semantic-tokens.scss';

 // LIGHT THEME (SEMANTIC TOKEN) DEFINITION - SASS MAP
 $theme-colors: () !default;
@@ -51,7 +51,7 @@ $theme-colors: map-merge(
 @import 'vue-select/src/scss/vue-select.scss';

 // SET USDR FONTS
-@import './fonts.scss';
+@import '@/scss/fonts.scss';

 /* Dropshadow for the header */
 .header-dropshadow {
```

jeffsmohan (changes requested):
Thanks for the cleanup, this is nice! Couple minor comments.

packages/client/src/scss/colors-semantic-tokens.scss
```diff
@@ -69,4 +69,7 @@ $interactive-active: $raw-blue-50;
 // TEXT
 $text-primary: $raw-gray-700;
 $text-secondary: $raw-gray-600;
-$text-placeholder: $raw-gray-500;
\ No newline at end of file
+$text-placeholder: $raw-gray-500;
+
+// LINKS
+$link-back: $raw-gray-500;
```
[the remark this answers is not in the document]
inaki: well, I think that is the point of semantic-tokens, there is variable for the different components. Although we could reduce this by using just color names. Right now for example the variable with same color is info-placeholder which is not very semantic for a link.

packages/client/src/App.vue
inaki: Right now our ralative path will work just in the src/...., so if we don't want to have scss dir on the src dir, we could change the configuration. But also it make sense to have the scss inside src dir. But right now is not working in the main branch as is.

jeffsmohan: I'm confused. Part of the change you implemented in this PR is in fact moving the scss inside the src directory. Which is exactly what you're saying makes it possible to use the `@` alias within the scss files. So isn't it now possible? When I attempted it on your branch locally, it did seem to work for me. 

End of the day, it's a minor style thing, so it's not worth blocking the PR over, but it seemed like a nice easy win.

jeffsmohan (changes requested):
I replied to our two threads. I don't consider those blocking issues. 

However, it looks like your latest commit incorrectly introduced a bunch of automated linter/prettier "fixes", which we should revert before landing this.

jeffsmohan (changes requested):
I replied to our two threads. I don't consider those blocking issues. 

However, it looks like your latest commit incorrectly introduced a bunch of automated linter/prettier "fixes", which we should revert before landing this.

packages/client/src/App.vue
inaki: We get this error now when importing in the main branch.

Screenshot 2024-03-12 at 9 35 07 AM

inaki: I could change the alias configurations also, both would work. I just thought moving this dir also make sense.

jeffsmohan (approved):
lgtm!

PR #2750, opened 2024-03-08
Drop the column grants.raw_body.

sanason:
## Description
This is a code clean-up PR. It contains a migration to drop the now redundant column `grants.raw_body`.

In https://github.com/usdigitalresponse/usdr-gost/pull/2435, we added a jsonb column grants.raw_body_json and populated it from the text column grants.raw_body. In https://github.com/usdigitalresponse/usdr-gost/pull/2454, we updated the grant ingestor to write to raw_body_json instead of to raw_body. In https://github.com/usdigitalresponse/usdr-gost/pull/2720, we deleted all code references to raw_body. It should be safe to drop the column now.

## Screenshots / Demo Video
N/A

## Testing
Run the migration, start up the app, click around and check that nothing is broken.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [ ] Added steps to test feature/functionality manually

## Checklist
- [x] Provided ticket and description
- [ ] Provided screenshots/demo
- [x] Provided testing information
- [ ] Provided adequate test coverage for all new code
- [ ] Added PR reviewers

PR #2751, opened 2024-03-09
Flag DB migrations in release notes

TylerHendrickson:
## Description

This PR updates the auto-generated release notes config so that PRs with the https://github.com/usdigitalresponse/usdr-gost/labels/database-changes label are grouped under a common heading. This should make it easier for releasers to check whether any Knex migrations are part of a release, which is necessary until we have an automated migration process.

## Screenshots / Demo Video

Sneak preview ahead of #2750:
image

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -28,6 +28,10 @@ categories:
     collapse-after: 10
     labels:
       - bug
+  - title: 🗂️ Database Changes
+    collapse-after: 1
+    labels:
+      - database-changes
   - title: 📖 Documentation improvements
     collapse-after: 10
     labels:


.github/release-drafter.yml
```diff
@@ -28,6 +28,10 @@ categories:
     collapse-after: 10
     labels:
       - bug
+  - title: 🗂️ Database Changes
+    collapse-after: 1
```
TylerHendrickson, line 32: Since the presence of at least 1 PR with a migration is all we really need to know, this should help minimize the real estate that this section takes up in the release notes.

as1729 (approved):

PR #2811, opened 2024-03-18
[Cleanup] Adopt eslint vue/recommended: Grant Finder: app files

jeffsmohan:
## Description
Brings the grants table vue files into compliance with new linting rules. 

- ./src/App.vue
- ./src/main.js
- ./src/components/Layout.vue

A few notes for the reviewer: 

1. Much of the diff is auto-fixed. Reviewing commit by commit should make this a bit easier to read. 
2. Renamed Layout->BaseLayout to comply with multi-word component lint rule. 

## Screenshots / Demo Video

Screenshot 2024-03-18 at 10 20 52 AM

## Testing
Confirmed that basic layout behavior remains unaffected by the change. 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/client/src/main.js
```diff
@@ -34,7 +34,7 @@ Vue.use(BootstrapVue);
 Vue.use(IconsPlugin);
 Vue.use(Vuelidate);
 Vue.use(VueRouter);
-Vue.component('v-select', vSelect);
+Vue.component('VSelect', vSelect);
```
sanason, line 37: I was surprised that this worked without updating all the places where `<v-select>` is used, until I read this - https://v2.vuejs.org/v2/guide/components-registration#Name-Casing. But it's still confusing maybe? Is there another PR where the `<v-select>` usages are updated to `<VSelect>`?

sanason (commented):
Just one comment/question.

packages/client/src/main.js
jeffsmohan: Yeah good question! I meant to leave a note to this effect on the PR but forgot when writing it up. 

As you saw at the link you pointed to, when a component is defined in PascalCase, you can reference it in either PascalCase or kebab-case in templates. That's why the change on this line didn't require the usages of `v-select` in templates across the site to change.

Do we want to enforce consistency here? I suppose it's arguable whether this falls under (a) "that's confusing, why is the name different?" or (b) "that's just how Vue works". But for my part, I tend to prefer consistency whenever possible (and especially when easy to enforce). 

There's a lint rule available to do this — [`vue/component-name-in-template-casing`](https://eslint.vuejs.org/rules/component-name-in-template-casing.html) — but apparently it's not included in the recommended set we adopted. So we have a couple options: 

1. We can add this rule to the linting migration before we go further down the implementation road. 
2. We could simply update existing v-select usages to match the VSelect casing, but skip enforcing anything with lint rules and accept that it might over time lose consistency. 
3. We could accept the status quo that Vue allows for either. (For example, we currently use most or all VueBootstrap components using kebab-casing.)

When I was talking with @TylerHendrickson about how to approach this migration, we went with the vue/recommended ruleset because it seemed like an easy way to adopt a community standard, but I'd personally be very open to going beyond and adding our own rules where it makes sense. And certainly including all the lint rules we want in a single migration project like https://github.com/usdigitalresponse/usdr-gost/milestone/21 is easier than kicking off another one later. 

If we wanted to make a quick pass on other lint rules to add, I can evaluate the uncategorized vue linting rules and come up with a list of appealing ones. (For that matter, we could also take this opportunity to review how much of the `@vue/airbnb` base ruleset is helpful, and whether there are other eslint rules beyond vue ones that we want to add. But that all really amps up the scope of the project.) 

Sorry for the essay, but what are your thoughts on a good path forward @sanason @TylerHendrickson ?

TylerHendrickson: I think it's perfectly fine if there's a desire to adopt additional standard conventions that go beyond `vue/recommended`. IMO the best path forward would be to document the proposed set of additional linting rules as an [ADR](https://github.com/usdigitalresponse/usdr-gost/tree/main/docs/decisions) in a pull request. That would give everyone a chance to review the proposal(s) and potentially weigh in with opinions on what should ultimately be accepted or rejected, which helps ensure that there's group accountability for the decision.

For my part, a few details I think would be useful for this ADR to include:
1. Rationale/justification that makes clear the perceived benefit/detriment of (not) adopting each given rule.
2. Whether any rules are not generally auto-fixable by eslint.

jeffsmohan:
@sanason Thanks for raising a good question! After talking it through, I think my proposal would be to punt on solving it, unfortunately — see the comment thread above. 

Merged main and fixed a couple conflicts. Please re-review when you get a chance!

sanason (approved):
👍

PR #2863, opened 2024-03-27
Adjust Datadog RUM sampling in Staging

TylerHendrickson:
## Description

This PR sets Datadog RUM sampling rates in our Staging environment to 100% for both session sampling and session replay (currently they are sampled at 10% and 1%, respectively). Normally, we don't need this much RUM data collected in this environment, but doing so temporarily allows us to test and confirm receipt proper event instrumentation ahead of the Grant Details redesign launch.

Once this is deployed, I'll create a separate PR to turn the rates back down to normal levels, which we can merge after testing has concluded.


jeffsmohan (approved):
lgtm!

PR #2890, opened 2024-03-31
Rough draft: scheduled task for send grant digest emails

jeffsmohan:
### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


terraform/main.tf
```diff
@@ -426,6 +437,52 @@ resource "aws_iam_role_policy" "api_task-publish_to_arpa_treasury_report_queue"
...
+  enabled = var.api_enabled && var.api_enable_grants_digest
+
+  name_prefix = "${var.namespace}-digest-email-send"
+  description = "Executes an ECS task that sends grants digest emails, between 9am - 10am ET"
+
+  # Schedule
+  schedule_expression          = "cron(0 9 * * ? *)"
+  schedule_expression_timezone = "America/New_York"
+  flexible_time_window         = { hours = 1 }
+  retry_policy_max_attempts    = 10
```
jeffsmohan, line 451: Set this to 1 to avoid multiple sends if there's an issue

terraform/main.tf
```diff
@@ -125,6 +125,16 @@ module "arpa_treasury_report_security_group" {
   allow_all_egress = true
 }
 
+module "digest_email_send_security_group" {
```
jeffsmohan, line 128: Need to add this to both api and postgres, as well as digest email send security groups

terraform/prod.tfvars
```diff
@@ -59,16 +59,17 @@ website_google_tag_id = "G-WCDTMFM6RG"
...
+api_enabled                            = true
+api_default_desired_task_count         = 3
+api_minumum_task_count                 = 2
+api_maximum_task_count                 = 5
+api_enable_grants_scraper              = false
+api_enable_grants_digest               = false
+api_enable_new_team_terminology        = true
+api_enable_my_profile                  = true
+api_enable_saved_search_grants_digest  = true
+api_enable_grant_digest_scheduled_task = false
```
jeffsmohan, line 71: As written, this PR uses this variable simultaneously to disable running the ECS scheduled task (`enabled = var.api_enabled && var.api_enable_grant_digest_scheduled_task`) and to short-circuit the script itself (`if (process.env.ENABLE_GRANT_DIGEST_SCHEDULED_TASK !== 'true') { return; }`). Does this feel redundant? Or helpful on the off chance the script gets triggered via some other means?

terraform/staging.tfvars
```diff
@@ -56,16 +56,17 @@ website_google_tag_id = "G-D5DFR7BN0N"
...
+api_enabled                            = true
+api_default_desired_task_count         = 1
+api_minumum_task_count                 = 1
+api_maximum_task_count                 = 5
+api_enable_grants_scraper              = false
+api_enable_grants_digest               = false
+api_enable_new_team_terminology        = true
+api_enable_my_profile                  = true
+api_enable_saved_search_grants_digest  = true
+api_enable_grant_digest_scheduled_task = false
```
jeffsmohan, line 68: Do we want this to be flipped to true in staging so we can test the functionality as soon as it lands?

terraform/main.tf
```diff
@@ -125,6 +125,16 @@ module "arpa_treasury_report_security_group" {
...
+module "grant_digest_scheduled_task_security_group" {
+  source  = "cloudposse/security-group/aws"
+  version = "2.2.0"
+
+  namespace        = var.namespace
+  vpc_id           = data.aws_ssm_parameter.vpc_id.value
+  attributes       = ["grant_digest_scheduled_task"]
+  allow_all_egress = true
+}
+
```
TylerHendrickson, line 137: It occurs to me that, since the `grant_digest_scheduled_task` module defined in `terraform/modules/gost_api/grant_digest.tf` is configured with all the same security groups that are passed in for any other task (specifically the main API workers) controlled by `terraform/modules/gost_api`, this security group is redundant.

To be super verbose: the `terraform/modules/gost_api` module source uses all provided values in `var.security_group_ids` when placing tasks in various security groups. More to the point, it passes this same list directly to its `module "grants_digest_scheduled_task"` resource. Since this list already contains a security group  common to our Postgres cluster*, we don't need to provide another. The only reason this new security group would be necessary is if `terraform/modules/gost_api` declared a new input variable like `var.grants_digest_scheduled_task_security_groups` and _only_ used those values when configuring networking for that task.

I should have caught this during our review yesterday, but I was (mistakenly) thinking that the `module "grant_digest_scheduled_task"` resource would be defined in the main project (`terraform/` directory) rather than as a sub-resource of `module "gost_api"`. 

*Feel free to ignore everything below, but looking at the current security group configuration, it appears we somehow mismatched the `api_to_postgres_security_group` and `consume_grants_to_postgres_security_group` usages. The configuration _should_ be:
- `module.api_to_postgres_security_group.id` passed to:
  - `module "gost_api"` via `security_group_ids` input variable array
  - `module "postgres"` via `ingress_security_groups` input variable array
- `module.consume_grants_to_postgres_security_group.id` passed to the `security_group_ids` input variable arrays for:
  - `module "consume_grants"` via `security_group_ids` input variable array
  - `module "postgres"` via `ingress_security_groups` input variable array

This mismatch isn't the most major problem in the world – both security groups are configured identically, so this is ultimately just a naming convention, and they could even be consolidated down to a single security group. Just mentioning here because I noticed while reviewing.

**TL;DR** You can safely remove the `module "grant_digest_scheduled_task_security_group"` resource and its usages.

terraform/modules/gost_api/main.tf
```diff
@@ -1,5 +1,9 @@
 data "aws_region" "current" {}
 data "aws_caller_identity" "current" {}
+data "aws_ecs_cluster" "default" {
+  count        = var.enabled ? 1 : 0
+  cluster_name = var.ecs_cluster_name
+}
```
TylerHendrickson, line 6: 👍

terraform/prod.tfvars
TylerHendrickson: Even if the configuration is ultimately sourced from one place, I think that being able to have a check at the infrastructure level as well as at the runtime level is a useful separation of concerns, so this feels right to me.

terraform/staging.tfvars
TylerHendrickson: Yes please!

TylerHendrickson (changes requested):

terraform/main.tf
jeffsmohan: Yeah, I think I was borrowing this idiom from things like `arpa_treasury_report_security_group`, but your assessment makes sense, and it does seem like consolidating security groups to a clear set of logical groupings, rather than "one per process", makes more sense to me. 

>I should have caught this during our review yesterday, but I was (mistakenly) thinking that the module "grant_digest_scheduled_task" resource would be defined in the main project (terraform/ directory) rather than as a sub-resource of module "gost_api".

Fwiw, it was located in `terraform/main.tf` originally, but I moved it under the `gost_api` module to reuse the `aws_ecs_task_definition` instead of having to redefine it. 

Anyway, I'll remove the new security group!

TylerHendrickson (approved):
Looks good!

Edit: Looks like there was a formatting issue after this change; pushed up results of `terraform fmt`.

terraform/modules/gost_api/grant_digest.tf
```diff
@@ -0,0 +1,46 @@
+module "grant_digest_scheduled_task" {
+  source  = "../scheduled_ecs_task"
+  enabled = var.enabled && var.enable_grant_digest_scheduled_task
+
+  name_prefix = "${var.namespace}-grant-digest-scheduled-task"
```
TylerHendrickson, line 5: This is annoying, but [this error](https://github.com/usdigitalresponse/usdr-gost/actions/runs/8546141230/job/23416072415?pr=2890#step:15:27) in the failing `terraform plan` CI job is due to this `name_prefix` value being too long for certain AWS resources' max character restrictions – the value for the Staging environment, `gost-staging-grant-digest-scheduled-task`, is 40 characters, and the max length for an IAM role prefix is 38 characters ☹️ 

The fix is fortunately easy, albeit not quite as readable as the current solution:
```suggestion
  name_prefix = "${var.namespace}-grant-digest-task"
```
Other preferences for shortening this value (`gost-staging-grant-digest-sched-task` is 37 characters, for example) are also acceptable.

TylerHendrickson (changes requested):
@jeffsmohan New requested change based on the current CI failure.

jeffsmohan:
@TylerHendrickson fixed!

TylerHendrickson (approved):

PR #2924, opened 2024-04-09
Define Terraform infrastructure for email metrics

sanason:
## Description
See ticket for details. I followed the implementation plan very closely. I think the only difference is that I'm not passing `SES_CONFIGURATION_SET_DEFAULT` into the `consume_grants` container because that task is not otherwise configured to send email.

Note that this branch is based on the branch in https://github.com/usdigitalresponse/usdr-gost/pull/2945, which I was hoping to get merged first. That PR isn't required for these changes to work and I'm willing to rebase on `main` but I would like that PR to be reviewed. 🙂 

## Screenshots / Demo Video
N/A

## Testing
Once this is deployed to staging, it should be testable through the AWS console:
1. Go to the SNS page and subscribe to the `datadog-forwarder` topic. I did this with an email subscription.
2. Go to the SES page and click on the "staging.grants.usdr.dev" identity.
3. Click "Send Test Email".
4. On the "Send Test Email" page, select 'gost-staging-default' under "Configuration Set"
5. Switch from "Formatted" to "Raw".
6. Select "Custom" under "Scenario" and enter your email address under "Custom Recipient".
7. Enter this text in the "Message" box:
```
Subject: Amazon SES Raw Email Test
MIME-Version: 1.0
Content-Type: text/html

<!DOCTYPE html>
<html>
<body>
<p>Here is a formatted link: <a href="https://staging.grants.usdr.dev/grants/113293">Go to grant</a>.</p>
</body>
</html>
```
8. Fill out any other required fields and send the test email.

Now you can confirm:

- You received the email.
- The link in the email has a url that starts with `https://mail-redirect.us-west-2.staging.grants.usdr.dev`. Click on the link.
- You received several "Amazon SES Email Event Notification" notifications (through whatever subscription channel you set up) - a Send event, a Delivery event, an Open event and a Click event.

All of this worked for me after running `terraform apply` against my personal AWS account, setting the `website_domain_name` variable to a domain that I own and that I verified in SES.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


terraform/email.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_route53_record" "mail_redirect" {
+  zone_id = data.aws_ssm_parameter.public_dns_zone_id.value
+  name    = "mail-redirect.${data.aws_region.current.name}"
+  type    = "CNAME"
+  records = ["r.${data.aws_region.current.name}.awstrack.me"]
+  ttl     = 86400
+}
+
```
sanason, line 8: If we used an `aws_sesv2_email_identity` resource to create our SES domain identity (which seems to be created manually currently?), then we could associate this configuration set with the domain identity and thus avoid the need to specify the configuration set in our email-calling code.

Bad idea or good idea? @TylerHendrickson

TylerHendrickson: @sanason Not a bad idea!

We currently use the [cloudposse/ses/aws module](https://github.com/cloudposse/terraform-aws-ses) to configure our SES identities via the (older) [`aws_ses_domain_identity` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ses_domain_identity), which is based on the [plain/v1 SES API](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ses/index.html). This is part of the limited "bootstrap" infrastructure that we define on a per-service basis outside of service-level infrastructure.

Broadly, our approach to bootstrapping is to configure the following types of infrastructure separately/ahead of anything defined in a particular service repo's Terraform code:
- Common infrastructure that is shared (or potentially shared) between multiple services, in order to avoid creating inter-service dependency chains.
- VPC networking infrastructure, since it's difficult to do CIDR math and subnet allocations without preexisting knowledge.
- IAM roles and permissions used by service-level CI/CD workflows, to ensure that these roles always have an attached permissions boundary and that they can only be assumed by properly auth'd workflows in the service's dedicated repository.
- DNS namespace allocation (Route53 hosted zones), to ensure that all service-level DNS management utilizes a dedicated namespace and avoid clashing with other services.
- Secrets (usually service keys) and other values that are provided as per-environment deployment parameters, either because the parameter comes from something that the service itself depends on OR (in the case of secrets) because we generally prefer to manage secrets outside of GitHub.
- **Infrastructure for sending emails via SES,** specifically an MX record in the Route53 hosted zone, plus the `cloudposse/ses/aws` module.

My general preference is to have less bootstrap-level infrastructure where possible and instead define it at the service level. Particularly in the case of the last bullet point, I can't think of a compelling reason why that **needs** to be bootstrapped given that all DNS records are provisioned within the service-specific hosted zone and none of the other [resources](https://github.com/cloudposse/terraform-aws-ses?tab=readme-ov-file#resources) managed by that Cloud Posse module are otherwise in danger of clashing with another service. My guess is that this seemed like a good idea at one point in part due to the need to ["un-sandbox" SES accounts](https://docs.aws.amazon.com/ses/latest/dg/request-production-access.html), but I don't really see that as particularly compelling.

If you think it would be useful to manage this service's SES identity (and related DNS records) in this repo rather than elsewhere, I'm definitely open to it. However, it might be useful to defer that (and any accompanying complexity) to a separate PR. One consideration is whether an `aws_sesv2_email_identity` resource can coexist with an `aws_ses_domain_identity` resource that specifies the same domain name; it might be that we need to delete the bootstrapped infrastructure before we can apply what you're suggesting at the service level. That's certainly achievable, but would of course require a bit of coordination to roll out the changes.

Finally, I'll add that while we do have DKIM properly configured for the `staging.grants.usdr.dev` domain identity and we get [automatic SPF authentication](https://docs.aws.amazon.com/ses/latest/dg/send-email-authentication-spf.html) provided by SES (which is based on `<subdomain>.amazonses.com`), but we do not have a proper DMARC policy, which we probably should. If we do decide to investigate improvements to our SES identity configuration, it would be useful to look into DMARC and (potentially) using our own subdomain for SPF instead of Amazon's.

terraform/email.tf
```diff
@@ -0,0 +1,49 @@
...
+      "COMPLAINT",
+      "DELIVERY",
+      "OPEN",
+      "CLICK",
+      "RENDERING_FAILURE",
+      "DELIVERY_DELAY",
+      "SUBSCRIPTION"
+    ]
+  }
+}
```
TylerHendrickson, line 49: Obligatory "please add a blank newline to the end of this file" 🙂

TylerHendrickson (changes requested):
@sanason This looks great! I responded to your question and provided some details for your consideration. Requesting one trivial change; let me know what you think about the SES identity topic.

TylerHendrickson (approved):

terraform/email.tf
sanason: +1 for thinking about this suggestion later and merging this PR now. The last paragraph of your comment, in particular, is throwing around a lot of terms that I need to look up (SPF authentication, DMARC policy).

PR #2792, opened 2024-03-14
Grant digest email adjustments

jeffsmohan:
## Description
Tweaks the grants digest email to bring it fully in line with design expectations. Specifically: 

1. Left-align the product name with the USDR logo
2. Logo clicks through to grant finder tool
3. Add a "from" name
4. Add preheader preview text for the email

## Screenshots / Demo Video

image

Note that the ethereal email tool doesn't display preview text in the inbox, so we'll want to double-check this functionality with real emails sent on staging. 

## Testing

Testing this in dev is a lot of pretty fiddly effort, but here are the steps: 

1. Set up so emails actually get sent/captured in dev
   1. Create an ethereal email to work with: https://ethereal.email/ (nice, one-click creation of an ephemeral email account)
   2. Update the `NODEMAILER_*` environment variables in your `packages/server/.env` file from the values listed on your ethereal account
   3. Comment out `secure: true` in our nodemailer setup (Gmail uses a secure connection, but ethereal doesn't): https://github.com/usdigitalresponse/usdr-gost/blob/05f688e75df1e3fbf6698f88b49281dd117eb280/packages/server/src/lib/email/email-nodemailer.js#L36
   4. Rebuild your docker container so it picks up the new env variables (`docker compose down app` and `docker compose up -d`)
2. Trigger the emails manually via a node REPL
   - Digest email
     1. `docker compose exec app yarn node`
     2. ```node
           const emailService = require('./packages/server/src/lib/email');
           const knex = require('./packages/server/src/db/connection')
           const grants = await knex('grants').limit(3)
           emailService.sendGrantDigest({name: 'Test agency', matchedGrants: grants, matchedGrantsTotal: 3, recipients: ['test@example.com'], openDate: '2024-01-01'})
           ```
3. Once the test email is sent, go to your "Messages" tab in your Ethereal mail account, and you should see the sent email listed there for viewing

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [ ] Added steps to test feature/functionality manually

## Checklist
- [ ] Provided ticket and description
- [ ] Provided screenshots/demo
- [ ] Provided testing information
- [ ] Provided adequate test coverage for all new code
- [ ] Added PR reviewers

PR #2940, opened 2024-04-12
Fix email "from name" via AWS SES and standardize names

jeffsmohan:
## Description
This PR does a couple related things: 

1. Implements the "from name" functionality for email sending via AWS SES. (This was previously implemented for nodemailer transport here: https://github.com/usdigitalresponse/usdr-gost/pull/2792)
2. Standardizes the "from name" used for grant finder and ARPA tools across all emails, per the comment linked above.

I've done my best to verify ahead of time that the [`name-addr` email format](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4) should be accepted by AWS SES (e.g., there's an example of this format [in their docs here](https://docs.aws.amazon.com/ses/latest/dg/send-an-email-using-sdk-programmatically.html).) However, we should carefully test that emails are still working as desired in staging before deploy. 

## Screenshots / Demo Video
Screenshot 2024-04-12 at 8 50 36 AM

## Testing
I've tested that the updated emails have the correct name in dev using the debug email tool and mailpit (see the screenshot). However, the key fix with AWS SES will need to be tested in staging, since that infra isn't set up to run in dev. 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (approved):
This looks correct to me. I also found some conflicting guidance for proper "display name" formatting, with [this follow-up post on SO](https://stackoverflow.com/a/68362600) suggesting that `From Name <person@example.org>` (without quotes) is the correct way to do that. 

However, the currently implemented solution is consistent with [RFC 5322](https://www.ietf.org/rfc/rfc5322.txt), which is generally favored by SES for this type of thing IME, so I agree it makes the most sense to start as-is and confirm in Staging.

PR #2965, opened 2024-04-22
Proof of concept: Webpack to Vite migration

jeffsmohan:
## Description

This PR is meant as a functioning demo of a migration to Vite as our frontend dev server and bundler for production. (It's not meant to be merged. If we agree on a path forward, this will get split up into smaller, more manageable PRs to land individually.)

I've verified the grants and arpa sites are basically functioning, but I haven't gone through in detail yet to make sure every detail is in place. (If you find anything that isn't working yet in vite, please flag it for me!)

**Architecture and product considerations:**

- Vite is generally more modern and easier to understand and work with than webpack
- Vite dev server is much quicker to start up and update, including support for hot module replacement
- We'd drop support for non-evergreen browsers (notably, IE11) — this was discussed live, and decided this is fine
- Migration to vite is not required but will make a migration to vue3 a bit simpler

**The main tasks required for the migration (essentially, a more organized list of what this PR is doing):**

1. Preparation prior to the Vite switchover
   1. Migrate stray CommonJS (`require` and `module.exports`) to ES6 (`import` and `export`) syntax
   2. Upgrade `vuelidate` package to v2 (our current v0.7 is not compatible with vite or vue3)
   3. Remove `lodash-checkit` dependency (relies on `global` reference, which is only available in CJS/node context; checkit does not appear to be safe to run in a browser context... but we're only using it for email format checking, and we should just migrate this usage to `vuelidate`)
   4. Switch our test runner to `vitest` (see separate proof of concept PR: https://github.com/usdigitalresponse/usdr-gost/pull/2998)
2. Vite switchover
   1. Update our package dependencies (mostly removing `@vue/cli*` and adding `vite`)
   2. Update config (replace `vue.config.js` with `vite.config.js`, updates to `.eslintrc.js`)
   3. Update `package.json` script references to run dev server and build via `vite` instead of `vue-cli-service`
   4. Reconfigure entrypoints for grants and arpa reporter (vite expects `index.html` in a particular location to indicate separate entries for the two builds in dev)
3. Plenty of QA in staging

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson:
I'm in favor of the proposed move to Vite, and this PoC makes sense/looks good to me.

I do have one concern, which is that migrating the `packages/client` code to ES6 while leaving the `packages/server` code in CJS feels a bit messy and likely confusing to newcomers (honestly I'd probably get a bit flummoxed if I was going back and forth between client/server source code). If we decide to head in this direction, it would be ideal to migrate the entire repository to ES6 so that coding conventions remain consistent, assuming that doesn't unreasonably blow up the scope of this effort.

jeffsmohan:
Thanks for reviewing, @TylerHendrickson! 

> If we decide to head in this direction, it would be ideal to migrate the entire repository to ES6 so that coding conventions remain consistent, assuming that doesn't unreasonably blow up the scope of this effort.

Completely agree on all counts! The problem is unfortunately I do think it'd be a big enough scope that it deserves a separate effort...

As a quick snapshot of the scale, while the `client` package uses `require(...)` 12 times, `server` uses `require(...)` 581 times across 126 files. Once we're at that scale, it deserves an incremental approach rather than a single PR. (Good news is we can use the same "eslint rule with ignored files as punch list" strategy we did with vue/recommended lint rules here. `eslint-plugin-import` has a [`no-commonjs` rule](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-commonjs.md) that we can use as a ratchet and eventual enforcement mechanism.) 

The other difference is that imports are all exercised on the frontend when we build for production, which is a nice check that they're all valid, whereas we don't get that for server code and our test coverage isn't stellar. (That said, we can likely achieve a similar result with `eslint-plugin-import`'s static analysis using [`no-unresolved`](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-unresolved.md) to verify imports exist.)

TylerHendrickson:
One other thing to mention: when I run `yarn workspace client build` locally, the resulting `dist` directory includes a `deploy-config.js` file. I realize this is just a PoC at the moment, but just dropping a note here that we'll want to figure out how to exclude that from builds.

jeffsmohan:
> One other thing to mention: when I run `yarn workspace client build` locally, the resulting `dist` directory includes a `deploy-config.js` file. I realize this is just a PoC at the moment, but just dropping a note here that we'll want to figure out how to exclude that from builds.

@TylerHendrickson This is a good callout, thanks! There's actually no 1:1 functionality to conditionally exclude a specific file from the public folder with vite. There are a few ways we could handle this: 

1. Do away with `deploy-config.js` and rely on environment variables instead. (Is there background/context on why we have the `deploy-config.js` setup?) This I think aligns us better with standard web dev practices and minimizes confusion by minimizing the surface area of differences between environments. (Unless there's some need for our environments to have different code, rather than just different data/config?)
2. Bit hacky, but very easy: we can update the build command in package.json to `vite build && rm dist/deploy-config.js`. 
3. Move the `deploy-config.js` into our `src/` instead of our `public/` directory, and then we can use rollup's configuration settings to manage splitting it out, naming it correctly, and excluding it. (I think this is possible, but the config would be a bit messy, and I haven't verified it would work.)
4. Do nothing. My understanding is the `resource "aws_s3_object" "deploy-config"` terraform block will overwrite the `deploy-config.js` file in the build even if one is present.

PR #2974, opened 2024-04-24
Send emails via SES using a configuration set.

sanason:
## Description
This PR does as the ticket describes. However, since we haven't fixed the problem noted in #2996 yet, this PR also adds a feature flag that disables all of the email tracking functionality that the configuration set is supposed to bring us. My plan is:
1. Merge this PR which should not break anything.
2. Enable email tracking in staging in the PR for #2996 and use staging to test that all of our emails still work correctly and that we are receiving email events in Datadog.
3. Enable email tracking in prod in a separate PR.

In [this PR comment thread](https://github.com/usdigitalresponse/usdr-gost/pull/2924#discussion_r1573065817), we discussed a different way of setting the configuration set used to send emails. My main takeaway from that conversation was that the different way was not an approach that I could code up myself as a volunteer. So, in the interest of getting to the point where we can start having email metrics, I'm taking the originally proposed approach, which seems perfectly fine too.

## Screenshots / Demo Video
N/A

## Testing
Not testable in dev since it depends on AWS infrastructure.

Since `enable_email_tracking` is disabled in staging, there won't be new functionality to test there. We should do some regression testing, namely:

1. Have a login passcode email sent from staging. Check that it was sent and that you received it.
4. Hover over the link in the email and check that the URL still looks like `https://api.staging.grants.usdr.dev/api/sessions?passcode=some_passcode`. Click on the link and make sure it works.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/terraform/email.tf
+++ b/terraform/email.tf
@@ -19,12 +19,12 @@ resource "aws_sesv2_configuration_set" "default" {
 }
 
 data "aws_sns_topic" "datadog_forwarder" {
-  count = var.ses_datadog_events_enabled ? 1 : 0
+  count = var.email_enable_tracking && var.ses_datadog_events_enabled ? 1 : 0
   name  = "datadog-forwarder"
 }
 
 resource "aws_sesv2_configuration_set_event_destination" "default" {
-  count                  = var.ses_datadog_events_enabled ? 1 : 0
+  count                  = var.email_enable_tracking && var.ses_datadog_events_enabled ? 1 : 0
   event_destination_name = "DatadogForwarderSNSTopic"
   configuration_set_name = aws_sesv2_configuration_set.default.configuration_set_name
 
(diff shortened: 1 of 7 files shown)


terraform/email.tf
```diff
@@ -19,12 +19,12 @@ resource "aws_sesv2_configuration_set" "default" {
 }
 
 data "aws_sns_topic" "datadog_forwarder" {
-  count = var.ses_datadog_events_enabled ? 1 : 0
+  count = var.email_enable_tracking && var.ses_datadog_events_enabled ? 1 : 0
```
sanason, line 22: I suppose I could have set `ses_datadog_events_enabled` to false instead of introducing a new variable but it felt like that variable was not meant to be used as a feature flag.

terraform/email.tf
sanason, line 1: Links in emails are not rewritten to point to the tracking domain as long as the configuration set that is used does not have an associated event destination. I tested that in my own AWS account by sending test emails in the AWS console.

TylerHendrickson (approved):
Looks great! 💯

PR #2997, opened 2024-04-29
Remove unused dependencies

jeffsmohan:
## Description
I noted a few unused dependencies in our client `package.json`, confirmed they are indeed unused, and removed them to ensure we're not building up codebase cruft. Specifically: 

- [core-js](https://www.npmjs.com/package/core-js) (library of polyfills for modern JS features in browsers)
  - Added in the PR migrating ARPA into the GOST codebase: https://github.com/usdigitalresponse/usdr-gost/pull/302
  - We’re only targeting evergreen browsers that pretty much support all modern JS, and regardless the library was never imported anywhere in our code, which means the polyfills were never actually used
- [vue2-filters](https://www.npmjs.com/package/vue2-filters) (collection of helpful Vue filters)
  - Introduced in commit https://github.com/usdigitalresponse/usdr-gost/commit/bbe11d4a0b035a72f1f2ef8f2b7ac5fd2523fc83
  - Usage commented out in commit https://github.com/usdigitalresponse/usdr-gost/commit/19011d98d3cea9a0e7d2813cfc5c0b1e0552d81f and removed entirely in https://github.com/usdigitalresponse/usdr-gost/commit/fde96976aaeb9fda3a0f7b2bc12ac80b3a85cd61
- [standard](https://www.npmjs.com/package/standard) and [eslint-config-standard](https://www.npmjs.com/package/eslint-config-standard) (JS code style enforcement)
  - Added in the PR migrating ARPA into the GOST codebase: https://github.com/usdigitalresponse/usdr-gost/pull/302
  - But not actually invoked anywhere (would expect to see it run in a `package.json` script or installed in our `.eslintrc.js` or maybe as a command in terraform, but none of these exist)
  - I assume this was invoked in the ARPA codebase somehow but that invocation didn’t make it in the migration; meanwhile, we’ve evolved how we’re enforcing js style, so it’s not really needed
  - Note that while I was focused on removing unused packages from the client package, I saw these were also unused in the server package, so I opted to remove them there as well

As an aside, I really wanted to remove jQuery and popper.js, as they're ancient, large, and seemed unused. However, they appear to be peer dependencies of vue-bootstrap, so they are in fact still needed. 

## Testing
This change should have no user-facing changes or effects. I verified manually that the site still loads and basic functionality is intact. 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -35,7 +35,6 @@
     "vue-multiselect": "^2.1.6",
     "vue-router": "^3.5.2",
     "vue-select": "^3.16.0",
-    "vue2-filters": "^0.14.0",
     "vuex": "^3.6.2"
   },
   "devDependencies": {
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -55,7 +55,6 @@
     "babel-plugin-istanbul": "^6.1.1",
     "chai": "^4.3.4",
     "eslint": "^7.12.1",
-    "eslint-config-standard": "^16.0.3",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-import-alias": "^1.2.0",
     "eslint-plugin-node": "^11.1.0",
@@ -69,7 +68,6 @@
     "sass": "1.52.1",
     "sass-loader": "13.0.0",
     "sinon": "^17.0.1",
-    "standard": "^16.0.3",
     "typescript": "^4.7.2",
     "vue-template-compiler": "^2.7.12",
     "webpack": "^5.74.0"
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -23,7 +23,6 @@
     "@vuelidate/validators": "^2.0.4",
     "bootstrap": "^4.6.1",
     "bootstrap-vue": "^2.19.0",
-    "core-js": "^3.21.1",
     "jquery": "3.6.0",
     "lodash": "^4.17.21",
     "luxon": "^3.4.3",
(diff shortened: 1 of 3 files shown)


TylerHendrickson (approved):
@jeffsmohan Changes look good! 

Do you have any thoughts/ideas about incorporating something like [depcheck](https://github.com/depcheck/depcheck) into our CI checks to help mitigate this kind of this going forward?

jeffsmohan:
> Do you have any thoughts/ideas about incorporating something like [depcheck](https://github.com/depcheck/depcheck) into our CI checks to help mitigate this kind of this going forward?

Interesting. In an ideal world, I think something like that would be a great addition, but I worry it's a pretty tough challenge to avoid false positives (deps that look unused but really are) with a tool like that. There's just so many non-standard ways to use the dependencies. 

Here's the output of depcheck run against our client code in `main`: 

image

It did a pretty good job with non-dev dependencies (it found `core-js` and `vue2-filters`, the ones I removed in this PR; it also found `vue-multiselect`, which I believe is unused but we actually [load its css direct from unpkg](https://github.com/usdigitalresponse/usdr-gost/blob/98f5d81a540c75c48784f8d557050deee66bd5f8/packages/client/public/index.html#L28), so I was leaving that for a separate PR). However, it flagged a bunch of dev dependencies that I believe are mostly useful, alongside some which are dead. 

Anyway, I'd hope this is a rare enough problem that fighting with an imperfect tool is just not worth the effort. (Also, the penalty for missing dead dependencies is minimal... they don't get bundled into the production build or anything, so it's just additional cycles for devs and CI installing them and having to consider them in our package.json.)

PR #3050, opened 2024-05-14
Fix markGrantAsViewed bug

jeffsmohan:
## Description
Fixes the bug described in the linked ticket. High-level, our `grants_viewed` database table used to have a compound PK on grant_id and agency_id only. This caused a duplicate record error when a second user in that agency viewed the grant. So this PR migrates the PK to be on grant_id, agency_id, AND user_id. Going forward, we should now have one record per user/agency/grant combo in that table. 

Related updates: 
- Updated the `markGrantAsViewed` method to upsert with a new updated timestamp instead of erroring if a record already exists for the given PK
- Updated existing queries on the `grants_viewed` table to expect multiple rows per grant/agency combination, and to select distinct on the relevant rows to ensure we still have an accurate count of agencies that have viewed a given grant. (If we eventually want to surface users who have viewed, that will now be possible going forward as well, but we don't want to alter the expected behavior in this fix.)
- Added tests for the method.  

## Testing
- Added unit tests
- Manually verified on devbox that multiple users could view the same grant without error, and that those were recorded as separate rows in the table

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/migrations/20240430210546_grants_viewed_unique_add_user_id.js
```diff
@@ -0,0 +1,48 @@
...
+    await knex.raw(`
+        ALTER TABLE grants_viewed 
+            DROP CONSTRAINT grants_viewed_pkey,
+            ADD CONSTRAINT grants_viewed_pkey PRIMARY KEY USING INDEX idx_grants_viewed_grant_id_agency_id_unique;  
+    `);
+    // In the forward migration, user_id gets automatically marked non-nullable when it's included
+    // in the PK so we want to reverse that here
+    await knex.schema.table('grants_viewed', (table) => {
+        table.setNullable('user_id');
+    });
```
TylerHendrickson, line 41: I'm not sure it matters much, but since this seems likely to fail once a grant has been viewed twice for the same agency, I wonder if it's better to skip the downward revision here and instead implement the upward revision so that it's a no-op if there's nothing to do.

TylerHendrickson (approved):
One comment for you to consider, but I think this looks good.

packages/server/migrations/20240430210546_grants_viewed_unique_add_user_id.js
jeffsmohan: Having the down migration was super-helpful in development, since I needed to revert this migration before switching back to another branch (or blow away the DB, but then I would need to re-load the staging dump). And it could conceivably be useful if there was a need to revert the deploy ~immediately, before any conflicting grant views were recorded. But yeah, you're totally right this would fail once you have multiple views recorded on a grant. 

What are the circumstances and likelihood the down migration would be used in production? (Should we raise an error to ensure we can't migrate down; should it silently noop so we can migrate down; should we drop duplicate rows and then restore the table to its original structure?)

It sounds like migrations are managed by hand (as opposed to being automatically run, and automatically rolled back for e.g. failed deploys), correct? If so, just a noop down migration would give us the ability to down-migrate through this to other down migrations, if the need arose. And if any issues arose with this PR/migration specifically, we could write up a quick fix PR expressly to address the issue. Is that what you were thinking? 

If so, I can update the down migration to a noop and the up migration to first check whether its changes have been applied so it can be up-migrated repeatedly without failure. But I wanted to check first, since it'll take a tiny bit of digging/testing to make sure the "check before applying the up migration" part works as expected.

TylerHendrickson: I'm not particularly worried about the down migration being run in Production environments (and yes, those are executed manually); this was more a concern focused on dev workflows that involve migrating the DB up, adding data, and then downgrading the DB to a state that's incompatible with the added data. It's enough of an edge-case concern that I'm not sure it's worth addressing proactively, especially given your report that having a working down migration path was quite useful for your own workflows.

Also, my own dev workflow involves regularly nuking and reloading the DB, so maybe that's skewing my judgement here :)

jeffsmohan: Yeah ok, I agree there's a more technically correct approach than what's currently in this PR, but I also agree it's probably not worth any more time to implement. I'm going to move forward with landing this PR, thanks for talking that through!

PR #3051, opened 2024-05-14
Ensure add form modals reset when re-opened

jeffsmohan:
## Description
The "Add" modals for users, teams, and organizations are not resetting to empty state between usages as desired. This PR ensures these modals reset their form values to blanks when opening for the correct UI. 

## Screenshots / Demo Video

https://github.com/usdigitalresponse/usdr-gost/assets/126250/1ef6e5cc-efb3-44fa-a76e-11bb4264978c

## Testing
Verified on devbox that the form resets on every modal opening

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (approved):
Is there any way we can make this behavior default? If not (or unreasonable effort), I'm good with the PR as-is too.

jeffsmohan:
> Is there any way we can make this behavior default? If not (or unreasonable effort), I'm good with the PR as-is too.

I wondered the same, but unfortunately I couldn't think of a particularly clean way to do that here. The problem stems from the fact that the modals are shown/hidden rather than created/destroyed. So the form logic doesn't get re-created between re-opening the modal, and it only knows the modal is being hidden/re-opened if we hook up the logic to tell it. 

One reasonable option would be to re-create modals every time we display, but that would be a bit of a bigger change, and it doesn't prevent someone implementing a modal that just hides/shows in the future, so it didn't seem like a meaningful win.

PR #3076, opened 2024-05-20
Vitest migration

jeffsmohan:
## Description
This PR migrates our current client-side test runner/assertions/mocking (based on `vue-cli`, `mocha`, `chai`, `sinon`) over to `vitest` (https://vitest.dev/). This supports our upcoming migration from `vue-cli` over to `vite` as our development server and production build utility. (See [Decision #0008](https://github.com/usdigitalresponse/usdr-gost/blob/main/docs/decisions/0008-migrate-to-vite.md) for background.)

This PR is meant to be easiest to review commit by commit. The key changes are: 

1. Codemod to automate transition from mocha/chai syntax to jest syntax for expect assertions
2. Set up vitest and vite.config.js
4. Update coverage settings to match previous workflow
5. Update CI/QA workflows to use the updated test/coverage commands
6. A few manual tweaks to the tests (replace `sinon` usage, a few minor fixes)
7. Remove old infrastructure (`nyc` config, `babel` config, packages no longer used)
8. Switch from implicit globals (e.g., `describe`, `expect`, `it`, etc.) to explicit imports
9. Update testing documentation

I also want to specifically call out a few design considerations that might be worthy of more discussion: 

1. Vitest supports Istanbul or V8 as coverage tool. I opted for Istanbul for consistency with the server side code, but it appears V8 is the default and more efficient option. (That didn't seem super compelling to me with our small codebase. More info here: https://github.com/jestjs/jest/issues/11188.)
2. Note that the coverage report itself is slightly different than it was. (Notably, some files that I think we'd want coverage for are not included in the previous reports.) If continuity in our coverage stats is important, I can try to chase down the differences and reconcile them, but my assumption is the coverage data is right now just providing helpful comments on PRs. 
3. I switched from implicit globals (e.g., `describe`, `expect`, `it`, etc.) to explicit imports in our tests. I have a strong preference personally for explicit, since it makes code clearer and less magic-seeming for new developers. (And the minor downside of writing slightly more code should be minimized by your editor's automation.) That said, it's a meaningful change, so I wanted to check in on it.
   - Also worth mulling over: Do we want to follow up this PR with a similar PR moving the server-side tests to equivalent systems? (We wouldn't use vitest, but we could move over to jest for equivalent syntax and pleasant test-runner UX.) Do we want to move to explicit imports there too? 
4. Do we have any interest in taking this opportunity to colocate tests with the files under test? (e.g., `Foo.vue` and `Foo.spec.vue` would live in the same folder.) My personal conviction here isn't strong enough to warrant changing the status quo in this PR, but the argument is that colocated tests are easier to find, easier to work on, and perhaps make it easier for devs to write more tests.

## Screenshots / Demo Video
image
image

## Testing
- The unit test suite continues to pass
- So long as CI continues passing, that should cover things

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


.github/workflows/ci.yml
```diff
@@ -1,7 +1,8 @@
 name: Continuous Integration
 
 on:
-  pull_request_target: {}
+  # pull_request_target: {}
+  pull_request: {} # TEST TRIGGER - DO NOT MERGE
```
TylerHendrickson, line 5: Adding a comment here that will be made outdated by reverting ff47a47c520dfb453124999d6d8968cf3fac309d

TylerHendrickson (changes requested):
Changes look good; marking this as "Request changes" so that the CI trigger modification is reverted. Once that happens, I'll bypass the `main` branch requirement and merge the PR.

TylerHendrickson:
@jeffsmohan Responding to your points on the main PR comment:

> 1. Vitest supports Istanbul or V8 as coverage tool. I opted for Istanbul for consistency with the server side code, but it appears V8 is the default and more efficient option. (That didn't seem super compelling to me with our small codebase. More info here: https://github.com/jestjs/jest/issues/11188.)

Agree that performance arguments don't seem compelling here – GHA is spending all of [12 seconds running these tests](https://github.com/usdigitalresponse/usdr-gost/actions/runs/9178774712/job/25239477749?pr=3076#step:8:1) and <1 second generating the coverage report, which doesn't feel too egregious. FWIW, I don't think that the client- and server-side coverage tools _need_ to be the same, but I would like to retain the ability to include a coverage report in a [`github-actions` bot comment](https://github.com/usdigitalresponse/usdr-gost/pull/3076#issuecomment-2121038740). The [`fingerprintjs/action-coverage-report-md` action](https://github.com/fingerprintjs/action-coverage-report-md) would either need to be compatible with a different coverage tool, or we would want to find an alternative for that as well.

> 2. Note that the coverage report itself is slightly different than it was. (Notably, some files that I think we'd want coverage for are not included in the previous reports.) If continuity in our coverage stats is important, I can try to chase down the differences and reconcile them, but my assumption is the coverage data is right now just providing helpful comments on PRs.

I'm not certain I'm following here – are you just saying that this PR increases coverage in files where it was previously lacking? Or would you mind elaborating a bit?

> 3. I switched from implicit globals (e.g., describe, expect, it, etc.) to explicit imports in our tests. I have a strong preference personally for explicit, since it makes code clearer and less magic-seeming for new developers. (And the minor downside of writing slightly more code should be minimized by your editor's automation.) That said, it's a meaningful change, so I wanted to check in on it.
>    - Also worth mulling over: Do we want to follow up this PR with a similar PR moving the server-side tests to equivalent systems? (We wouldn't use vitest, but we could move over to jest for equivalent syntax and pleasant test-runner UX.) Do we want to move to explicit imports there too?

Agree with your preference, and yes, I think that having more code that's explicit/obvious and less magical is especially useful for our codebases.

> Do we have any interest in taking this opportunity to colocate tests with the files under test? (e.g., Foo.vue and Foo.spec.vue would live in the same folder.) My personal conviction here isn't strong enough to warrant changing the status quo in this PR, but the argument is that colocated tests are easier to find, easier to work on, and perhaps make it easier for devs to write more tests.

There can be weird implications for doing this in Python with good reason to avoid, but colocated tests are generally my preference. Are there any caveats we should be aware of / reasons why having tests organized in a mirrored directory structure (as they are today) would be beneficial? If not, I think it's worth changing (in a separate PR).

jeffsmohan:
> I'm not certain I'm following here – are you just saying that this PR increases coverage in files where it was previously lacking? Or would you mind elaborating a bit?

Scratch that, the coverage reports come out the same. I was looking at a report from before I had tweaked some settings. 

> There can be weird implications for doing this in Python with good reason to avoid, but colocated tests are generally my preference. Are there any caveats we should be aware of / reasons why having tests organized in a mirrored directory structure (as they are today) would be beneficial? If not, I think it's worth changing (in a separate PR).

Yeah, I think it's worth following language conventions when we don't have a good reason not to. For JS, it's well supported and most common to colocate the tests next to the code files. I'm not aware of any real downsides. I'll work up a follow-up PR to relocate the code.

jeffsmohan:
@TylerHendrickson thanks for the review! Responded to comments, reverted the workflow testing commit, and this should be ready to go.

TylerHendrickson (approved):

PR #3081, opened 2024-05-23
Fix: Missing `terraform validate` output from `stderr` in CI status reports

TylerHendrickson:
## Description

This PR fixes an issue where output is missing from `terraform validate` in CI report comments when the output is an error message. 

For example, when `terraform validate` succeeds, we get this:
```
Success! The configuration is valid.


```

But when the same command fails, we get this:
```


```

This PR updates the relevant GHA workflow's output to capture both stream outputs from the corresponding step:

- Success output:
  ```
  stdout:
  Success! The configuration is valid.
  -------------------------------------
  stderr:

  ```
- Error output:
  ```
  stdout:
  
  -------------------------------------
  stderr:
  
  Error: Unsupported attribute
  
    on main.tf line 68, in module "website":
    68:   namespace                = var.namespace.invalid
      ├────────────────
      │ var.namespace is a string
  
  Can't access attributes on a primitive-typed value (string).
  ```


.github/workflows/terraform-plan.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       fmt_outcome: ${{ steps.fmt.outcome }}
       init_outcome: ${{ steps.init.outcome }}
       validate_outcome: ${{ steps.validate.outcome }}
-      validate_output: ${{ steps.validate.outputs.stdout }}
+      validate_output: ${{ steps.validate.outputs.stdout || steps.validate.outputs.stderr }}
```
jeffsmohan, line 94: This looks extremely reasonable to me, and feel free to ignore this comment and merge. 

Only question is whether there's an easy way to ensure we get both stdout and stderr if both are present, or at least to preference stderr so we always make sure we're seeing errors. So either (a) flipping the order of the two arguments around the `||` operator, or (b) possibly using the [`join` expression GHA supports](https://docs.github.com/en/actions/learn-github-actions/expressions#join), though I have no experience with that.

jeffsmohan (approved):
Lgtm! One very optional comment.

pchj (approved):

.github/workflows/terraform-plan.yml
TylerHendrickson: Since there's no documented guarantee that only one of stderr or stdout will receive output (and at the very least, there are cases where `terraform plan` produces both), I updated this to always include output from both streams. You can see what this looks like in the updated [OP](https://github.com/usdigitalresponse/usdr-gost/pull/3081#issue-2311710957).

PR #3116, opened 2024-05-31
fix: project summaries v2 tab pulls metadata from last *EC* record

ajhyndman:
## Description

Fixes a bug in which project descriptions and expenditure category group could sometimes be missing from the audit report's "Project Summaries V2" tab.

While implementing the previous version of this feature, it slipped my mind that not all "records" generated from an upload are rows from the EC tabs.  Instead of blindly selecting the last record associated with a project and hoping that it is from an EC tab, this change filters for EC tab records explicitly and selects the last of those.

To test locally, upload the following input spreadsheet to your local instance and generate the audit report.

https://staging.grants.usdr.dev/arpa_reporter/uploads/ca1e3747-03fd-4477-9baf-aab9a911702c

## Screenshots / Demo Video

BEFORE

image

AFTER

image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -312,7 +312,7 @@ async function createReportsGroupedByProject(periodId, tenantId, dateFormat = RE
         const lastProjectRecord = projectSheetRecords[projectSheetRecords.length - 1];
         const row = {
             'Project ID': projectId,
-            'Project Description': lastProjectRecord?.content.Project_Description__c,
+            'Project Description': lastProjectRecord?.content.Project_Description__c ?? '[project not listed in any upload]',
             'Project Expenditure Category Group': ec(lastProjectRecord?.type),
             'Project Expenditure Category': lastProjectRecord?.subcategory,
             'Capital Expenditure Amount': 0,
--- a/packages/server/src/arpa_reporter/lib/audit-report.js
+++ b/packages/server/src/arpa_reporter/lib/audit-report.js
@@ -6,7 +6,7 @@ const XLSX = require('xlsx');
 const { PutObjectCommand } = require('@aws-sdk/client-s3');
 const { log } = require('../../lib/logging');
 const aws = require('../../lib/gost-aws');
-const { currencyNumeric, ec } = require('./format');
+const { EXPENDITURE_CATEGORIES, currencyNumeric, ec } = require('./format');
 
 const { getPreviousReportingPeriods, getAllReportingPeriods, getReportingPeriod } = require('../db/reporting-periods');
 const { getCurrentReportingPeriodID } = require('../db/settings');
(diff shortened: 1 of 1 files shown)


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -308,12 +308,13 @@ async function createReportsGroupedByProject(periodId, tenantId, dateFormat = RE
...
-        const lastProjectRecord = projectRecords[projectRecords.length - 1];
+        const projectSheetRecords = projectRecords.filter((record) => Object.keys(EXPENDITURE_CATEGORIES).includes(record.type));
+        const lastProjectRecord = projectSheetRecords[projectSheetRecords.length - 1];
         const row = {
             'Project ID': projectId,
-            'Project Description': lastProjectRecord.content.Project_Description__c,
-            'Project Expenditure Category Group': ec(lastProjectRecord.type),
-            'Project Expenditure Category': lastProjectRecord.subcategory,
+            'Project Description': lastProjectRecord?.content.Project_Description__c,
+            'Project Expenditure Category Group': ec(lastProjectRecord?.type),
```
tzinckgraf, line 316: What if this is null? Would that give an `undefined`? Do we want some sort of default value here?

packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -308,12 +308,13 @@ async function createReportsGroupedByProject(periodId, tenantId, dateFormat = RE
         projectLogger.debug('populating row from records in project');
 
         // set values for columns that are common across all records of projectId
-        const lastProjectRecord = projectRecords[projectRecords.length - 1];
+        const projectSheetRecords = projectRecords.filter((record) => Object.keys(EXPENDITURE_CATEGORIES).includes(record.type));
+        const lastProjectRecord = projectSheetRecords[projectSheetRecords.length - 1];
```
tzinckgraf, line 312: Can this ever be null, or is the null checking below for a type checker?

tzinckgraf (approved):
Approved, but my comments are just around null handling. Not sure if that's a problem that can happen.

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: `lastProjectRecord` can indeed be null!

It will only happen if someone uploads a spreadsheet with expenditures or awards that reference a project ID that hasn't been defined, but staging does have a test file on it that simulates that scenario, so I assume it happens.

I do think there are at least two options for handling that.  We could skip the row entirely, or we can show whatever information we do have.  I suspect more information is more useful than less, but that's probably something we could confirm with partners (or iterate on).

packages/server/src/arpa_reporter/lib/audit-report.js
ajhyndman: If `lastProjectRecord` is null, then these values become empty, yep.

I actually do like the idea of a default value... maybe something like `"project with ID not listed in any upload"`?

tzinckgraf (approved):

ClaireValdivia (approved):
after this is merged, I'll test in staging and confirm functionality!

PR #3152, opened 2024-06-10
Fix failing `dependency-review-action` jobs

TylerHendrickson:
## Description

This PR fixes an issue introduced by the recent `actions/dependency-review-action` upgrade to [v4.3.3](https://github.com/actions/dependency-review-action/releases/tag/v4.3.3) in #3135. New API endpoints are now being called by this action, and those domains were missing from our `dependency-review` job's egress policy. This PR adds the following allowed endpoints:
- `api.deps.dev:443`
- `api.securityscorecards.dev:443`

See also https://github.com/actions/dependency-review-action/issues/736.

### DIFF
--- a/.github/workflows/code-scanning.yml
+++ b/.github/workflows/code-scanning.yml
@@ -24,7 +24,9 @@ jobs:
           disable-sudo: true
           egress-policy: block
           allowed-endpoints: >
+            api.deps.dev:443
             api.github.com:443
+            api.securityscorecards.dev:443
             github.com:443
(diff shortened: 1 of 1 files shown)


as1729 (approved):
I had a bit of trouble finding where they were using `api.securityscorecards.dev` in the changelog but finally found it here: https://github.com/actions/dependency-review-action/blob/df5d74f5d3fc9748a904ea2f1dc6bdddea6439d6/src/scorecard.ts#L52

Looks good, thanks @TylerHendrickson !

PR #3160, opened 2024-06-11
Adding share terminology to the My Grants view

mayabose:
## Description
Implements the following changes to the `My Grants` page:
- The title and header of the `Assigned` tab is updated to `Shared With Your Team`
- The tab-based Vue router from `/my-grants/assigned` is updated to `/my-grants/shared-with-your-team`. The Vue router also redirects from `/assigned` to `/shared-with-your-team`.
- Tabs are reordered so that `Shared With Your Team` / `Assigned` is first. 

All of these changes (except for the order of the tabs) is dependent on the state of the `shareTerminologyEnabled` feature flag.

## Testing
The tab, title, and url of the `Assigned` tab is correctly updated to `Shared With Your Team` when the feature flag is set to `true`, and remains unchanged when the flag is set to `false`.

## Checklist
- [ ] Added PR reviewers


packages/client/src/router/index.js
```diff
@@ -103,6 +103,23 @@ export const routes = [
           }
         },
       },
+      {
+        path: '/my-grants/assigned',
+        name: 'assigned',
+        redirect: shareTerminologyEnabled() ? '/my-grants/shared-with-your-team' : undefined,
```
jeffsmohan, line 109: It's a bit more DRY to use the named route here  — `{ name: 'shared-with-your-team' }` — instead of the raw route string

packages/client/src/router/index.js
```diff
@@ -92,7 +92,7 @@ export const routes = [
         name: 'myGrants',
         component: () => import('../views/MyGrantsView.vue'),
         meta: {
-          tabNames: ['interested', 'assigned', 'not-applying', 'applied'],
+          tabNames: [shareTerminologyEnabled() ? 'shared-with-your-team' : 'assigned', 'interested', 'not-applying', 'applied'],
```
jeffsmohan, line 95: I'm seeing two minor issues here that I think we could improve with a small refactor. The issues: 

1. Now that we're including a ternary in the `tabNames` list here, it's feeling a bit more complex than would be ideal to inline in the meta definition here, as it's getting a bit hard to read. 
2. We really want the `/my-grants` route above to redirect to the first tab (as opposed to explicitly the interested tab. 

I think we can refactor this to solve both issues pretty neatly by determining the list of tab names elsewhere, like at the top of the file here. (Ideally, we might someday split up route definitions into multiple files and then we could keep this closer to the routes themselves. But for now, top of this file feels fine.) So...

```js
const myGrantsTabs = [
  shareTerminologyEnabled() ? 'shared-with-your-team' : 'assigned', 
  'interested', 
  'not-applying', 
  'applied',
];
```

And then the `/my-grants` route can redirect based on the first tab: `redirect: { name: 'myGrants', params: { tab: myGrantsTabs[0] } }`. And the tabNames here can simply reference `myGrantsTabs`.

packages/client/src/views/MyGrantsView.vue
```diff
@@ -7,20 +7,27 @@
       style="margin-top: 1.5rem"
       lazy
     >
-      <b-tab title="Interested">
+      <b-tab v-if="shareTerminologyEnabled" title="Shared With Your Team">
```
jeffsmohan, line 10: Have you set up linting and auto-formatting in your editor yet? You should hopefully be able to get running with auto-fixing for lint errors like what's showing up on this and the line below. 

Bit of documentation is here, but I'm happy to help as well if you need! https://github.com/usdigitalresponse/usdr-gost/blob/main/docs/development.md#linting

jeffsmohan: That aside, given that the functionality doesn't really change, but only the tab title, it's a bit more DRY here to just make the titles dynamic: 

```js
<b-tab :title="shareTerminologyEnabled ? 'Shared With Your Team' : 'Assigned'">
  <GrantsTable
    :search-title="shareTerminologyEnabled ? 'Shared With Your Team' : 'Assigned'"
    :show-assigned-to-agency="selectedAgencyId"
    :show-search-controls="false"
  />
</b-tab>
```

You could also turn the `shareTerminologyEnabled ? 'Shared With Your Team' : 'Assigned'` into a computed property if you want, but one repetition I'm on the fence about whether it's worth it.

jeffsmohan (changes requested):
This is looking great! Totally works, but I had a few suggested tweaks to make things a bit more readable/maintainable. Let me know if anything's unclear, and I'm happy to talk it through.

packages/client/src/views/MyGrantsView.vue
mayabose: Thank you for the detailed response and code suggestions!

packages/client/src/router/index.js
```diff
@@ -103,6 +110,23 @@ export const routes = [
           }
         },
       },
+      {
```
jeffsmohan, line 113: Sorry, I missed this in my first review, but these new route definitions are never actually being hit. `vue-router` (in Vue 2, at least) evaluates a URL against the routes in order and uses the first matching route. So the above `/my-grants/:tab` always matches before getting to these more-specific routes. 

If we want the gating and redirecting defined below to work, you'll need to move these two new route definitions above the `/my-grants/:tab` route. 

(Interesting note: If we migrate to Vue 3 soon, that version of vue-router actually has a path ranking algorithm that would have identified your new routes as more specific than the generic tab route, and would have prioritized them regardless of the order. [Example evaluation here](https://paths.esm.dev/?p=AAMeJWyBwRMcaE7AMASAHQxcAe65cxgBAQUIDLahacqWgKDygCkkJhyFpUmYuGwBC5EBWGvd66MAGgB7ia6kY5JsIBEAAA..#).)

packages/client/src/router/index.js
```diff
@@ -103,6 +110,23 @@ export const routes = [
...
+        path: '/my-grants/assigned',
+        name: 'assigned',
+        redirect: { name: 'shared-with-your-team' },
```
jeffsmohan, line 116: I think in the last set of changes, you lost the flag check here. We want to only redirect if the new terminology flag is on.

jeffsmohan (changes requested):
Great stuff, just about there! I found something I missed in my first review (sorry).

packages/client/src/router/index.js
masimons: This might be a good justification for writing a test for the changes made to this route :)

jeffsmohan:
@mayabose If this is ready for a re-review, please hit the "re-request review" button so I know when it's ready for me to take another look! (Otherwise, I'm not sure if you're maybe working on writing tests like @masimons proposed, for example.)

jeffsmohan (approved):
lgtm!

PR #3162, opened 2024-06-11
feat: update email preferences copy

masimons:
## Description
This change is part of the Assign-to-Share project. We want the copy within the Email Notifications settings to change if the feature flag `shareTerminologyEnabled` is set to true.

## Screenshots / Demo Video
"Assign" copy:
image

"Share" copy:
image

## Testing
I verified this change locally. `MyProfileView.vue` doesn't seem to have a corresponding test file yet. I've been on engineering teams where the process in this case would be to create a separate chore to write tests for `MyProfileView.vue` (especially since the change made in this PR is small). But if the process here is to commit a test without exception, then I may ask to pair briefly with someone on a unit test for this code (I'm not super familiar with Vue.js, and I hit my timebox for getting a test working on my own).

Manual steps for reviewer:
Navigate to `My Profile -> Email Notifications` to verify the copy change (the `shareTerminologyEnabledFlag` will need to be manually set to false/true in the code to see the change).

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/views/MyProfileView.vue
+++ b/packages/client/src/views/MyProfileView.vue
@@ -72,6 +72,7 @@
 
 <script>
 import { mapGetters, mapActions } from 'vuex';
+import { shareTerminologyEnabled } from '@/helpers/featureFlags';
 import EditUserModal from '@/components/Modals/EditUser.vue';
 import UserAvatar from '@/components/UserAvatar.vue';
 
@@ -81,6 +82,10 @@ export default {
     UserAvatar,
   },
   data() {
+    const shareTerminologyEnabledFlag = shareTerminologyEnabled() === true;
+    const grantAssignmentDescription = 'Send me notifications if a grant has been assigned to my USDR Grants team.';
+    const grantShareDescription = 'Send me an email notification when someone shares a grant with my team.';
+
     return {
       prefs: [
         {
@@ -89,9 +94,9 @@ export default {
           description: 'Send me a daily email if new grants match my saved search(es).',
         },
         {
-          name: 'Grants Assignment',
+          name: shareTerminologyEnabledFlag ? 'Shared Grants' : 'Grants Assignment',
           key: 'GRANT_ASSIGNMENT',
-          description: 'Send me notifications if a grant has been assigned to my USDR Grants team.',
+          description: shareTerminologyEnabledFlag ? grantShareDescription : grantAssignmentDescription,
         },
       ],
       breadcrumb_items: [
--- a/packages/client/src/views/MyProfileView.spec.js
+++ b/packages/client/src/views/MyProfileView.spec.js
@@ -1,0 +1,59 @@
+import {
+  describe, it, beforeEach, expect, vi,
+} from 'vitest';
+import { createLocalVue, shallowMount } from '@vue/test-utils';
+import Vuex from 'vuex';
+import MyProfileView from '@/views/MyProfileView.vue';
+import { shareTerminologyEnabled } from '@/helpers/featureFlags';
+
+describe('MyProfileView.vue', () => {
+  const localVue = createLocalVue();
+  localVue.use(Vuex);
+  const store = new Vuex.Store({
+    getters: {
+      'users/loggedInUser': () => ({
+        emailPreferences: {
+          GRANT_ASSIGNMENT: 'SUBSCRIBED',
+          GRANT_INTEREST: 'SUBSCRIBED',
+          GRANT_DIGEST: 'SUBSCRIBED',
+          GRANT_FINDER_UPDATES: 'SUBSCRIBED',
+        },
+      }),
+    },
+  });
+
+  vi.mock('@/helpers/featureFlags', async (importOriginal) => ({
+    ...await importOriginal(),
+    shareTerminologyEnabled: vi.fn(),
+  }));
+
+  describe('when share terminology flag is off', () => {
+    beforeEach(() => {
+      vi.mocked(shareTerminologyEnabled).mockReturnValue(false);
+    });
+
+    it('should show assign terminology', () => {
+      const wrapper = shallowMount(MyProfileView, { store, localVue });
+      const text = wrapper.text();
+      expect(text).toContain('Grants Assignment');
+      expect(text).toContain(
+        'Send me notifications if a grant has been assigned to my USDR Grants team.',
+      );
+    });
+  });
+
+  describe('when share terminology flag is on', () => {
+    beforeEach(() => {
+      vi.mocked(shareTerminologyEnabled).mockReturnValue(true);
+    });
+
+    it('should show share terminology', () => {
+      const wrapper = shallowMount(MyProfileView, { store, localVue });
+      const text = wrapper.text();
+      expect(text).toContain('Shared Grants');
+      expect(text).toContain(
+        'Send me an email notification when someone shares a grant with my team.',
+      );
+    });
+  });
+});


TylerHendrickson (approved):
@masimons TL;DR Approving this because I think the change looks good overall and it's working for me in manual testing. I do think there's some work to do around testing (see below) but I don't know that it needs to hold up this PR. 

Regarding tests: we don't currently have any hard rule that says all Vue template changes need automated tests to be accepted. However, I think it's a good idea to try adding them as we go. Ideally, this change would be tested in a `MyProfileView.spec.js` file. 

I checked out this PR branch and tried my hand at writing some tests; a few observations to share (some or all of which you no doubt also encountered):
- Because the `shareTerminologyEnabled()` function from the `featureFlags` module is called within the Vue template's `data()` export, we don't have a way to configure a flag value when mounting a `Vuex.Store` during test setup. By contrast, in `TeamsView.spec.js`, the `newTerminologyEnabled` feature flag value is set by configuring a `computed` method for the view. 
- Simply creating a `computed` helper method like `shareTerminologyEnabled` in `MyProfileView.vue` wouldn't be all that useful, since `this.shareTerminologyEnabled()` wouldn't be callable from the `data()` export. 
- I'm wondering if the "correct" way to do this would be to perform a more intensive refactoring of the `MyProfileView.vue` template, i.e. that utilizes a custom component for the email toggles or else removes the `<v-for="pref in prefs">` loop in favor of templating each preference toggle explicitly, which would allow us to evaluate the feature flag in the template logic rather than when `data()` is executed.
- All that said, I **was** (almost) able to get tests working without making any changes to `MyProfileView.vue` by calling `localStorage.set('featureFlags', 'Mock JSON-serialized feature flag here')` (see below). I also think there's a similar approach here that could work by mocking the `@/helpers/featureFlag` module's exported `shareTerminologyEnabled()` function, but I wasn't quite able to figure out Vitest's mocking system.
  - Note: I say "almost" because I wasn't able to identify a good selector to use when defining the `prefName` variable. Presumably this could be fixed by tweaking the `id` attributes in the template a bit, e.g. by adding something like `:id="pref.key"` in the `<b-row>` element that loops on `v-for="pref in prefs"`.

<details>
<summary>Expand for example test file</summary>

```js
import {
  describe, beforeEach, afterEach, it, expect,
} from 'vitest';
import { createLocalVue, shallowMount } from '@vue/test-utils';
import Vuex from 'vuex';
import MyProfileView from '@/views/MyProfileView.vue';

const localVue = createLocalVue();
localVue.use(Vuex);

let store;
let wrapper;

const stubs = [];
const noOpGetters = {
  'users/loggedInUser': () => undefined,
};
const noOpActions = {
  'users/updateEmailSubscriptionPreferencesForLoggedInUser': () => {},
};
afterEach(() => {
  store = undefined;
  wrapper = undefined;
});

describe('MyProfileView.vue', () => {
  describe('shareTerminologyEnabled flag is enabled', () => {
    const originalSessionStorageFeatureFlags = sessionStorage.getItem('featureFlags');
    beforeEach(() => {
      sessionStorage.setItem(
        'featureFlags',
        JSON.stringify({
          ...(originalSessionStorageFeatureFlags || {}),
          shareTerminologyEnabled: true,
        }),
      );
      store = new Vuex.Store({
        getters: {
          ...noOpGetters,
          'users/loggedInUser': () => ({
            name: 'Someone',
            emailPreferences: { GRANT_ASSIGNMENT: 'SUBSCRIBED' },
          }),
        },
        actions: {
          ...noOpActions,
        },
      });
      afterEach(() => {
        sessionStorage.setItem(
          'featureFlags',
          originalSessionStorageFeatureFlags,
        );
      });
      wrapper = shallowMount(MyProfileView, {
        store,
        localVue,
        stubs,
        computed: {
          shareTerminologyEnabled: () => true,
        },
      });
    });
    it('should present GRANT_ASSIGNMENT email preferences with share terminology', () => {
      expect(JSON.parse(window.sessionStorage.getItem('featureFlags'))).toEqual(
        { shareTerminologyEnabled: true },
      );
      const prefName = wrapper.find('!!!Could not figure out a selector!!!');
      expect(prefName.text()).toEqual('Shared Grants');
    });
  });
});
```

</details>

At this point, my feeling is that it **is** worth having tests, but there's enough to unpack here that it makes more sense to (as you suggested) create a separate chore issue to discuss a path forward and encapsulate this work. Before I totally commit to that, I do want to see if @jeffsmohan has anything to add here, because he's much more familiar with Vue and Vitest than I am.

jeffsmohan:
Oh this brings up more feelings for me than expected, lol. 😝 (To be clear, absolutely nothing to do with your approach here, @masimons  — I love the timebox and then check in tactic!) 

The categories of my feelings: 

1. **Feature flags.** I'd love if we had a more robust ecosystem around our feature flags. (Ability to flip them without a code deploy, nice helpers to toggle them in tests, etc.) That's obviously a much bigger topic, but it sure would be nice to have a simple helper method for tests here. 
2. **Writing tests.** Totally agree that we'd benefit from better frontend test coverage! That said, a lot of kinds of frontend changes really don't lend themselves to good tests (changing a background color, reordering stuff on the page, etc.), so I tend to favor a bias toward testing rather than an iron-clad rule that all PRs must come with tests. In specific for something like the change in this PR, it's on the edge but I personally don't tend to think tests are worth it — given the simplicity of the logic involved, it verges on writing a test for `expect(1 + 1).toEqual(2)`, where you're kind of just testing whether core language functionality works, as opposed to testing your own business logic. (This is just my personal viewpoint, fwiw... there's lots of room for other opinions on testing, and I'd definitely defer to USDR's norms here!)
3. **Non-reactive data in `data`.** The definition of `prefs` and `breadcrumb_items` in the `data` option for the `MyProfileView` component is not really the cleanest. Ideally, the `data` option is where you define reactive data for your component, but here it's just data structures. Vue 3 has some better idioms for this, but in Vue 2, the best options are:
   - Write it out in the template (and factor out a new component if it's getting too repetitive) like @TylerHendrickson suggested:
     ```html
     // For prefs
     <CheckboxInput name="New Grant Digest" key="GRANT_DIGEST" description="Send me a daily..." />
     <CehckboxInput name="Grants Assignment" key="GRANT_ASSIGNMENT" description="Send me notifications..." />
     // For breadcrumbs
     <b-breadcrumb :items="[
       { text: 'Home', to: 'grants' },
       { text: 'My Profile', href: '#' },
     ]" />
     ```
   - Stash it directly in the component options (bit awkward in Vue 2, but for very large/complex values this can be preferable to the above):
     ```js
     // define the data directly on the component options
     {
       data() { ... },
       prefs: [ ... ], // the existing prefs value currently defined in data
       ...
     }
     // use the data in the template via `$options`
     <b-row v-for="pref in $options.prefs>...</b-row>
     ```
4. **Testing feature flags.** For now, the cleanest way to test functionality behind a feature flag is to mock the flag at the helper level. Here's a working rough draft of tests for the functionality in this PR, for example: 
    ```js
    import {
      describe, it, expect, vi,
    } from 'vitest';
    import { createLocalVue, shallowMount } from '@vue/test-utils';
    import Vuex from 'vuex';
    import MyProfileView from '@/views/MyProfileView.vue';
    import { shareTerminologyEnabled } from '@/helpers/featureFlags';

    describe('MyProfileView', () => {
      // Set up a mock store for testing
      const localVue = createLocalVue();
      localVue.use(Vuex);
      const store = new Vuex.Store({
        getters: {
          'users/loggedInUser': () => ({
            emailPreferences: {
              GRANT_ASSIGNMENT: 'SUBSCRIBED',
              GRANT_INTEREST: 'SUBSCRIBED',
              GRANT_DIGEST: 'SUBSCRIBED',
              GRANT_FINDER_UPDATES: 'SUBSCRIBED',
            },
          }),
        },
      });

      // Mock the relevant flag
      vi.mock('@/helpers/featureFlags', async (importOriginal) => ({
        ...await importOriginal(),
        shareTerminologyEnabled: vi.fn(),
      }));

      it('uses assign terminology when flag is off', () => {
        vi.mocked(shareTerminologyEnabled).mockReturnValue(false);
        const wrapper = shallowMount(MyProfileView, { store, localVue });
        expect(wrapper.text()).toContain('Grants Assignment');
      });

      it('uses share terminology when flag is on', () => {
        vi.mocked(shareTerminologyEnabled).mockReturnValue(true);
        const wrapper = shallowMount(MyProfileView, { store, localVue });
        expect(wrapper.text()).toContain('Shared Grants');
      });
    });
    ```

Sorry for the essay! Upshot is, if you wanted to add tests in this PR, I don't think it's too onerous, fft grab the test code above and finesse it just a bit, and you should be good to go. :)

packages/client/src/views/MyProfileView.vue
```diff
@@ -89,9 +94,9 @@ export default {
           description: 'Send me a daily email if new grants match my saved search(es).',
         },
         {
-          name: 'Grants Assignment',
+          name: `${shareTerminologyEnabledFlag ? 'Shared Grants' : 'Grants Assignment'}`,
           key: 'GRANT_ASSIGNMENT',
-          description: 'Send me notifications if a grant has been assigned to my USDR Grants team.',
+          description: `${shareTerminologyEnabledFlag ? grantShareDescription : grantAssignmentDescription}`,
```
jeffsmohan, line 99: (minor nit, optional and non-blocking) The js format string with the backticks are unnecessary for the `name` and `description` here, since we're not really interpolating anything. You can just do e.g. `name: shareTerminologyEnabledFlag ? 'Shared Grants' : 'Grants Assignment', ...`

jeffsmohan (approved):
Nice work, change also lgtm as-is, fwiw! Though I have a small optional improvement, and I left a giganto comment that ends with how to test this change if you choose to. 😸

jeffsmohan (approved):
Updates lgtm! You'll just need to get things linted so the check passes. Have you got auto-formatting set up for your editor? A bit of info can be found in our [docs here](https://github.com/usdigitalresponse/usdr-gost/blob/main/docs/development.md#linting), but I'm happy to talk you through it as the docs aren't exactly extensive.

masimons:
@jeffsmohan a) Thank you SO MUCH for the time you invested in your response, and for sharing working tests and great insight. Really appreciate it 🙏 
b) I just rebased main, and ran the linter, so I'm pretty sure this is the final version for reals and that once you re-re-approve this PR, it will be the last time haha
c) Indeed I have not set up the auto-formatting for vs code yet, thank you for sharing that doc 👍 
d) Since I've got you here, from what I understand, merging a PR will deploy to the staging env, NOT production. Just want to make sure this is the case, because I have NOT tested this on staging (in the case that deploys to staging must be done before merge)

jeffsmohan (approved):
lgtm! Thanks for asking good questions and reading through the very long answers :)

PR #3175, opened 2024-06-14
Include assignee name in Recent Activity view

masimons:
## Description
The Recent Activity log includes information about interested grants, rejected grants, assigned grants, etc. This change updates the language around assigned grants and includes the name of the user who shared the grant.

## Screenshots / Demo Video
Last item in the Recent Activity table shows that a user shared a grant.
image

## Testing
Navigate to the Recent Activity log (`/dashboard`). In order for an assigned grant to show up, the record in the `assigned_grants_agency` table must have a non-null `assigned_by` value.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/components/ActivityTable.spec.js
+++ b/packages/client/src/components/ActivityTable.spec.js
@@ -1,0 +1,44 @@
+import {
+  beforeEach, describe, it, expect,
+} from 'vitest';
+import { createLocalVue, mount } from '@vue/test-utils';
+import { BootstrapVue } from 'bootstrap-vue';
+import ActivityTable from '@/components/ActivityTable.vue';
+
+let wrapper;
+const stubs = ['b-icon'];
+const localVue = createLocalVue();
+localVue.use(BootstrapVue);
+
+describe('ActivityTable.vue', () => {
+  describe('when view is loaded', () => {
+    beforeEach(() => {
+      wrapper = mount(ActivityTable, {
+        localVue,
+        stubs,
+        propsData: {
+          grantsInterested: [],
+          onRowSelected: () => {},
+          onRowClicked: () => {},
+        },
+      });
+    });
+
+    it('displays information about interested grants', async () => {
+      await wrapper.setProps({
+        grantsInterested: [{
+          created_at: '2024-06-16T00:00:00.000Z', name: 'USDR', status_code: 'Rejected', interested_name: 'Will Apply', agency_id: 0, title: 'Rejected Grant', grant_id: '666997', assigned_by: null, assigned_by_user_name: null,
+        }, {
+          created_at: '2024-06-15T00:00:00.000Z', name: 'USDR', status_code: 'Interested', interested_name: 'Will Apply', agency_id: 0, title: 'Test Grant 666999', grant_id: '666999', assigned_by: null, assigned_by_user_name: null,
+        }, {
+          created_at: '2024-06-20T00:00:00.000Z', name: 'USDR', status_code: null, interested_name: null, agency_id: 0, title: 'EAR Postdoctoral Fellowships', grant_id: '333334', assigned_by: 17, assigned_by_user_name: 'USDR Staff',
+        }],
+      });
+
+      const text = wrapper.text();
+      expect(text).toContain('USDR  rejected Rejected Grant');
+      expect(text).toContain('USDR  is  interested  in Test Grant 666999');
+      expect(text).toContain('USDR Staff shared  EAR Postdoctoral Fellowships with USDR');
+    });
+  });
+});
--- a/packages/client/src/components/ActivityTable.vue
+++ b/packages/client/src/components/ActivityTable.vue
@@ -59,7 +59,7 @@
       </div>
     </template>
     <template #cell(agencyAndGrant)="agencies">
-      <div>
+      <div v-if="agencies.item.interested !== statuses.assigned">
         {{ agencies.item.agency }}
         <span
           v-if="agencies.item.interested === statuses.rejected"
@@ -70,7 +70,6 @@
             <strong> interested </strong>
           </span> in
         </span>
-        <span v-if="agencies.item.interested === statuses.assigned"> was<strong> assigned </strong> </span>
         <span v-if="agencies.item.interested === statuses.awarded"> was<strong><span
           class="color-yellow"
         > awarded </span></strong> </span>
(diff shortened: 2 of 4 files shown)


packages/client/src/components/ActivityTable.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  beforeEach, describe, it, expect,
+} from 'vitest';
+import Vuex from 'vuex';
+import { createLocalVue, shallowMount } from '@vue/test-utils';
+import ActivityTable from '@/components/ActivityTable.vue';
+
+const localVue = createLocalVue();
+localVue.use(Vuex);
+let wrapper;
+const stubs = ['b-table', 'b-icon', 'b-iconstack'];
```
masimons, line 11: bit confused about the pattern of stubbing these out

packages/client/src/components/ActivityTable.spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('ActivityTable.vue', () => {
+  describe('when view is loaded', () => {
+    beforeEach(() => {
+      wrapper = shallowMount(ActivityTable, {
```
masimons, line 16: I dont _think_ that mocks or store are needed for this file? maybe?

masimons:
@jeffsmohan This PR is in super draft mode, but I was wondering if you could help nudge me in the right direction re: the ActivityTable test? And perhaps your thoughts on how to tackle a test for the db query, which I've only given superficial thought to so far (or maybe we just make a chore for that whole file). Thanks in advance for your guidance!

packages/client/src/components/ActivityTable.spec.js
masimons: omitting stubs entirely doesn't seem to change the behavior anyways tho, at least with current setup

jeffsmohan: I agree, you can omit these stubs. (They might provide some minor performance benefit in preventing a full mount from rendering everything, but it's pretty negligible overall.) And, in fact, for your test you need `b-table` to render to evaluate what you want to test.

packages/client/src/components/ActivityTable.spec.js
jeffsmohan: Agreed!

jeffsmohan:
@masimons This one was a bit tough to puzzle out (I'm still getting my head fully around the testing setup here). For your test to work here, you need to actually render the `b-table` component, which in turn requires a few things: 

1. Removing the `b-table` stub. (As you noted, you can just remove all of them, they're not doing a whole lot.)
2. Rendering using `mount` rather than `shallowMount`. (Shallow mounting will run faster, so it's great to use when possible, but in this case we're actually trying to test how the `b-table` sub-component renders out, so we need to do a full mount.)
3. Install the `VueBootstrap` library on the local Vue instance you're using, so Vue knows about the `b-table` component. (We could consider installing this globally for our test suite so it doesn't have to be done test by test, but it'd be worth investigating how much overhead that adds to tests that don't actually need it. Presumably not much?)

Bringing it all together, you get something like this, which is passing for me locally: 

```js
import {
  beforeEach, describe, it, expect,
} from 'vitest';
import { createLocalVue, mount } from '@vue/test-utils';
import { BootstrapVue } from 'bootstrap-vue';
import ActivityTable from '@/components/ActivityTable.vue';

let wrapper;
const localVue = createLocalVue();
localVue.use(BootstrapVue);

describe('ActivityTable.vue', () => {
  describe('when view is loaded', () => {
    beforeEach(() => {
      wrapper = mount(ActivityTable, {
        localVue,
        propsData: {
          grantsInterested: [],
          onRowSelected: () => {},
          onRowClicked: () => {},
        },
      });
    });

    it('should contain <div>', async () => {
      await wrapper.setProps({
        grantsInterested: [{
          created_at: '2024-06-19T00:00:00.000Z', name: 'USDR', status_code: 'Interested', interested_name: 'Will Support', agency_id: 0, title: 'Test Grant 666656', grant_id: '666656', assigned_by: null, assigned_by_user_name: null,
        }, {
          created_at: '2024-06-18T00:00:00.000Z', name: 'USDR', status_code: 'Interested', interested_name: 'Will Support', agency_id: 0, title: 'Test Grant 666666', grant_id: '666666', assigned_by: null, assigned_by_user_name: null,
        }, {
          created_at: '2024-06-17T00:00:00.000Z', name: 'USDR', status_code: 'Interested', interested_name: 'Will Apply', agency_id: 0, title: 'Community Health Aide Program:  County Planning & Implementation', grant_id: '333333', assigned_by: null, assigned_by_user_name: null,
        }, {
          created_at: '2024-06-15T00:00:00.000Z', name: 'USDR', status_code: 'Interested', interested_name: 'Will Apply', agency_id: 0, title: 'Test Grant 666999', grant_id: '666999', assigned_by: null, assigned_by_user_name: null,
        }],
      });
      console.log(wrapper.html());
      expect(wrapper.contains('div')).toBe(true);
    });
  });
});
```

Note that I also switched to defining the props rather than mocking out the computed method, since it's generally most effective to test components as black box-y unit tests... props go in, HTML comes out.

jeffsmohan:
Testing the db query change should hopefully be a bit more straightforward. If you look in `db.test.js`, you'll see there are already some tests for `getGrantsInterested`. You'll probably just update those with some additional `expect`s verifying the new information you're querying. 

The only other thing you might have to consider is whether the existing db seed data used for the server side tests contains the right information to test your change. If not, you may need to tweak what's in `__tests__/db/seeds/fixtures.js` to fit your needs.

packages/client/src/components/ActivityTable.spec.js
masimons: thanks for confirming - yeah, what you're seeing was my latest iteration of throwing spaghetti at the wall and seeing if the behavior of the test changed to try to understand it better. Most of the time, I swear I wasn't stubbing out the `b-table` :D

masimons:
@jeffsmohan Sorry for the delay in looping back around on this - a particularly unpleasant cold wiped me out for the better part of the past two days.

Infinite thanks (again) for sharing a working test.

And I appreciate the explanation about the `shallowMount` vs `mount` differences, as well as defining props vs passing computed data. Makes total sense.

I'm still wrapping my head around BootstrapVue and Vuex (and testing Vue components in general :D). Looks like the ActivityTable spec we add here is currently the only test file to import `BootstrapVue` - as such, it may make sense to hold off on installing globally for the test suite? Or perhaps you meant something different.

I've cleaned up the PR, if you'd be willing to take a glance at your convenience. But I will also ping the grants-developers group.

jeffsmohan:
Oh no, that sucks. I hope you're feeling better! 

> I'm still wrapping my head around BootstrapVue and Vuex (and testing Vue components in general :D). Looks like the ActivityTable spec we add here is currently the only test file to import `BootstrapVue` - as such, it may make sense to hold off on installing globally for the test suite? Or perhaps you meant something different.

Yeah, totally agree! I wouldn't go about doing any global test changes just yet; I just wanted to flag that I'm not super happy that individual tests have to worry about whether or not they've installed this VueBootstrap library that we've got installed globally. Something to look into down the road as we write more tests.

jeffsmohan (approved):
This lgtm! Thanks for caring about getting everything well tested, I really appreciate that!

PR #3184, opened 2024-06-18
Grant Details Page - move grants.gov, copy, print buttons 

amyewang:
## Description
Update the Grant Details page main actions buttons to match expected state in the [Figma designs](https://www.figma.com/design/L2gfVMBH0x1daOlOe50sJl/Follow-%2F-Leave-a-Note?node-id=1101-13791&t=Co6m063mfZx7NQ1c-0)  
Moved to the right column, updated text and icons, updated "copy" and "print" to be links. 

## Screenshots / Demo Video
image

## Testing
Tested various window widths in local dev env 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/client/src/views/GrantDetailsView.vue
+++ b/packages/client/src/views/GrantDetailsView.vue
@@ -34,12 +34,11 @@
           <div class="d-flex justify-content-between align-items-center">
             <b-button
               variant="primary"
-              block
-              class="mr-5"
-              style="width: fit-content; white-space: nowrap; "
+              class="mr-5 text-nowrap"
               :href="`https://www.grants.gov/search-results-detail/${currentGrant.grant_id}`"
               target="_blank"
               rel="noopener noreferrer"
+              size="sm"
               data-dd-action-name="view on grants.gov"
               @click="onOpenGrantsGov"
             >
@@ -52,35 +51,33 @@
             </b-button>
             <div class="w-20 d-flex">
               <a
-                class="link-primary"
+                class="link-primary text-nowrap"
                 role="button"
                 :variant="copyUrlSuccessTimeout === null ? 'outline-primary' : 'outline-success'"
                 data-dd-action-name="copy btn"
-                style="white-space: nowrap; "
                 @click="copyUrl"
               >
                 <b-icon
                   :icon="copyUrlSuccessTimeout === null ? 'paperclip' : 'check2'"
                   aria-hidden="true"
-                  class="mr-2"
+                  class="mr-1"
                 />
                 <span v-if="copyUrlSuccessTimeout === null">Copy Link</span>
                 <span v-else>Link Copied</span>
               </a>
-              <div class="col-1 border-right border-dark p-2"/>
-              <div class="col-1 p-2"/>
+              <div class="col-1 border-right border-dark p-2" />
+              <div class="col-1 p-2" />
               <a
-                class="link-primary"
+                class="link-primary text-nowrap"
                 role="button"
                 variant="outline-primary"
                 data-dd-action-name="print btn"
-                style="white-space: nowrap; "
                 @click="printPage"
               >
                 <b-icon
                   icon="printer"
                   aria-hidden="true"
-                  class="mr-2"
+                  class="mr-1"
                 />
                 Print
               </a>
--- a/packages/client/src/views/GrantDetailsView.vue
+++ b/packages/client/src/views/GrantDetailsView.vue
@@ -29,6 +29,65 @@
           {{ currentGrant.title }}
         </h2>
 
+        <!-- Left page column: main print/copy/grants.gov buttons -->
+        <div class="grant-details-main-actions print-d-none">
+          <div class="d-flex justify-content-between align-items-center">
+            <b-button
+              variant="primary"
+              block
+              class="mr-5"
+              style="width: fit-content; white-space: nowrap; "
+              :href="`https://www.grants.gov/search-results-detail/${currentGrant.grant_id}`"
+              target="_blank"
+              rel="noopener noreferrer"
+              data-dd-action-name="view on grants.gov"
+              @click="onOpenGrantsGov"
+            >
+              <b-icon
+                icon="box-arrow-up-right"
+                aria-hidden="true"
+                class="mr-2"
+              />
+              View on Grants.gov
+            </b-button>
+            <div class="w-20 d-flex">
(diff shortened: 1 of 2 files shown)


packages/client/src/views/GrantDetailsView.vue
```diff
@@ -29,6 +29,65 @@
...
+              />
+              View on Grants.gov
+            </b-button>
+            <div class="w-20 d-flex">
+              <a
+                class="link-primary"
+                role="button"
+                :variant="copyUrlSuccessTimeout === null ? 'outline-primary' : 'outline-success'"
+                data-dd-action-name="copy btn"
+                style="white-space: nowrap; "
```
jeffsmohan, line 59: In general, we should avoid inline styles where possible, since they're harder to maintain, keep consistent, override, etc. Instead, we want to favor existing [utility classes](https://bootstrap-vue.org/docs/reference/utility-classes) where available, or styles defined in a `style` section of the component otherwise. 

In this case, you should be able to use the `text-nowrap` utility class provided by Bootstrap. ([Docs here](https://getbootstrap.com/docs/4.5/utilities/text/#text-wrapping-and-overflow))

packages/client/src/views/GrantDetailsView.vue
```diff
@@ -29,6 +29,65 @@
...
         </h2>
 
+        <!-- Left page column: main print/copy/grants.gov buttons -->
+        <div class="grant-details-main-actions print-d-none">
+          <div class="d-flex justify-content-between align-items-center">
+            <b-button
+              variant="primary"
+              block
+              class="mr-5"
+              style="width: fit-content; white-space: nowrap; "
```
jeffsmohan, line 39: A cleaner way to achieve the new layout, using existing bootstrap helpers: 

1. Remove the `block` option (instead of the `width: fit-content` style)
2. Add the `text-nowrap` utility class (instead of the `white-space: nowrap` style)
3. You can remove the `mr-5` class now that we don't need a right margin on the button

packages/client/src/views/GrantDetailsView.vue
```diff
@@ -29,6 +29,65 @@
...
+                <span v-else>Link Copied</span>
+              </a>
+              <div class="col-1 border-right border-dark p-2"/>
+              <div class="col-1 p-2"/>
+              <a
+                class="link-primary"
+                role="button"
+                variant="outline-primary"
+                data-dd-action-name="print btn"
+                style="white-space: nowrap; "
```
jeffsmohan, line 77: Same as above

packages/client/src/views/GrantDetailsView.vue
```diff
@@ -29,6 +29,65 @@
...
+              >
+                <b-icon
+                  :icon="copyUrlSuccessTimeout === null ? 'paperclip' : 'check2'"
+                  aria-hidden="true"
+                  class="mr-2"
+                />
+                <span v-if="copyUrlSuccessTimeout === null">Copy Link</span>
+                <span v-else>Link Copied</span>
+              </a>
+              <div class="col-1 border-right border-dark p-2"/>
```
jeffsmohan, line 70: Seeing some linting errors here. Ideally you shouldn't have to think about style stuff like this, and you can get your editor set up to auto-format on save. A bit of documentation on getting set up is here — https://github.com/usdigitalresponse/usdr-gost/blob/main/docs/development.md#linting — but lmk if you have any trouble getting that working!

jeffsmohan (changes requested):
This is looking really good! Just a few minor cleanups in inline comments, plus: 

- Might be worth checking with the designers, but to my eye I think the designs had the button going down a size smaller, plus the margin between the button/links and the rest of the content being a bit smaller as well

packages/client/src/views/GrantDetailsView.vue
amyewang: thanks! getting used to Bootstrap

packages/client/src/views/GrantDetailsView.vue
amyewang: image

I left a margin to the right of the button to add some negative space between "view" and "copy" in the case of the minimum viewport width - this wasn't really specified in the designs though. Let me know what you think

amyewang:
Thanks for the review @jeffsmohan ! I fixed the inline styling, updated the overall button & link sizing (thanks for that callout), and updated my linting. Left 1 small comment about the button margin.

packages/client/src/views/GrantDetailsView.vue
jeffsmohan: Sigh... me too... :P

packages/client/src/views/GrantDetailsView.vue
jeffsmohan: Good call! Sorry, I just assumed it had been carried over without intention from the previous arrangement, my bad. 

Longer-term, I know the team is looking at supporting mobile designs on smaller screen sizes, so we might be able to address this sort of thing even more robustly...

usdr-gost.code-workspace
```diff
@@ -26,6 +26,6 @@
 		// competing user settings and workplace settings. We use VSCode's "codeActionsOnSave"
 		// to manage linting/fixing, but users can have other settings via "formatOnSave".
 		// If both are on, the editor can "fix" things that are immediately flagged as lint failures.
-		"editor.formatOnSave": false,
+		"editor.formatOnSave": true,
```
jeffsmohan, line 29: I don't think we want to check this change in. Were you not able to get automatic linting on save working without this change? 

I'd love to look into that a bit more with you... the `codeActionsOnSave` is supposed to supersede the `formatOnSave` setup in VSCode I believe, but the setup is finicky, and I didn't have many users to test out the settings with when I was changing them last...

jeffsmohan (changes requested):
This looks basically ready to go! 

Only thing is let's avoid changing the workspace settings, at least in this PR. I'd love to discuss what linting issues you were running into that this solved for you, and then maybe we can turn that into a separate PR to get linting set up more robustly for folks!

usdr-gost.code-workspace
amyewang: whoops sorry this was careless, I was playing with settings and accidentally committed this. my bad.

jeffsmohan (approved):
lgtm!

PR #3191, opened 2024-06-18
[Decision Discussion] Migrate to Vue 3

jeffsmohan:
## Description

This PR adds a decision doc for upgrading from Vue 2 to Vue 3, and comparing that against the alternatives. Please comment on this PR with your thoughts, concerns, and questions so we can discuss! 

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (changes requested):
I'm altogether in favor of moving to Vue 3, given the Vue 2 is pretty out-of-date and React would be a larger lift than I think is worth doing at the moment.

My only real reservation, which I'd like to see this (or a subsequent) proposal address would be a methodology for the team to keep track of outstanding usages of the compatibility layer and track progress against that over time. IMO, the strategy that we used for ESLint updates a while back was **really** nice, and while I understand that this isn't exactly apples-to-apples, I wonder if there are ways we could do something along the lines of:
1. Identify all places where we are not fully compliant with Vue 3.
2. Add all of those places to a "manifest" that would ignore/silence those warnings.
3. Adopt a requirement that all (non-emergency) PRs not introduce new warnings.
4. Create engineering tasks to migrate bit-size portions of the code to Vue 3 and un-ignore warnings for those portions in the same PR (i.e. so any new code relying on the compatibility layer would generate a warning).
5. Work our way up to the point that the manifest becomes empty.

Coming up with the exact methodology might be out of scope for this proposal, but I'd like to discuss options and/or include some details around how we'll handle fully migrating (as opposed to, say, moving to Vue 3 and then just living with compatibility layer warnings in perpetuity) before this proposal is fully accepted.

jeffsmohan:
@TylerHendrickson thanks for the feedback! I've addressed the deprecation warning handling (per your comment above) and the performance considerations (per discussion at our last team meeting), and I'd appreciate another look

TylerHendrickson (approved):

PR #3208, opened 2024-06-22
Assign-to-share: Update grant detail page

thucdtran:
## Description
This ticket implements the UI changes specified by the update grant detail issue.

## Screenshots / Demo Video
Screenshot 2024-06-22 at 4 12 01 PM


packages/client/src/views/GrantDetailsView.vue
```diff
@@ -378,6 +401,9 @@ export default {
     newTerminologyEnabled() {
       return newTerminologyEnabled();
     },
+    shareTerminologyEnabled() {
+      return shareTerminologyEnabled();
+    },
```
jeffsmohan, line 406: (very minor/optional/non-blocking) This can be simplified to `shareTerminologyEnabled,` which is a bit more terse and makes it exceedingly clear that we're not wrapping the helper method to change its functionality in any way.

jeffsmohan (changes requested):
This is looking great! Just a couple very minor suggested tweaks.

packages/client/src/views/GrantDetailsView.vue
```diff
@@ -134,30 +134,53 @@
...
-              class="d-flex justify-content-between align-items-start my-3"
-            >
-              <div class="mr-3">
-                <p class="m-0">
-                  {{ agency.name }}
-                </p>
-                <p class="m-0 text-muted">
-                  <small>{{ formatDateTime(agency.created_at) }}</small>
-                </p>
+            <div v-if="!shareTerminologyEnabled">
```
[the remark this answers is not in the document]
thucdtran: Done.

packages/client/src/views/GrantDetailsView.vue
thucdtran: Not quite sure, what you mean here; does this mean to change this to appear like:

```
   newTerminologyEnabled() {
      return newTerminologyEnabled();
    },
    shareTerminologyEnabled,
```

?

jeffsmohan: Ah, sorry, I think the suggestion became so terse that it lost its context and was a bit tough to understand. To be more clear, you can apply this diff and the behavior will be identical: 

```diff
diff --git a/packages/client/src/views/GrantDetailsView.vue b/packages/client/src/views/GrantDetailsView.vue
index afc16fbb..623178c9 100644
--- a/packages/client/src/views/GrantDetailsView.vue
+++ b/packages/client/src/views/GrantDetailsView.vue
@@ -404,12 +404,8 @@ export default {
         (interested) => interested.agency_id.toString() === this.selectedAgencyId,
       );
     },
-    newTerminologyEnabled() {
-      return newTerminologyEnabled();
-    },
-    shareTerminologyEnabled() {
-      return shareTerminologyEnabled();
-    },
+    newTerminologyEnabled,
+    shareTerminologyEnabled,
     unassignedAgencies() {
       return this.agencies.filter(
         (agency) => !this.assignedAgencies.map((assigned) => assigned.id).includes(agency.id),
```

Again, totally optional and non-blocking!

jeffsmohan (approved):
lgtm! 

I replied to your question about my comment, but like I said, that's totally optional, so this is good to go as-is if you like!

packages/client/src/views/GrantDetailsView.vue
thucdtran: Done!

jeffsmohan (approved):

Issue #3200, opened 2024-06-21
Follow+Notes: Add grant notes tables to database


TylerHendrickson:
Subtask of **[[STORY]: Update 'Status' to 'Follow + Note' feature](https://github.com/usdigitalresponse/usdr-gost/issues/2960)** #2960 

#### Blocked by

N/A

#### Blocks

- (Library)

### Definition of Done

A new Knex migration is added that creates the tables needed for storing user-submitted notes about grants.

### Implementation Details

This requires the following new tables to be created:
- Table `grant_notes`
  - Columns:
    - `id`: serial, not null, primary key
    - `grant_id`: text, not null, foreign key to `grants.grant_id`
    - `user_id`: int, not null, foreign key to `users.id`
    - `created_at`: timestamp with time zone, not null, default `now()`
  - Unique index `idx_grant_id_user_id` on columns `(grant_id, user_id)`
- Table `grant_notes_revisions`
  - Columns:
    - `id`: serial, not null, primary key
    - `grant_note_id`: int, not null, foreign key to `grant_notes.id`
    - `created_at`: timestamp with time zone, not null, default `now()`
    - `text`: text, not null

Issue #3201, opened 2024-06-21
Follow+Notes: Create backend library for notes management


TylerHendrickson:
Subtask of **[[STORY]: Update 'Status' to 'Follow + Note' feature](https://github.com/usdigitalresponse/usdr-gost/issues/2960)** #2960 

#### Blocked by

- #3200

#### Blocks

TBD

### Definition of Done

A new library, `packages/server/src/lib/grantNotes` exists and provides an interface for saving and retrieving notes created by users about particular grants.

### Implementation Details

- A user may only have one note per grant (which is enforced by a unique constraint on `(grant_notes.grant_id, grant_notes.user_id)`) but may have infinite revisions.
- Create 
- Create a new library module at `packages/server/src/lib/grantNotes` which exports the following interface:
  - `saveNoteRevision(knex, grantId, userId, text)`
    - The `knex` argument represents a `knex` connection object, e.g. as exported by `packages/server/src/db/index.js`.
    - Determines whether a row in the `grant_notes` table already exists for the given `grantId` and `userId`. If a row for this combination does not already exist, a new one must be inserted. In case of an insert, the `id` of the newly-inserted row must be retained. Otherwise, the `id` of the existing row must be retained.
    - Using the retained `grant_notes.id` value, insert a new row into the `grant_notes_revisions` table with the provided text.
    - Note that none of the provided arguments to this function may be `null` or `undefined` or else the database insert operation(s) will fail.
    - To ensure atomic writes, this function should create a database transaction that encapsulates all write operations. The transaction should be committed before the function returns.
    - This function should return an object that matches the following structure:
      ```javascript
      { id: grant_notes_revisions.id }
      ```
  - `getOrganizationNotesForGrant(knex, grantId, organizationId, { afterRevision, limit = 50 } = {})`
    - Queries for grant note revisions pertaining to the identified grant and organization.
    - Results should be ordered by `grant_notes_revisions.created_at` in descending order (most recent first).
    - If a non-null/-undefined value is provided for the `afterRevision` argument, the query should be limited to results where `grant_notes_revisions.id` is greater than that value.
    - This function should return an object that matches the following structure:
      ```javascript
      {
        notes: [{
          id: grant_notes_revisions.latest_revision_id,
          createdAt: grant_notes_revisions.created_at,
          text: grant_notes_revisions.text,
          grant: { id: grant_notes.grant_id },
          user: {
            id: users.id,
            name: users.name,
            team: {
              id: users.agency_id,
[...]

PR #3209, opened 2024-06-23
Issue #3200 Add grant_notes and grant_notes_revisions table migration

thucdtran:
## Description
Add a knex migration for issue #3200. 

## Screenshots / Demo Video
Screenshot 2024-06-22 at 10 12 36 PM

## Testing
Checked the table schemas in postgres CLI. 

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/migrations/20240622205204_add-grant-notes-table.js
+++ b/packages/server/migrations/20240622205204_add-grant-notes-table.js
@@ -13,7 +13,10 @@ exports.up = function (knex) {
             table.foreign('grant_id').references('grants.grant_id');
             table.foreign('user_id').references('users.id');
 
-            table.unique(['grant_id', 'user_id'], 'idx_grant_id_user_id');
+            table.unique(['grant_id', 'user_id'], {
+                indexName: 'idx_grant_id_user_id',
+                useConstraint: false,
+              });
         })
         .createTable('grant_notes_revisions', (table) => {
             table.increments('id').notNullable().primary();
--- a/packages/server/migrations/20240622205204_add-grant-notes-table.js
+++ b/packages/server/migrations/20240622205204_add-grant-notes-table.js
@@ -1,0 +1,34 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.schema
+        .createTable('grant_notes', (table) => {
+            table.increments('id').notNullable().primary();
+            table.text('grant_id').notNullable();
+            table.integer('user_id').notNullable();
+            table.timestamp('created_at', { useTz: true }).notNullable().defaultTo(knex.fn.now());
+
+            table.foreign('grant_id').references('grants.grant_id');
+            table.foreign('user_id').references('users.id');
+
+            table.index(['grant_id', 'user_id'], 'idx_grant_id_user_id');
+        })
+        .createTable('grant_notes_revisions', (table) => {
+            table.increments('id').notNullable().primary();
+            table.integer('grant_note_id').notNullable();
+            table.timestamp('created_at', { useTz: true }).notNullable().defaultTo(knex.fn.now());
+            table.text('text').notNullable();
+            table.foreign('grant_note_id').references('grant_notes.id');
+        });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+    return knex.schema.dropTable('grant_notes')
+        .dropTable('grant_notes_revisions');
+};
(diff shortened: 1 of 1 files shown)


jeffsmohan:
This PR looks correct to the spec, but before we land the change I wanted to better understand why this is being structured as two tables instead of one. Are there query patterns or other considerations that make this setup better than a single table? 

To make my question more concrete, here's my first instinct of how I would structure the new notes data, if we want to capture all the revisions: 

- table `grant_notes`
  - `id`: serial, not null, primary key
  - `grant_id`: text, not null, foreign key to `grants.grant_id`
  - `user_id`: int, not null, foreign key to `users.id`
  - `text`: text, not null
  - `created_at`: timestamp with time zone, not null, default now()

It seems like with either setup (as spec'd in the issue, or the table I laid out here) you have to do a slightly complicated query to get what you want to display on the page: (note that I haven't tested this query... it probably could use some tweaking, but it seemed useful for illustration)

```sql
with current_grant_notes as (
  select 
    id, 
    user_id, 
    text, 
    row_number() over(partition by user_id order by created_at desc) as rank
  from grant_notes
  where grant_id = {grant_id}
  -- also probably need to join in agency to filter this down to within-agency notes
)
select * from current_grant_notes
where rank = 1
```

If we wanted that query pattern to be simpler, I think we'd need to denormalize the concept of "current note" in the database (either as a separate table or as a flag on the table I'm proposing), which would then require a bit more upkeep when writing to the table. If we don't mind the above query pattern, then I'm not yet seeing the benefit of the two tables vs the one. 

Sorry to hijack the PR for a design discussion, but I wanted to make sure we've nailed down the schema before committing to it. (And my apologies if this was already discussed/captured elsewhere and I missed it!) 

What do you think, @TylerHendrickson and @thucdtran ?

TylerHendrickson (approved):

TylerHendrickson:
> This PR looks correct to the spec, but before we land the change I wanted to better understand why this is being structured as two tables instead of one. Are there query patterns or other considerations that make this setup better than a single table?
>  ...
> What do you think, @TylerHendrickson and @thucdtran ?

@jeffsmohan Thanks for the thoughts here – I wrestled with this a bit also. A few things to mention:
- I did document a query pattern for getting the most recent revision for each grant/user here: https://github.com/usdigitalresponse/usdr-gost/issues/3201. It uses a `LEFT JOIN LATERAL` (which maybe isn't a lot better) instead of a CTE, but just wanted to provide a link to the proposed querying pattern.
- The challenge with using a single-table design has to do with enforcing the "one 1 note per-user-per-grant" constraint, and particularly doing that in a way that is compatible with us deciding to do away with this constraint at some point (which I think is pretty likely). With a single table, we would need:
  - Not implement a unique index/constraint on `(grant_id, user_id)`
  -  Add another column like `common_note_id` (self-referential FK to `id`)
  - Enforce that every row sharing the same `common_note_id` also shares the same combination of `(grant_id, user_id)` values. The only way I know of that could do this would be a `CHECK` constraint, and given that we don't know much about usage patterns of this feature just yet, I'm not confident that the associated query would be very optimal.
- While there's no explicit requirement to soft-delete comments at the moment, a two-table design allows us to do things like add an `is_deleted` boolean column (and other note metadata that's common to all revisions) to the `grant_notes` table; the alternative would either require adding `is_deleted` to all revision rows of a single note or else some other design choice that would be hard to anticipate before knowing more about the feature.

I think it's entirely possible that as the feature evolves and we understand more about its usage, we'll either need to migrate the data in this table to a new design or even to a different storage backend entirely, like DynamoDB or a document store, which are much more suitable for storing revisions and eventually consistent behaviors that I associate with distributed commenting systems. So I'm really viewing this table design as "works for the known short-term, somewhat robust enough for fast-follow/medium-term changes, with a decent chance that it will eventually be replaced in its entirety", while remaining something that's unlikely to have performance issues as a forcing function in making those decisions.

jeffsmohan:
@TylerHendrickson gotcha, thanks for catching me up, and sorry I missed the background in the other ticket! 

I see where the design is coming from, and I agree we should be prepared to evolve the schema as these early feature requirements evolve anyway, so it's not worth holding this up any further. (In other words, PR looks good to merge!)

I'll just call out that it seems the requirement to store all note revisions ends up making the schema and the query patterns significantly more complicated to write (and whether it ends up mattering or not, less efficient queries). I'm trying to remember, was the storage of past revisions just to satisfy our internal desire to understand if/when/how users are modifying their notes? If so, I wonder if there's a better path where we only store the current note per user in the DB, but spew all revisions out to datadog as logs or some other less structured storage solution that we can query ad-hoc. Just noodling... if it spurs something useful, great; if not, consider me more than sufficiently happy to continue down the current path!

PR #3345, opened 2024-08-05
feat(grants-collaboration): API endpoints for grant notes creation

amyewang:
## Description
Created endpoint and unit tests for the endpoint. 
@TylerHendrickson I was slightly confused by `Requires that the organization (tenant) of the authenticated user matches the :organizationId request path parameter.` - I added a path param `:organizationId` to compare to `user.tenant_id`. I also grabbed the agency based on the organizationId passed in to check for authorization. I'm not totally sure if this is the expected pattern. 

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/src/routes/grants.js
```diff
@@ -422,4 +424,41 @@ router.delete('/:grantId/interested/:agencyId', requireUser, async (req, res) =>
...
+    const { user } = req.session;
+    const reqTenant = await db.getTenant(organizationId);
+    const reqMainAgencyId = reqTenant.main_agency_id;
+    const allowed = await isUserAuthorized(user, reqMainAgencyId);
+    const userTenant = await db.getTenant(user.tenant_id);
+
+    if (userTenant.id !== parseInt(organizationId, 10) || !allowed) {
+        res.sendStatus(403);
+        return;
+    }
```
TylerHendrickson, line 438: Re
> @TylerHendrickson I was slightly confused by `Requires that the organization (tenant) of the authenticated user matches the :organizationId request path parameter.` - I added a path param `:organizationId` to compare to `user.tenant_id`. I also grabbed the agency based on the organizationId passed in to check for authorization. I'm not totally sure if this is the expected pattern. 

There is actually already an `organizationId` path parameter – it's [included in the `configureApiRoutes()` function](https://github.com/usdigitalresponse/usdr-gost/blob/09ac319ee6fd67fa5489debede6a533fcbcdda55/packages/server/src/configure.js#L19) in `packages/server/src/configure.js` as part of the base path for all routes in this module.

Additionally, the `requireUser` middleware already performs the necessary access checks and aborts the request early with a 403 status code if things don't match up, so this can be simplified a bit:

```suggestion
router.put('/:grantId/notes/revision', requireUser, async (req, res) => {
    const { grantId } = req.params;
    const { user } = req.session;

```

TylerHendrickson (changes requested):

packages/server/src/routes/grants.js
amyewang: ahh I see the middleware now. thanks!

TylerHendrickson (approved):
@amyewang Looks great!

PR #3408, opened 2024-08-16
Bug fix for disabled button

greg-adams:
## Description
Bug fix for unwanted disabled submit button in Team Status section. The template syntax for this boolean is correct, however, the issue is caused by seed data which uses an id of `0`, which is falsy in JS.  I'd suggest we should always avoid using number 0 for a record id, and it appears the default in Postgres is to start a sequence with 1. 

## Screenshots / Demo Video
Screenshot from 2024-08-16 10-46-37

## Testing
Manual testing notes:
- Bug was occurring specifically after selecting "Not applicable to needs/goals" option
- This fix will require re-seeding the db.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,7 +154,7 @@ jobs:
           BAKEFILE_PATH: ${{ steps.meta.outputs.bake-file }}
       - name: Build and push Docker image
         id: build-push
-        uses: docker/build-push-action@16ebe778df0e7752d2cfcbd924afdbbd89c1a755 # v6.6.1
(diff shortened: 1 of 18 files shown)


TylerHendrickson (changes requested):
@greg-adams I like the direction here, but given that our Staging and Production environments have the ID of `0` I'm not sure this change will resolve the problem outside of dev environments with no existing data.

If we wanted to go this route in Staging / Production, I think we would need to migrate existing records in the DB, as well as any FK references to the current IDs.

greg-adams:
updated to resolve issue without data changes

TylerHendrickson (approved):

PR #3147, opened 2024-06-07
Adding share feature flag

mayabose:
## Description
This request adds the share terminology feature flag which will eventually control whether assign terms are replaced with share terms. The flag is currently enabled for staging and dev environments and disabled for prod environments.

## Testing
The feature flag value is reflected correctly in the dev console for the grant finder website, both when set to true and false.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [ ] Added steps to test feature/functionality manually

## Checklist
- [ ] Provided ticket and description
- [ ] Provided testing information
- [ ] Provided adequate test coverage for all new code
- [ ] Added PR reviewers

Issue #3060, opened 2024-05-15
Assign-to-Share: Add feature flags


TylerHendrickson:
Subtask of **[[STORY]: Change "assign" feature to "share"](https://github.com/usdigitalresponse/usdr-gost/issues/2823)** #2823 

#### Blocked by

N/A

#### Blocks

TBD

### Definition of Done

A server-side environment variable and corresponding client-side feature flag exist to control whether changes related to #2823 are activated in a particular environment.

The goal is to enable incremental integration of updates to the Staging environment for testing, while delaying user-facing changes in Production until all changes are ready.

### Implementation Details:

- In `terraform/staging.tfvars`, add `SHARE_TERMINOLOGY_ENABLED = true` to the `api_container_environment` map of environment variables.
- In `packages/server/.env.example`, add `SHARE_TERMINOLOGY_ENABLED = true` to the default environment variables for dev environments.
- Read over the [Feature Flags documentation](https://github.com/usdigitalresponse/usdr-gost/blob/main/docs/feature-flags.md) in `docs/feature-flags.md` to learn about how we implement and manage (client-side) feature flags.
- Define a client-side feature flag as `window.APP_CONFIG.shareTerminologyEnabled`. The per-environment values for this flag should be:
  - Dev environments: enabled
  - Staging environment: enabled (note: managed by Terraform)
  - Production environment: disabled (note: managed by Terraform)
- Provide a [helper function](https://github.com/usdigitalresponse/usdr-gost/blob/9f989d4851a03544dab6bc2b0921884d1df734d2/packages/client/src/helpers/featureFlags/index.js) to access the state of the feature flag.

PR #3456, opened 2024-09-02
Hide forecasted grants while UI is being developed to support forecasted grants

masimons:
## Description
Hide forecasted grants on production during development and testing.

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


TylerHendrickson (commented):
@masimons Confirming that the direction this is headed makes sense to me 👍 

I'll mention that I think it could be useful to have an environment variable control the exclusion of forecasted grants. Due to how we ship and run Docker images, it's not something that we could toggle in a Staging/Prod environment without re-deploying, but 1) it might be helpful for development purposes, and 2) might make it easier to activate once we get to the point where we want to start including forecasted grants in one or both of those environments.

masimons:
@TylerHendrickson thanks for checking this out, I missed this comment. Out of curiosity, do you have an example off the top of your head where we use an env var in that way? I was looking into how we use feature flags, but it seems that those are scoped to the client side?

TylerHendrickson:
@masimons re
> Out of curiosity, do you have an example off the top of your head where we use an env var in that way? I was looking into how we use feature flags, but it seems that those are scoped to the client side?

The implementation details for #3060 provide some instructions for implementing an API-side feature flag using a `SHARE_TERMINOLOGY_ENABLED` environment variable (plus a client-side feature flag, which doesn't apply here). This was implemented by PR #3147. 

Fundamentally, this sort of thing would entail the following:
- An env var default value for dev environments, defined in `packages/server/.env.example`.
- Configuring the value for Staging environments as a key/value pair in the [`api_container_environment` input variable](https://github.com/usdigitalresponse/usdr-gost/blob/d27bbe50cefea8e23f5ef8676d9d75653db75242/terraform/staging.tfvars#L74-L77) in `terraform/staging.tfvars`.
- Configuring the value for Production environments via the [same input variable](https://github.com/usdigitalresponse/usdr-gost/blob/d27bbe50cefea8e23f5ef8676d9d75653db75242/terraform/prod.tfvars#L75-L78) in `terraform/prod.tfvars`.

I'd recommend that an environment variable that provides a binary feature flag be evaluated as a string using strict equality, e.g. `my_feature_is_enabled = process.env.MY_FEATURE_ENABLED === 'true'`.

TylerHendrickson (changes requested):
@masimons I really like how this implementation parameterizes the flag in `src/db/index.js` and allows the caller to decide whether to pull in the toggle from env var or something else. This is looking good overall!

I did notice in local testing that setting `SHOW_FORECASTED_GRANTS=false` in `packages/server/.env` doesn't seem to have the desired effect (I could still see forecasted grants in search results), although when the env var is altogether missing, forecasted grants are hidden like we'd expect. 

I believe this is because the `packages/server/db/index.js` functions are expecting a true boolean value (or else `undefined`, so that the arg default of `false` in the function signature is used), but when a value is provided from `process.env.SHOW_FORECASTED_GRANTS`, it's a string rather than a boolean, and `!"false"` is evaluating to `true`. I think the simplest way to resolve this is to have callers explicitly check whether `process.env.SHOW_FORECASTED_GRANTS === "true"`, which I provided in a comment suggestion.

lsr-explore:
> @masimons I really like how this implementation parameterizes the flag in `src/db/index.js` and allows the caller to decide whether to pull in the toggle from env var or something else. This is looking good overall!
> 
> I did notice in local testing that setting `SHOW_FORECASTED_GRANTS=false` in `packages/server/.env` doesn't seem to have the desired effect (I could still see forecasted grants in search results), although when the env var is altogether missing, forecasted grants are hidden like we'd expect.
> 
> I believe this is because the `packages/server/db/index.js` functions are expecting a true boolean value (or else `undefined`, so that the arg default of `false` in the function signature is used), but when a value is provided from `process.env.SHOW_FORECASTED_GRANTS`, it's a string rather than a boolean, and `!"false"` is evaluating to `true`. I think the simplest way to resolve this is to have callers explicitly check whether `process.env.SHOW_FORECASTED_GRANTS === "true"`, which I provided in a comment suggestion.

@TylerHendrickson , @masimons - I've accepted Ty's changes - this should be ready for review

packages/server/src/lib/email.js
```diff
@@ -423,7 +423,7 @@ async function sendGrantAssignedEmails({ grantId, agencyIds, userId }) {
             i. Send email
     */
     try {
-        const grant = await db.getGrant({ grantId });
+        const grant = await db.getGrant({ grantId, showForecastedGrants: process.env.SHOW_FORECASTED_GRANTS });
```
TylerHendrickson, line 426:
```suggestion
        const grant = await db.getGrant({ grantId, showForecastedGrants: process.env.SHOW_FORECASTED_GRANTS === 'true' });
```

packages/server/src/routes/grants.js
```diff
@@ -42,6 +42,7 @@ router.get('/', requireUser, async (req, res) => {
         },
         orderBy: req.query.orderBy,
         orderDesc: req.query.orderDesc,
+        showForecastedGrants: process.env.SHOW_FORECASTED_GRANTS,
```
TylerHendrickson, line 45:
```suggestion
        showForecastedGrants: process.env.SHOW_FORECASTED_GRANTS === 'true',
```

packages/server/src/routes/grants.js
```diff
@@ -92,6 +93,7 @@ router.get('/next', requireUser, async (req, res) => {
         user.tenant_id,
         user.agency_id,
         false,
+        process.env.SHOW_FORECASTED_GRANTS,
```
TylerHendrickson, line 96:
```suggestion
        process.env.SHOW_FORECASTED_GRANTS === 'true',
```

packages/server/src/routes/grants.js
```diff
@@ -128,6 +130,7 @@ router.get('/exportCSVNew', requireUser, async (req, res) => {
         user.tenant_id,
         user.agency_id,
         true,
+        process.env.SHOW_FORECASTED_GRANTS,
```
TylerHendrickson, line 133:
```suggestion
        process.env.SHOW_FORECASTED_GRANTS === 'true',
```

TylerHendrickson (changes requested):
@lsr-explore Thanks for the updates. I think we need the strict `=== 'true'` comparision for all usages of the environment variable, so I added a few additional comment suggestions to address those as well.

lsr-explore:
> @lsr-explore Thanks for the updates. I think we need the strict `=== 'true'` comparison for all usages of the environment variable, so I added a few additional comment suggestions to address those as well. I _think_ that I covered all of them, but feel free to double check me on that 🙂

Sorry that I missed this @TylerHendrickson .  I applied the changes and did another sanity check of the code.  I noticed a couple additional places where the flag needed to be sent and added a commit.

I noticed that the db.getGrantsInterested didn't have this check.   This would only be an issue if we needed to turn off grants after enabling it and user had clicked follow.   

Do you think we need the check on getGrantsInterested?

masimons:
> 
> I noticed that the db.getGrantsInterested didn't have this check. This would only be an issue if we needed to turn off grants after enabling it and user had clicked follow.
> 
> Do you think we need the check on getGrantsInterested?


That's a fair point - I don't know if that particular case came up when we were thinking about this

PR #3440, opened 2024-08-24
feat: Add client-side feature flag for Follow+Notes functionality

sushilrajeeva:
### Ticket #3405 
Fixes #3405 

## Description

This PR introduces the followNotesEnabled feature flag in the client-side library, enabling conditional display of new Follow+Notes features. The flag is registered as true by default in development and false in staging and production environments. Lines 16-17 in deploy-config.js have been reformatted for ESLint compliance.
### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer
- [ ] Added steps to test feature/functionality manually

## Checklist
- [x] Provided ticket and description
- [ ] Provided screenshots/demo
- [ ] Provided testing information
- [ ] Provided adequate test coverage for all new code
- [x] Added PR reviewers

Issue #3428, opened 2024-08-22
Follow+Notes: Implement note input box on Grant Details view


TylerHendrickson:
Subtask of **[[STORY]: Update 'Status' to 'Follow + Note' feature](https://github.com/usdigitalresponse/usdr-gost/issues/2960)** #2960 

#### Blocked by

- #3204
- #3205
- #3406

#### Blocks

N/A

### Definition of Done

[See Figma Designs](https://www.figma.com/design/L2gfVMBH0x1daOlOe50sJl/Follow-%2F-Leave-a-Note?node-id=1304-6485)

- Develop a new reusable `GrantNote` component, which displays as the following:
  - User's avatar, name, team name (truncated to fit) on the first line
  - User's email on the second line, as a hyperlink-style "Copy Button" component which copies the email address to the clipboard when clicked
  - Date of the note (this will show "Today", "[#] days ago" up to seven days, and more than seven days will show the date followed in the format "[Month] [Date]", i.e. May 4)
  - "Edit" link, which conditionally displays on the same line as the date (e.g. toggled via a boolean `editable` slot)
- On the [Grant Details view](https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/client/src/views/GrantDetailsView.vue), update the Grant Activity card to include the following:
  - Note Input, which displays as the following
    -  The current user's icon.
    - A text entry box with the following example text 'Leave a note with tips or barriers to applying...' followed by a send icon.
    - Below the text entry box, the following description text: 'e.g. need co-applicants, we applied last year'
    - To the right of the description text, a character count will display - when nothing has been entered, it will show "0/300" - as the user types in the note box, the count will reflect the current characters (i.e. "50/300"). When the character count hits 300, the counter will show in red text as "300/300", and the user will not be able to enter any additional characters.
    - As the user types, the note box will expand to show the full note
    - Clicking return will also submit the note. Upon successful submission, the Note Input is replaced with a corresponding Pinned Note
    - Displays in the following cases:
      - When the current user has not previously submitted a note for the grant being viewed
      - When the current user clicks the "Edit" button on the Pinned Note, in which case the Pinned Note is replaced with Note Input, and the Note Input text entry box is prefilled with the text of the Pinned Note (see below) 
  - Pinned Note
    - An occurrence of the `GrantNote` component, which displays an "Edit" link.
    - Displays only when the current user has previously submitted a note for the grant being viewed
[...]

Issue #3407, opened 2024-08-16
Follow+Notes: Implement Grant Followers modal component


TylerHendrickson:
Subtask of **[[STORY]: Update 'Status' to 'Follow + Note' feature](https://github.com/usdigitalresponse/usdr-gost/issues/2960)** #2960 

#### Blocked by

- #3406

#### Blocks

N/A

### Definition of Done

[Modal Design Link](https://www.figma.com/design/L2gfVMBH0x1daOlOe50sJl/Follow-%2F-Leave-a-Note?node-id=1101-14493&t=g60L7oc1fDDTaruj-4)

- Create a new "Grant Followers" modal component that provides the following:
   - A header of 'Grant Followers'
   - A list of all followers sorted alphabetically in ascending order by name.
   - User Avatar, User Name, User Team Name, User Email Address, date followed (this will show "Today", "[#] days ago" up to seven days, and more than seven days will show the date followed in the format "[Month] [Date]", i.e. May 4)
   - To the right of each follower's name will be a "Copy Email" button. Clicking the button will copy the email address for the individual user to clipboard
   - At the bottom of the modal there will be a button that allows users to copy all emails to their clipboard. (This should copy as a comma separated list of emails.)
   - When the user scrolls to the bottom of the modal list, a button should allow the ability to fetch the next page of followers for the grant. 
     - If the request for the next page returns no results, the button should be hidden until the modal is relaunched.
- Update the "Grant Activity" component so that the follower information text (`"Followed by {{Follower Name}} and {{Count Followers - 1}} others"`) is hyperlinked so that when clicked, the new "Grant Followers" modal displays.

### Implementation details

- The `GET /:grantId/followers` route implemented in #3399 supports requests to fetch (paginated) followers of a grant within the same organization as the currently-authenticated user. This endpoint should provide all data necessary to populate this modal (name, email, team name of following users, plus timestamp of the follow).
- For "Copy Email" and "Copy all Emails" buttons, use the [`CopyButton` component](https://github.com/usdigitalresponse/usdr-gost/blob/860ea950038ca0e0fe2f444576a6d68de71c33f9/packages/client/src/components/CopyButton.vue) defined in `packages/client/src/components/CopyButton.vue`.
  - Ideally, this component should be updated so that its core functionality may be reused while also being styled as a button (either filled or unfilled) in a manner consistent with the Figma designs linked above.
[...]

Issue #3406, opened 2024-08-16
Follow+Notes: Implement Grant Activity card component in Grant Details view


TylerHendrickson:
Subtask of **[[STORY]: Update 'Status' to 'Follow + Note' feature](https://github.com/usdigitalresponse/usdr-gost/issues/2960)** #2960 

#### Blocked by

N/A

#### Blocks

TBD

### Definition of Done

The "Share Grant" component in the Grant Details view is replaced by a new "Grant Activity" component that provides the following:
- 'Grant Activity' header.
- A section description of:
   - Stay up to date with this grant.
   - to the right of the text, an info icon (circle with an "i") will display - on hover, the following text will display within a tool tip: "Follow this grant to receive an email notification when others follow or leave a note." Use the [tool tip component from the design library here](https://www.figma.com/design/HWu4iIjcoX2txuN543qzIB/USDR-Design-System?node-id=166-2841)    
 - 'Follow' button
    - The button will have two states, primary and secondary, and appear differently based on the current state.
        - If the User has not yet followed, they will see the primary button state of a blue button with white text.
        - If Primary, button text will be a white circle with a checkmark and the text "Follow".
        - If the User has followed, they will see the secondary button state with a green button with white text.
        - If Secondary, button text will be a white circle with a green checkmark and "Following".
        - The User will be able to click the Secondary version of the button to un-follow the grant. This will return the button to the Primary state.
 - Grant Activity Summary
    - This will display below the Follow button and include both Follower information AND a count of notes. (If both have values, they will be separated by a bullet point.)
    - The Follower information will display the following:
       - "Followed by {{Follower Name}} and {{Count Followers - 1}} others" _(Note: If there are notes related to the grant, this will be followed by a bullet point and count of notes. If there are no followers we will suppress this link fully.)_
       - All of the text will also be a link that will open a modal the Grant Followers modal that allows you to see a full list of followers for the grant.

### Implementation details

TBD

ClaireValdivia:
@greg-adams seeing a PR for this issue in staging and just want to confirm if I should be expecting that PR to address all aspects of this issue? a couple quick notes of what I'm seeing:

- the card itself looks great! button, help text looking as expected. 
- However, I get an error whenever I go to any grant details page, and additional errors when I try clicking the follow button
- are you planning to do a separate PR for the follow button to work as expected and for activities to display?

greg-adams:
[...]

PR #3470, opened 2024-09-05
feat(notes+follow) - grant activity card

greg-adams:
## Description
- Adds `GrantActivity` component card
- Migrates Share Grant card to separate component

## Screenshots / Demo Video
Screenshot from 2024-09-05 09-59-13

## Testing
- Dependent on https://github.com/usdigitalresponse/usdr-gost/pull/3440 for testing in staging

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/client/src/views/GrantActivity.vue
```diff
@@ -0,0 +1,167 @@
...
+
+<script>
+import { mapActions, mapGetters } from 'vuex';
+
+export default {
+  components: {},
+  data() {
+    return {
+      userIsFollowing: false,
+      loadingFollowState: false,
```
TylerHendrickson, line 67: I'm not sure about the best way to resolve (perhaps initializing as `loadingFollowingState: true`, but open to suggestions), but I noticed that when testing in my local dev environment, the button initially defaults to the blue "Follow" state before the underlying API request completes. This makes me wonder if we can initialize to a loading/unknown state instead.

Testing flow that reveals what I noticed:
1. Given: Grant 1234 is not being followed by the current user
2. Go to the details page for Grant 1234. A blue "Follow" button displays.
3. Click "Follow". The button will remain blue and go into a disabled (greyed-out) state (as expected).
4. Once the underlying PUT request completes successfully, the button will turn green, say "Following", and return to enabled state.
5. Reload (or leave and return to) the details page for Grant 1234. Initially, the button shows as blue, enabled, and says "Follow" before eventually changing to green and "Following". 
    - This is easier to observe for testing by adding `await new Promise((cb) => setTimeout(cb, 2000));` at the start of the [GET `/:grantId/follow` request handler](https://github.com/usdigitalresponse/usdr-gost/blob/20f5dcad6390856a119e568a750913152cd8d836/packages/server/src/routes/grants.js#L508-L520) in `packages/server/src/routes/grants.js`.

packages/client/src/views/GrantActivity.vue
```diff
@@ -0,0 +1,167 @@
...
+      currentGrant: 'grants/currentGrant',
+    }),
+    followBtnLabel() {
+      return this.userIsFollowing ? 'Following' : 'Follow';
+    },
+    followBtnVariant() {
+      return this.userIsFollowing ? 'success' : 'primary';
+    },
+    followSummaryText() {
+      const userIsFollower = this.followers.some((follower) => follower.user.id === this.loggedInUser?.id);
```
TylerHendrickson, line 84: I wonder if this should instead be based on `this.userIsFollowing` – currently, the `userIsFollower` constant will be `false` if the current user is (for example) the first of 52 total followers, or any other scenario where the current user is a follower but isn't represented within the first page of the result set requested by `fetchFollowersState()`.

packages/client/src/views/GrantActivity.vue
```diff
@@ -0,0 +1,167 @@
...
+        this.followers = result.followers;
+      }
+    },
+    async fetchNotes() {
+      const result = await this.getNotesForGrant({ grantId: this.currentGrant.grant_id });
+
+      if (result) {
+        this.notes = result.notes;
+      }
+    },
```
TylerHendrickson, line 147: These should both be fine for the purposes of #3406, but just in case it's relevant, I want to note that `fetchNotes()` will eventually (per #3428) need to deal with pagination. 

Same thing potentially goes for `fetchFollowersState()` as well, depending on how #3407 is accomplished.

packages/client/src/views/GrantActivity.vue
TylerHendrickson, line 1: I could be wrong about this, but should this file be located in `packages/client/src/components`? I noticed that it's being treated as a component in `GrantDetailsView.vue`, but I'm not sure whether there are other considerations in Vue as to what constitutes a view vs component.

packages/client/src/views/ShareGrant.vue
TylerHendrickson, line 1: Same question regarding whether this should be moved to the `components` directory.

TylerHendrickson (changes requested):
@greg-adams Sending back to address a few things and run a couple of questions past you, but overall this is looking good!

packages/client/src/views/GrantActivity.vue
greg-adams: good catch, I've updated to be disabled until verified

packages/client/src/views/GrantActivity.vue
greg-adams: yes this is better - updated

packages/client/src/views/GrantActivity.vue
greg-adams: Not sure how next steps will be implemented just yet, but added support for query params

packages/client/src/views/GrantActivity.vue
greg-adams: I don't think Vue differentiates, I believe a component is a component. I think of a `components` folder for more generic, reusable ones, which doesn't seem to be strictly the case here, so I've moved these two in there as they are more widget-like

packages/client/src/helpers/fetchApi.js
```diff
@@ -11,6 +11,23 @@ function getDefaultHeaders() {
...
+    if (value !== undefined) {
+      pairs.push(
+        `${encodeURIComponent(key)}=${encodeURIComponent(value)}`,
+      );
+    }
+  });
+
+  return pairs.length ? `?${pairs.join('&')}` : '';
+};
+
```
TylerHendrickson, line 30: I think it might be best to use native way of doing this in JS:
```js
export const serializeQuery = (params = {}) => `?${new URLSearchParams(params).toString()}`;
```
Note that the above will return a string of `'?'` when `params` is empty; it should be perfectly fine to have a dangling `?` for this use-case, but could be modified to match the behavior that you currently have implemented:
```js
export const serializeQuery = (params = {}) => {
  const searchParams = new URLSearchParams(params);
  return searchParams.size ? `?${searchParams.toString()}` : '';
}
```

packages/client/src/store/modules/grants.js
```diff
@@ -212,6 +213,43 @@ export default {
...
+      }
+    },
+    async getFollowersForGrant({ rootGetters }, { grantId, limit, paginateFrom }) {
+      const queryParams = serializeQuery({ limit, paginateFrom });
+      try {
+        return await fetchApi.get(`/api/organizations/${rootGetters['users/selectedAgencyId']}/grants/${grantId}/followers${queryParams}`);
+      } catch (e) {
+        return null;
+      }
+    },
```
TylerHendrickson, line 230: Feel free to disregard if you still think the helper function is convenient and/or you want to avoid dangling `?` characters in URLs, but per my previous comment, this could also be implemented to use `URLSearchParams` directly, e.g.
```js
    async getFollowersForGrant({ rootGetters }, { grantId, limit, paginateFrom }) {
      const queryParams = new URLSearchParams({ limit, paginateFrom });
      try {
        return await fetchApi.get(`/api/organizations/${rootGetters['users/selectedAgencyId']}/grants/${grantId}/followers?${queryParams}`);
      } catch (e) {
        return null;
      }
    },
```

packages/client/src/store/modules/grants.js
```diff
@@ -212,6 +213,43 @@ export default {
       fetchApi.get(`/api/organizations/${rootGetters['users/selectedAgencyId']}/interested-codes`)
         .then((data) => commit('SET_INTERESTED_CODES', data));
     },
+    async getFollowerForGrant({ rootGetters }, { grantId }) {
+      try {
+        return await fetchApi.get(`/api/organizations/${rootGetters['users/selectedAgencyId']}/grants/${grantId}/follow`);
+      } catch (e) {
+        return null;
+      }
```
TylerHendrickson, line 221: I'm a bit wary of the error-suppression here; would you mind you elaborating a bit?

Edit: Presumably this is because non-`2xx` status codes from these endpoints are not necessarily indicative of an application error, correct? If so, what would you think of doing something like this in the `fetchApi` methods?
```js
export function get(url) {
  const options = {
    credentials: 'include',
    headers: getDefaultHeaders(),
  };
  return fetch(apiURL(url), options).then((r) => {
    if (r.ok) {
      return r.json();
    }
    const err = new Error(text || r.statusText);
    err.response = r;
    return r
      .text()
      .then(() => Promise.reject(err));
  });
}
```
This would (I think) preserve existing usages, but give functions like `getFollowersForGrant()` the ability to inspect responses for additional information, e.g.
```js
async getFollowerForGrant({ rootGetters }, { grantId }) {
  try {
    return await fetchApi.get(`/api/organizations/${rootGetters['users/selectedAgencyId']}/grants/${grantId}/follow`);
  } catch (e) {
    if (e.response.status === 404) {
      return null;
    }
    throw e;
  }
},
```

TylerHendrickson (changes requested):
@greg-adams Consider all of these open for discussion, but a couple additional things I'd like to get your take on.

packages/client/src/store/modules/grants.js
greg-adams: good question - yes, initially this was because a rejection is expected, and errors wouldn't make sense to be shown in the UI here.  I think its normal if these requests throw (and including the response helps), but maybe the question is how to handle errors that occur outside user actions - can we send to datadog, or other?

packages/client/src/components/GrantActivity.vue
```diff
@@ -0,0 +1,167 @@
...
+            class="mr-2"
+          />
+        </span>
+        <span class="h5">
+          {{ followBtnLabel }}
+        </span>
+      </b-button>
+      <div>
+        <span
+          v-if="showFollowSummary"
```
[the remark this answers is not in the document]
greg-adams: I've massaged the flow a little bit to account for this case

packages/client/src/helpers/fetchApi.js
greg-adams: updated to include native approach w/ minor adjustments to remove empty values

packages/client/src/store/modules/grants.js
TylerHendrickson: Datadog RUM is [capturing](https://app.datadoghq.com/rum/error-tracking) any unhandled exceptions and `console.error()` calls (along with [a few other error sources](https://docs.datadoghq.com/real_user_monitoring/browser/collecting_browser_errors/?tab=npm#error-sources)).

In general, I think we should ensure that any error that results from a user's action causes an alert/message being displayed to the user, along with collecting it in rum (if the app can display a message and call `datadogRum.addError()`, then no need to re-`throw`). I'm all for keeping a clean console, but only if it means we're not losing the opportunity to learn that an error exists – a `catch` block should either decide that the error is not an error at all, call `datadogRum.addError()`, or re-`throw` the error (as a last resort).

More specifically for this use-case, I think we should ensure the following:
- `2xx` and `404` status codes are not treated as errors
- Any other `4xx` or `5xx` status code results in a user-facing alert **and** an error should be collected in Datadog RUM.

greg-adams: Got it - thanks! Added use of datadog/alert for error handling

TylerHendrickson (approved):
@greg-adams Looks good!

ClaireValdivia:
@greg-adams this is looking great! just noting that when I'm testing in staging, I immediately get these errors when I go to a grant details page - I'm imagining this is expected at this point because we don't have followers hooked up yet, but just wanted to flag in case it's a surprise

PR #3481, opened 2024-09-09
Fix `terraform_map_duplicate_keys` tflint check error

TylerHendrickson:
## Description

This PR fixes an occurrence of a duplicate key in a Terraform map (see [`terraform_map_duplicate_keys` rule](https://github.com/terraform-linters/tflint-ruleset-terraform/blob/main/docs/rules/terraform_map_duplicate_keys.md)), which regularly surfaces as a `tflint` check error (see [example](https://github.com/usdigitalresponse/usdr-gost/pull/3480/files#annotation_26068508106)).


terraform/modules/gost_api/gateway.tf
```diff
@@ -69,23 +69,32 @@ module "api_gateway" {
...
+    caller          = "$context.identity.caller"
+    user            = "$context.identity.user"
     integration = {
       error         = "$context.integration.error"
       serviceStatus = "$context.integration.integrationStatus"
       latency       = "$context.integration.latency"
       requestId     = "$context.integration.requestId"
       status        = "$context.integration.status"
       errorMessage  = "$context.integrationErrorMessage"
-      latency       = "$context.integrationLatency"
```
TylerHendrickson, line 88: This is the line that was causing the `tflint` check failures, since `lantency` is also defined on [line 94 (previously line 84)](https://github.com/usdigitalresponse/usdr-gost/pull/3481/files#diff-aac5295bae76e49b720fd47277cb4d113f53e00e5f71ab9bc7ce5aaefbb8cdaeL84) above. According to API Gateway [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html), the `$context.integration.latency` and `$context.integrationLatency` variables are equivalent.

as1729 (approved):

PR #3503, opened 2024-09-13
feat: update logic to reflect July 2024 IAA and Beneficiary changes

as1729:
## Description

The PR ensures the following:

For Beneficiary subrecipient type.
  If the beneficiary was created prior to July 1st, 2024 - require either a UEI or TIN.
  If the beneficiary was created after July 1st, 2024 - require a TIN.

For IAA subrecipient type.
  Raise an error if there is no UEI or TIN letting the user know that we currently do not support it.

Note: There will soon be a PR that supports ability to import IAA subrecipients without a UEI or TIN.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer
- [ ] Added steps to test feature/functionality manually

## Checklist
- [x] Provided ticket and description
- [ ] Provided screenshots/demo
- [x] Provided testing information
- [x] Provided adequate test coverage for all new code
- [x] Added PR reviewers

Issue #3372, opened 2024-08-12
[Issue]: v20240823 update - UEI & TIN adjustments


ClaireValdivia:
### Why is this issue important?

In July we learned of changes to the field validations for UEI & TIN fields and adjustments are needed for upload validation to reflect these

### Current State

If entity type = subrecipient, UEI is required and TIN is optional; error will generate if UEI is blank
If entity type = contractor, either a UEI or TIN is required; error will generate if both are blank
If entity type = beneficiary, either a UEI or TIN is required; error will generate if both are blank

IAA is an entity type that did not exist before July 2024

### Expected State
Recognize v20240823 as the current version number, and the version warning should generate if the version number of the upload file is before or after the current version:
“Upload template version ([uploaded template version]) is [older/ newer] than the latest input template ([latest template version])”

Updates to UEI & TIN requirements:
Current state requirements are correct and should remain unchanged when entity type = contractor or subrecipient. However, the following changes should be made:

if IAA is selected as the entity type, UEI & TIN are both optional. No error message will generate if both are blank
If Beneficiary is selected as the entity type, TIN is required and UEI is optional. If TIN is blank, the following error message will generate: "You must enter a TIN for this subrecipient."

### Implementation Plan

The following functions need to be added...

ClaireValdivia:
@as1729 only for new beneficiaries

PR #3522, opened 2024-09-18
3372 - adds database changes to support managing IAA subrecipients

as1729:
## Description
This PR adds support for the new IAA subrecipient field. The goal here is to ensure that we can add subrecipients that do not have either a UEI or TIN value. In order to do this the PR does the following:

1. Adds a new `name` field to `arpa_subrecipients` table and a corresponding `unique index` that ensures that when both `uei` and `tin` are missing, that we do not create duplicate subrecipients with the same `name`.
2. Adds application-layer validation when `createSubrecipient` function is called.
3. Removes `unique` constraint on `uei` and `tin` fields and replaces with a `unique` index that only applies when these fields are `NOT NULL`.

### State Before migrations
```
usdr_grants=# \d arpa_subrecipients
                                          Table "public.arpa_subrecipients"
      Column       |           Type           | Collation | Nullable |                    Default                     
-------------------+--------------------------+-----------+----------+------------------------------------------------
 id                | integer                  |           | not null | nextval('arpa_subrecipients_id_seq'::regclass)
 tenant_id         | integer                  |           | not null | 
 created_at        | timestamp with time zone |           | not null | CURRENT_TIMESTAMP
 updated_at        | timestamp with time zone |           |          | 
 updated_by        | integer                  |           |          | 
 uei               | text                     |           |          | 
 tin               | text                     |           |          | 
 record            | text                     |           |          | 
 upload_id         | uuid                     |           |          | 
 treasury_id       | text                     |           |          | 
 has_been_reported | boolean                  |           | not null | false
 archived_at       | timestamp with time zone |           |          | 
Indexes:
    "arpa_subrecipients_pkey" PRIMARY KEY, btree (id)
    "arpa_subrecipients_tenant_id_tin_unique" UNIQUE CONSTRAINT, btree (tenant_id, tin)
    "arpa_subrecipients_tenant_id_treasury_id_unique" UNIQUE CONSTRAINT, btree (tenant_id, treasury_id)
    "arpa_subrecipients_tenant_id_uei_unique" UNIQUE CONSTRAINT, btree (tenant_id, uei)
Foreign-key constraints:
    "arpa_subrecipients_tenant_id_foreign" FOREIGN KEY (tenant_id) REFERENCES tenants(id)
    "arpa_subrecipients_updated_by_foreign" FOREIGN KEY (updated_by) REFERENCES users(id)
    "arpa_subrecipients_upload_id_foreign" FOREIGN KEY (upload_id) REFERENCES uploads(id) ON DELETE CASCADE

```
### State After Migrations
```
fdcc06a80a69:/app/packages/server# yarn db:migrate
yarn run v1.22.19
$ knex migrate:latest
Using environment: development
Batch 2 run: 3 migrations
Done in 0.92s.
fdcc06a80a69:/app/packages/server# 

usdr_grants=# \d arpa_subrecipients
                                          Table "public.arpa_subrecipients"
      Column       |           Type           | Collation | Nullable |                    Default                     
-------------------+--------------------------+-----------+----------+------------------------------------------------
 id                | integer                  |           | not null | nextval('arpa_subrecipients_id_seq'::regclass)
 tenant_id         | integer                  |           | not null | 
 created_at        | timestamp with time zone |           | not null | CURRENT_TIMESTAMP
 updated_at        | timestamp with time zone |           |          | 
 updated_by        | integer                  |           |          | 
 uei               | text                     |           |          | 
 tin               | text                     |           |          | 
 record            | text                     |           |          | 
 upload_id         | uuid                     |           |          | 
 treasury_id       | text                     |           |          | 
 has_been_reported | boolean                  |           | not null | false
 archived_at       | timestamp with time zone |           |          | 
 name              | text                     |           |          | 
Indexes:
    "arpa_subrecipients_pkey" PRIMARY KEY, btree (id)
    "arpa_subrecipients_tenant_id_treasury_id_unique" UNIQUE CONSTRAINT, btree (tenant_id, treasury_id)
    "idx_arpa_subrecipients_tenant_id_name_unique" UNIQUE, btree (tenant_id, name) WHERE name IS NOT NULL AND uei IS NULL AND tin IS NULL
    "idx_arpa_subrecipients_tenant_id_tin_unique" UNIQUE, btree (tenant_id, tin) WHERE tin IS NOT NULL
    "idx_arpa_subrecipients_tenant_id_uei_unique" UNIQUE, btree (tenant_id, uei) WHERE uei IS NOT NULL
Check constraints:
    "chk_at_least_one_of_uei_tin_name_not_null" CHECK (num_nonnulls(uei, tin, name) > 0)
Foreign-key constraints:
    "arpa_subrecipients_tenant_id_foreign" FOREIGN KEY (tenant_id) REFERENCES tenants(id)
    "arpa_subrecipients_updated_by_foreign" FOREIGN KEY (updated_by) REFERENCES users(id)
    "arpa_subrecipients_upload_id_foreign" FOREIGN KEY (upload_id) REFERENCES uploads(id) ON DELETE CASCADE
```

### Confirming that Rollback works
```
fdcc06a80a69:/app/packages/server# yarn db:rollback
yarn run v1.22.19
$ knex migrate:rollback
Using environment: development
Batch 2 rolled back: 3 migrations
Done in 0.87s.
fdcc06a80a69:/app/packages/server# 

```

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers

### DIFF
--- a/packages/server/__tests__/arpa_reporter/server/helpers/with-tenant-id.js
+++ b/packages/server/__tests__/arpa_reporter/server/helpers/with-tenant-id.js
@@ -8,12 +8,12 @@ const { requestProviderMiddleware } = require('../../../../src/arpa_reporter/use
  * @returns {Promise}
  * A promise that resolves to the return value of the callback, if any.
  */
-async function withTenantId(tenantId, callback, ...args) {
+async function withTenantId(tenantId, callback) {
     return new Promise((resolve) => {
         requestProviderMiddleware(
             { session: { user: { tenant_id: tenantId } } },
             null,
-            () => resolve(callback(...args)),
+            () => resolve(callback()),
         );
     });
 }
--- a/packages/server/__tests__/arpa_reporter/server/helpers/with-tenant-id.js
+++ b/packages/server/__tests__/arpa_reporter/server/helpers/with-tenant-id.js
@@ -8,12 +8,12 @@ const { requestProviderMiddleware } = require('../../../../src/arpa_reporter/use
  * @returns {Promise}
  * A promise that resolves to the return value of the callback, if any.
  */
-async function withTenantId(tenantId, callback) {
+async function withTenantId(tenantId, callback, ...args) {
     return new Promise((resolve) => {
         requestProviderMiddleware(
             { session: { user: { tenant_id: tenantId } } },
             null,
-            () => resolve(callback()),
+            () => resolve(callback(...args)),
         );
     });
 }
--- a/packages/server/migrations/20240918183745_drop_constraints_to_arpa_subrecipients.js
+++ b/packages/server/migrations/20240918183745_drop_constraints_to_arpa_subrecipients.js
@@ -1,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.schema.alterTable('arpa_subrecipients', (table) => {
+        table.dropUnique(['tenant_id', 'uei']);
+        table.dropUnique(['tenant_id', 'tin']);
+    });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+    return knex.schema.alterTable('arpa_subrecipients', (table) => {
+        table.unique(['tenant_id', 'uei']);
+        table.unique(['tenant_id', 'tin']);
+    });
+};
--- a/packages/server/migrations/20240918183745_drop_constraints_to_arpa_subrecipients.js
+++ b/packages/server/migrations/20240918183745_drop_constraints_to_arpa_subrecipients.js
@@ -2,17 +2,48 @@
  * @param { import("knex").Knex } knex
  * @returns { Promise<void> }
  */
-exports.up = function (knex) {
-    return knex.schema.alterTable('arpa_subrecipients', (table) => {
+exports.up = async function (knex) {
+    await knex.schema.alterTable('arpa_subrecipients', (table) => {
         table.dropUnique(['tenant_id', 'uei']);
-        table.index(['tenant_id', 'uei'], 'idx_arpa_subrecipients_tenant_id_uei_unique', {
-            predicate: knex.whereNotNull('uei'),
-        });
         table.dropUnique(['tenant_id', 'tin']);
-        table.index(['tenant_id', 'tin'], 'idx_arpa_subrecipients_tenant_id_tin_unique', {
-            predicate: knex.whereNotNull('tin'),
-        });
+
+        // table.index(
+        //     ['tenant_id', 'uei'],
+        //     'idx_arpa_subrecipients_tenant_id_uei_unique',
+        //     {
+        //         predicate: knex.whereNotNull('uei'),
+        //     },
+        // );
+        // table.index(
+        //     ['tenant_id', 'tin'],
+        //     'idx_arpa_subrecipients_tenant_id_tin_unique',
+        //     {
+        //         predicate: knex.whereNotNull('tin'),
+        //     },
+        // );
+        // table.unique(
+        //     ['tenant_id', 'uei'],
+        //     {
+        //         indexName: 'idx_arpa_subrecipients_tenant_id_uei_unique',
+        //         predicate: knex.whereNotNull('uei'),
+        //         useConstraint: false,
+        //     },
+        // );
+        // table.unique(
+        //     ['tenant_id', 'tin'],
+        //     {
+        //         indexName: 'idx_arpa_subrecipients_tenant_id_tin_unique',
+        //         predicate: knex.whereNotNull('tin'),
+        //         useConstraint: false,
+        //     },
+        // );
     });
+    await knex.raw(
+        'CREATE UNIQUE INDEX idx_arpa_subrecipients_tenant_id_uei_unique ON arpa_subrecipients (tenant_id, uei) WHERE uei IS NOT NULL',
+    );
+    await knex.raw(
+        'CREATE UNIQUE INDEX idx_arpa_subrecipients_tenant_id_tin_unique ON arpa_subrecipients (tenant_id, tin) WHERE tin IS NOT NULL',
+    );
 };
 
 /**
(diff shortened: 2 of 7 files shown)


TylerHendrickson (commented):
@as1729 A few initial comments for you to consider.

Additionally, would you mind filling out the description for this PR when you get a chance? I think that would help clarify the intended changes regarding the behavior of the new and existing fields (`name`, `uei`, and `tin`) for reviewers, especially since #3372 is showing up as closed by #3503.

packages/server/__tests__/arpa_reporter/server/helpers/with-tenant-id.js
```diff
@@ -8,12 +8,12 @@ const { requestProviderMiddleware } = require('../../../../src/arpa_reporter/use
  * @returns {Promise}
  * A promise that resolves to the return value of the callback, if any.
  */
-async function withTenantId(tenantId, callback) {
+async function withTenantId(tenantId, callback, ...args) {
```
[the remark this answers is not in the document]
as1729: ah, thank you! 
I wrote this code mainly because I was initially doing this:`await withTenantId(TENANT_A, findRecipient('name', 'IAA'));` which was running into a TypeError since I was passing in the response of the function-call rather than the function itself. This is now updated.

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
```diff
@@ -72,21 +60,54 @@ async function getRecipient(id, trns = knex) {
...
-    } else if (tin) {
-        query.where('arpa_subrecipients.tin', tin);
+    if (fieldType === 'uei') {
+        query.where('arpa_subrecipients.uei', value);
+    } else if (fieldType === 'tin') {
+        query.where('arpa_subrecipients.tin', value);
+    } else if (fieldType === 'name') {
+        query.where('arpa_subrecipients.name', value);
     } else {
         return null;
```
tzinckgraf, line 74: It might be worth logging or printing something here or failing if a fieldType is passed in that doesn't match uei, tin or name.

We should also document the expected `fieldType` values in a docstring.

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
```diff
@@ -72,21 +60,54 @@ async function getRecipient(id, trns = knex) {
...
+        }
+    }
+    if (recipient.tin) {
+        const existingRecipient = await findRecipient('tin', recipient.tin, trns);
+        if (existingRecipient) {
+            throw new Error('A recipient with this TIN already exists');
+        }
+    }
+    if (!recipient.tin && !recipient.uei && recipient.name) {
+        const existingRecipient = await findRecipient('name', recipient.name, trns);
```
tzinckgraf, line 99: Can the same name exist across multiple tenants?

as1729: yes, `findRecipient` scopes by tenant by default, so we should only be looking for duplicates within a tenant.

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
as1729: good call- I updated docstrings as well as updated the method to throw an error instead of returning `null` as it can be confusing for the caller.

packages/server/migrations/20240918183745_drop_constraints_to_arpa_subrecipients.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.schema.alterTable('arpa_subrecipients', (table) => {
+        table.dropUnique(['tenant_id', 'uei']);
+        table.dropUnique(['tenant_id', 'tin']);
```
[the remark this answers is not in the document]
as1729: good call- I initially figured I'll move all the validations to the application layer, but this type of index makes more sense. I'm trying to get it working with `knex` `predicate` feature, if that doesn't work will create a raw-index using the statement above.

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
```diff
@@ -72,21 +60,67 @@ async function getRecipient(id, trns = knex) {
...
     }
 
     return query.then((rows) => rows[0]);
 }
 
+async function createRecipient(recipient, trns = knex) {
+    const tenantId = useTenantId();
+    if (!(recipient.uei || recipient.tin || recipient.name)) {
+        throw new Error('recipient row must include a `uei`, `tin`, or `name` field');
+    }
```
TylerHendrickson, line 97: I like how the `createRecipient()` function's various `if ... throw` blocks make the business logic clear to developers and provide friendly error messages for users; I'm wondering if we could preserve that clarity while making better use of the db-level constraints by checking the cause of a failure _after_ inserting rather than checking ahead of time (which can fail to identify the problem ahead of time in a double-click situation or similar race condition). 

I think something like this might work:
```js
async function createRecipient(recipient, trns = knex) {
    const tenantId = useTenantId();
    try {
        return await trns('arpa_subrecipients')
            .insert({ ...recipient, tenant_id: tenantId })
            .returning('*')
            .then((rows) => rows[0]);
    } catch (e) {
        switch (e.constraint) {
            case 'chk_at_least_one_of_uei_tin_name_not_null':
                throw new Error('recipient row must include a `uei`, `tin`, or `name` field');
            case 'idx_arpa_subrecipients_tenant_id_uei_unique':
                throw new Error('A recipient with this UEI already exists');
            case 'idx_arpa_subrecipients_tenant_id_tin_unique':
                throw new Error('A recipient with this TIN already exists');
            // Other `case 'name_of_constraint_or_index'` statements go here...
            default:
                // Something else
                throw e;
        }
    }
}
```
(See my previous comment regarding the `chk_at_least_one_of_uei_tin_name_not_null` constraint.)

packages/server/migrations/20240918183915_add_name_to_arpa_subrecipients.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.schema.table('arpa_subrecipients', (table) => {
+        table.text('name');
+    });
+    await knex.raw(
+        'CREATE UNIQUE INDEX idx_arpa_subrecipients_tenant_id_name_unique ON arpa_subrecipients (tenant_id, name) WHERE name is not null and uei is null and tin is null',
```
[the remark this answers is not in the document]
as1729: > i.e. should duplicate non-null names for the same tenant allowed in any circumstance, regardless of uei or tin?

Right now we're working with insufficient data here. Working assumption is that there _could_ be subrecipients in the system that share the same name but have slightly different UEI/TIN since this was never enforced in the past.

> Also we may want an additional constraint to ensure that at least one of uei/tin/name is provided. If that sounds like a good idea to you, I think that could be implemented as a CHECK constraint like:

This makes sense to add since we enforce this on the application layer, can translate to a database check as well.

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
as1729: This is much cleaner since we don't need to duplicate the validation logic in both layers. Will update the code 👍

packages/server/migrations/20240918183915_add_name_to_arpa_subrecipients.js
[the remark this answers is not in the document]
as1729: Resolved

packages/server/src/arpa_reporter/db/arpa-subrecipients.js
as1729: added in the latest commit

TylerHendrickson (approved):
Looks good!

PR #3553, opened 2024-09-26
Fix: unit tests for db module

greg-adams:
## Description
- Update db tests to resolve build issues where tests occasionally fail due to tests not being isolated from one another.
- Updates seed script to reset id sequences

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


packages/server/__tests__/db/seeds/fixtures.js
```diff
@@ -35,7 +35,7 @@ const agencies = {
         abbreviation: 'Sub SBA',
         code: 'SSBA',
         name: 'State Board of Sub Accountancy',
-        parent: 0,
+        parent: tenants.SBA.id,
```
TylerHendrickson, line 38: Doesn't look like it would cause issues for now, but given that `agencies.parent` is a reference to `agencies.id` (not `tenants.id`), this might result in a hard-to-spot error down the road. 

I'd suggest either reverting this change or breaking it out of the `const agencies` definition so that it can be set to `agencies.accountancy.id`.

TylerHendrickson (changes requested):
This is fantastic – thanks for putting it together! 

One small tweak for you to look at, but otherwise looks great.

packages/server/__tests__/db/seeds/fixtures.js
greg-adams: good catch - I've updated/replaced the old ID and fixed another test issue.

TylerHendrickson (approved):
@greg-adams Looks great!

PR #3592, opened 2024-10-07
Feat: Add Edit/Delete dropdown

greg-adams:
## Description
-  Adds support for soft-deleting a grant note for the current user
   -  Adds new grants note route for note deletion (BE)
   -  Adds new fields to `grant_notes` table to support soft-delete of notes: `is_published` and `updated_at` 
   -  Adds delete as dropdown on user grant note (FE)

## Screenshots / Demo Video
Screenshot from 2024-10-07 16-02-49

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/server/src/lib/grantsCollaboration/notes.js
```diff
@@ -147,9 +147,23 @@ async function getOrganizationNotesForGrant(
...
+    const grantNote = knex
+        .select('id')
+        .from('grant_notes')
+        .where({ grant_id: grantId, user_id: userId });
+
+    await knex('grant_notes_revisions')
+        .whereIn('grant_note_id', grantNote)
+        .del();
+
+    await grantNote.del();
```
TylerHendrickson, line 160: I think it would be preferable to handle this on the Postgres side using `ON DELETE CASCADE`. Unfortunately, that doesn't currently seem to be implemented on the FK constraint, so it's probably best that we do that now since there's no data in Production just yet.

A few details that feel worth mentioning:
- Postgres doesn't support this kind of alteration on foreign keys, so the migration will need to drop the existing FK constraint and then re-add it with the `ON DELETE CASCADE` specification.
- Knex seems to have its own naming convention for foreign key constraints (it doesn't follow the Postgres standard convention of `{tablename}_{columnname(s)}_{suffix}`) so rather than using `knex.raw()`, it might be best to use the Knex schema builder functions (see [`table.dropForeign()`](https://knexjs.org/guide/schema-builder.html#dropforeign) and [`table.foreign()`](https://knexjs.org/guide/schema-builder.html#foreign)).

packages/server/src/routes/grants.js
```diff
@@ -491,6 +459,38 @@ router.get('/:grantId/notes', requireUser, async (req, res) => {
...
+            organizationId,
+            userId,
+            grantId,
+            { cursor, limit: limitInt },
+        );
+
+        // sending the notes as JSON response
+        res.json(notes);
+    } catch (error) {
+        res.status(500).json({ error: 'Failed to retrieve notes' });
```
TylerHendrickson, line 490: Is it possible to re-throw `error` on the next line so that we can preserve the error/trace details, or will that mess with the ability to provide the `Failed to retrieve notes` message in the response?

TylerHendrickson (changes requested):
@greg-adams Everything looks good in testing! A couple server-side things to potentially address and/or get your take on.

packages/server/src/routes/grants.js
greg-adams: yes, I removed the try..catch here as we aren't using this message on the FE and will still result in a 500

packages/server/src/lib/grantsCollaboration/notes.js
greg-adams: yes certainly makes sense here - updated!

packages/server/migrations/20241011165653_cascade-delete-grant-notes.js
```diff
@@ -0,0 +1,44 @@
+const { onUpdateTrigger } = require('../knexfile');
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    const ON_UPDATE_TIMESTAMP_FUNCTION = `
+        CREATE OR REPLACE FUNCTION on_update_timestamp()
```
TylerHendrickson, line 9: Just making it a bit easier to locate quickly during a `psql` session :) 
```suggestion
        CREATE OR REPLACE FUNCTION before_update_set_updated_at()
```

packages/server/knexfile.js
```diff
@@ -111,4 +111,9 @@ module.exports = {
             tableName: 'migrations',
         },
     },
+    // Set trigger for any updated timestamps
+    onUpdateTrigger: (table) => `
+        CREATE TRIGGER ${table}_updated_at BEFORE UPDATE ON ${table}
+        FOR EACH ROW EXECUTE PROCEDURE on_update_timestamp();
```
TylerHendrickson, line 117: Per my previous comment
```suggestion
        FOR EACH ROW EXECUTE PROCEDURE before_update_set_updated_at();
```

TylerHendrickson (changes requested):
@greg-adams A few quick suggestions for the new Postgres function, but I think this is otherwise ready.

greg-adams:
@TylerHendrickson thanks! I incorporated your changes and some small changes to another migration `email_subscriptions` I noticed

packages/server/migrations/20241004230649_add_grant_activity_to_email_subscriptions_notification_type.js
```diff
@@ -17,7 +17,8 @@ exports.up = function (knex) {
  * @param { import("knex").Knex } knex
  * @returns { Promise<void> }
  */
-exports.down = function (knex) {
+exports.down = async function (knex) {
+    await knex('email_subscriptions').where({ notification_type: 'GRANT_ACTIVITY' }).del();
```
TylerHendrickson, line 21: Nice catch 👍

Issue #2961, opened 2024-04-22
[STORY]: Data Migration from 'Status' to 'Mark as Interested'


ekhoffman:
### Full User Story

As an organization, we would like to foster more collaboration between our users to allow for the sharing of knowledge and resources to empower them to more efficiently and effectively apply for the grants discovered on our platform.

As a grantseeker, I want to be able to easily share grants I’ve discovered with my team.

As a grantseeker, I want to be able to easily find collaborators who may be able to help me when preparing and applying for grants.

As a grantseeker, I want to be able to quickly see what grants my team members are interested in.

As an admin, I want to be able to easily surface and resolve issues and blockers my team is running into.

### Why is this issue important?

As we make the move from 'Status' to 'Mark as Interested', we want to ensure that we don't lose any of the work our users have put in to record status so far. Additionally, if we can kickstart usage of this feature with previously entered information, we may be able to boost adoption.

### Action Plan

- [ ] Initial design drafted
- [ ] Initial acceptance criteria drafted
- [ ] High level technical approach & feature flag use determined
- [ ] Designs & Acceptance criteria finalized
- [ ] Eng task breakout complete
- [ ] Assessment of whether partner training & communication actions are needed
- [ ] Acceptance criteria testing completed & passed in staging
- [ ] Feature flags enabled in production
- [ ] Feature flags are removed once the feature is deemed stable

### Engineering Tasks

Add links to engineering tasks

### Acceptance Criteria

**Currently a status is set for an entire team, moving forward 'Mark as Interested' will apply to an individual user.**

1. For the user who set the team status, we will migrate from 'Status' to 'Mark as Interested' as outlined below.
    - If the current team status is 'Interested - Will Apply', we will set the user who selected this status to 'Interested' for that grant.
    - If the current team status is 'Interested - Will Consider', we will set the user who selected this status to 'Interested' and set the Interest Note to 'Will Consider'.
    - If the current team status is 'Interested - Will Support', we will set the user who selected this status to 'Interested' and set the Interest Note to 'Will Support'.
    
_Note: We will not currently be migrating any of the 'Applied' or 'Not Applying' statuses._

ClaireValdivia:
@amyewang based on the implementation plan, is it possible for initial work to be done for this issue, but then to actually run the migration closer to when we are launching the new feature? ideally we want this migration to be able to be run close to the launch so it migrates as much recent data as possible.

amyewang:
[...]

PR #3593, opened 2024-10-08
feat: Migrate grants_interested data to grant_followers and update PUT handler

sushilrajeeva:
…T
## Description
This PR addresses the requirements outlined in the issue #2961 as follows:

1. **Knex Migration:** Migrates existing data from the `grants_interested` table to the `grant_followers` table where the `interested_code` references records with `status_code = 'Interested'`.
2. **PUT Handler Update:** Modified the `PUT /:grantId/interested/:agencyId` route to keep the `grant_followers` table in sync when users update their interest status. After marking a grant as "Interested" via the `markGrantAsInterested()` function, the handler now follows/unfollows grants based on the `interestedCode` provided.

## Screenshots / Demo Video
No UI changes. Backend logic updates and database migrations.
### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [X] Added PR reviewers


packages/server/src/routes/grants.js
```diff
@@ -431,15 +431,33 @@ router.put('/:grantId/interested/:agencyId', requireUser, async (req, res) => {
...
+            .select('status_code')
+            .where({ id: interestedCode })
+            .first();
+
+        // Follow  or Unfollow the grant based on the interestedCode
+        if (interestedStatus?.status_code === 'Interested') {
+            await followGrant(knex, grantId, user.id);
+        } else {
+            await unfollowGrant(knex, grantId, user.id);
+        }
```
TylerHendrickson, line 453: This should come before the call to `db.markGrantAsInterested()`, and be wrapped in a transaction so that we can roll back in case `db.markGrantAsInterested()` fails:
```js
// Follow  or Unfollow the grant based on the interestedCode
const trx = await knex.transaction();
if (interestedStatus?.status_code === 'Interested') {
    await followGrant(trx, grantId, user.id);
} else {
    await unfollowGrant(trx, grantId, user.id);
}
try {
    // Calling the existing function to mark the grant as interested
    await db.markGrantAsInterested({
        grantId,
        agencyId,
        userId: user.id,
        interestedCode,
    });
} catch (error) {
    // Roll back the follow/unfollow operation since marking grant as interested failed
    trx.rollback();
    console.error('Error in marking as interested:', error);
    throw error;
}
// Commit the follow/unfollow operation since everything succeeded
trx.commit();
```

Note: The order of `db.markGrantAsInterested()` and either `followGrant()` or `unfollowGrant()` wouldn't matter if we were able to pass the transaction to `db.markGrantAsInterested()`, since that would allow us to commit or rollback as a unit. Modifying that function is arguably out of scope for this issue, but just to illustrate what I mean:
```js
async function markGrantAsInterested({
    trx = knex, grantId, agencyId, userId, interestedCode,
}) {
    const results = await trx(TABLES.grants_interested)
        .insert({
            agency_id: agencyId,
            grant_id: grantId,
            user_id: userId,
            interested_code_id: +interestedCode,
        });
    return results;
}
```

packages/server/src/routes/grants.js
```diff
@@ -431,15 +431,33 @@ router.put('/:grantId/interested/:agencyId', requireUser, async (req, res) => {
...
+        // Retruning updated interested agencies
+        const interestedAgencies = await db.getInterestedAgencies({ grantIds: [grantId], tenantId: user.tenant_id });
+        res.json(interestedAgencies);
+    } catch (error) {
+        console.error('Error in marking as interested:', error);
+        res.status(500).send({ success: false, error: 'Internal Server Error' });
+    }
```
TylerHendrickson, line 460: I'd prefer to remove this wrapping try/catch block since we're not inspecting it or otherwise using it to provide any additional context to the user if something were to fail. Otherwise the try/catch will suppress the error and make this harder to monitor and debug (in Datadog).

TylerHendrickson (changes requested):
@sushilrajeeva This is looking good. I have a few comments related to error handling and atomicity of database writes for you to take a look at.

Additionally, it would be preferable for this PR to add tests for the new behavior in `packages/server/__tests__/api/grants.test.js`.

phm200 (approved):
I tested this locally and both the migration and PUT behavior met expectations.

Specifically I tested the following locally:
- w/feature flag off, set myself as interested in two grants, then ran this branch, ran migration and turned on feature flag -> was following those two grants as expected
- turned off feature flag, marked as interested in a new grant, turned on feature flag -> was following the grant as expected
- spot checked a few grants I had not marked as interested -> not following the grant as expected
- with feature flag off, marked as interested, then un-marked/deleted interested, turn on feature flag -> not following the grant as expected

ClaireValdivia:
@sushilrajeeva just want to make sure you see Ty's comments above for change requests!

packages/server/src/routes/grants.js
sushilrajeeva: @TylerHendrickson adding transaction to rollback if markGrantAsInterested fails is a great suggestion, thanks for the detailed implementation. I have incorporated your suggestion, I will push the code with the suggested changes.

sushilrajeeva: @TylerHendrickson I have added await to the transaction.commit() and transaction.rollback().

sushilrajeeva:
@TylerHendrickson I have updated the test case for the PUT function. Specifically, I modified the previous implementation by adding a GET operation after performing the PUT request. This allows me to validate whether the agency.agency_id returned is correct. If there are any additional checks or improvements you'd recommend for this test, please let me know.

TylerHendrickson (approved):
@sushilrajeeva Thanks for the updates, looks great!

PR #3605, opened 2024-10-09
Chore: Minor doc update for local dev and staging deployment

phm200:
## Description
These are two items I noticed while onboarding and doing my first PR
1. Vitest extension was not working correctly because I was using NVM and vitest extension was finding the wrong version of Node
2. Deployed my first PR to staging and didn't realize database migrations must be manually run

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/usdr-gost.code-workspace
+++ b/usdr-gost.code-workspace
@@ -31,7 +31,6 @@
 		//   you see vitest extension errors in the output. The extension
 		//   sometimes cannot find the right version of Node unless it using a terminal shell
 		//   to spawn the test running process.
-		//[]
 		//  see: https://github.com/vitest-dev/vscode/issues/486 
 		//         https://github.com/vitest-dev/vscode/pull/490
 		//  "vitest.shellType": "terminal",
--- a/usdr-gost.code-workspace
+++ b/usdr-gost.code-workspace
@@ -27,12 +27,7 @@
 		// to manage linting/fixing, but users can have other settings via "formatOnSave".
 		// If both are on, the editor can "fix" things that are immediately flagged as lint failures.
 		"editor.formatOnSave": false,
-		// Uncomment the vitest.shellType setting if the "Testing" panel is empty and/or 
-		//   you see vitest extension errors in the output. The extension
-		//   sometimes cannot find the right version of Node unless it using a terminal shell
-		//   to spawn the test running process.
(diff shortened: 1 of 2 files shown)


lsr-explore (approved):

lsr-explore:
Thanks for adding these!

usdr-gost.code-workspace
```diff
@@ -27,5 +27,12 @@
 		// to manage linting/fixing, but users can have other settings via "formatOnSave".
 		// If both are on, the editor can "fix" things that are immediately flagged as lint failures.
 		"editor.formatOnSave": false,
+		// Uncomment the vitest.shellType setting if the "Testing" panel is empty and/or 
+		//   you see vitest extension errors in the output. The extension
+		//   sometimes cannot find the right version of Node unless it using a terminal shell
+		//   to spawn the test running process.
+		//  see: https://github.com/vitest-dev/vscode/issues/486 
+		//         https://github.com/vitest-dev/vscode/pull/490
+		//  "vitest.shellType": "terminal",
```
TylerHendrickson, line 36: @phm200 Are there any downsides to this being uncommented by default?

phm200: good question. I'm not entirely sure. Based on the GH activity around this feature, this should still work the same for folks who were fine with the default "child process" setting. If you or someone else whose vitest plugin is working normally can try it out and confirm it works, then I can uncomment it for all

to your point, if we cannot uncomment it for all, probably better for me to just add a note to the readme and people can make that change in their own settings json and not have to make an un-commitable change to the .code-workspace every time

TylerHendrickson: Just checked, and I seem to be able to run tests using the vitest plugin successfully, regardless of whether this line is (un)commented, as well as when explicitly set to `child_process`.

In case you think it matters, I'll add that my dev environment are fairly ephemeral, so the only Node I have is what's installed by nvm (there's no global/system default), and I only typically have the version specified in this repo's `.nvmrc` installed.

phm200: will do, uncommented in [566a194](https://github.com/usdigitalresponse/usdr-gost/pull/3605/commits/566a1946fafa28669bac5687cbb69637008b8064)

PR #3634, opened 2024-10-14
chore(grants): add eslint-plugin-vuejs-accessibility and disable errors flagged

lsr-explore:
## Description

Part of a Design Record to add support for 

[eslint-plugin-vuejs-accessibility](https://www.npmjs.com/package/eslint-plugin-vuejs-accessibility)

* Add the plugin
* Disable rules flagged

## Screenshots / Demo Video

Rules flagged
image

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers


docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
```diff
@@ -0,0 +1,88 @@
...
+- Good, because we would catch accessibility issues in our Vue.js applications before they are deployed to production.
+<!-- numbers of pros and cons can vary -->
+
+### eslint-plugin-vue-a11y
+
+[NPM package - eslint-plugin-vue-a11y](https://www.npmjs.com/package/eslint-plugin-vue-a11y)  <!-- optional -->
+
+- Bad, because it is not being actively maintained - was last published 5 years ago and only has 3k weekly downloads.
+- Good, because we would catch accessibility issues in our Vue.js applications before they are deployed to production.
+<!-- numbers of pros and cons can vary -->
```
TylerHendrickson, line 63: Does this plugin have any advantages relative to `eslint-plugin-vuejs-accessibility`, or would it be redundant/unnecessary to adopt both?

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Positive Consequences <!-- optional -->
+
+- Catching the accessibility issues in our Vue.js applications before they are deployed to production.
+- Helping developers to write more accessible Vue.js applications.
+- Reducing future remediation costs.
+
+### Negative Consequences <!-- optional -->
+
+- Some devs may not know how to fix accessibility issues in Vue.js applications, and this may cause some friction in development or result in a number of rules being disabled.
```
TylerHendrickson, line 36: I think it's likely that we could encounter this from time to time, although IMO that's certainly not a deal-breaker. That said, I'm wondering if we might be able to mitigate issues for newcomers by providing documentation – do you think there are any useful resource(s) that would be worth linking to from our `docs/` directory, as a starting point?

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Option 1 Code Example <!-- optional -->
+
+n/a - no changes from current setup
+
+### Option 2 Code Example <!-- optional -->
+
+[Changes are in this PR with the Decision Record - #3634](https://github.com/usdigitalresponse/usdr-gost/pull/3634/files)
+
+I can separate out the changes into separate PR if needed.
```
TylerHendrickson, line 75: Assuming we end up adopting at least one of the recommended linters, I don't think it should be necessary to separate the code changes from this PR into a separate PR. However, I do think that we should make a plan to address the newly-introduced `eslint-disable` directives, which might entail having individual issues (either per-file, or else something more aligned with perceived level of effort) along with a parent issue we can use to track burn-down progress.

TylerHendrickson (commented):
@lsr-explore Adding a few comments/considerations to the discussion. Thanks for putting this together!

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
lsr-explore: It would be redundant to adopt both.

https://www.npmjs.com/package/eslint-plugin-vuejs-accessibility

On the NPM page - it mentions that it is built on top of the eslint-plugin-vue-a11y

image

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
lsr-explore: If this was adopted, I would write up some introductory documentation.   The package has some documentation, but their solution isn't always the most optimal for accessibility.   

https://vue-a11y.github.io/eslint-plugin-vuejs-accessibility/rules/click-events-have-key-events.html


---------------
There is a ton of resources - will need to look for something that would be not overwhelming.
* [Deque has some good resources](https://media.dequeuniversity.com/public/en/docs/deque_web_accessibility_checklist.pdf)

I like your idea of creating issues if items need to be temporarily disabled, and we can use a common tag, so devs can see examples from the code.

------------

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
lsr-explore: Sounds good.  If this is adopted, I plan to work on addressing the issues.   Before merging, I'll create tech debt tickets and add them to the disable lines.  I agree with your suggestion of one ticket per page.

One of the findings - autofocus would need to be reviewed w/ UX as it would change the behavior of the application.  I don't know if this behavior was added by devs or specified by UX.

* https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

> [Accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_concerns)
> Automatically focusing a form control can confuse visually-impaired people using screen-reading technology and people with cognitive impairments. When autofocus is assigned, screen-readers "teleport" their user to the form control without warning them beforehand.
> 
> Use careful consideration for accessibility when applying the autofocus attribute. Automatically focusing on a control can cause the page to scroll on load. The focus can also cause dynamic keyboards to display on some touch devices. While a screen reader will announce the label of the form control receiving focus, the screen reader will not announce anything before the label, and the sighted user on a small device will equally miss the context created by the preceding content.

docs/decisions/0010-include-eslint-plugin-vuejs-accessibility.md
TylerHendrickson: That would be excellent – thank you!

TylerHendrickson (approved):
@lsr-explore I think this proposal is in a good place and we haven't received any other concerns/pushback from the rest of the team, so I think we can consider this "Accepted" and move on with the items discussed in the comments. Thanks again for putting this together!

TylerHendrickson (approved):

PR #3734, opened 2024-11-18
Fix: cascade delete `users` to grant notes and follows

TylerHendrickson:
## Description

Adds `ON DELETE CASCADE` to FK references so that when a user is deleted, associated rows in `grant_notes` and `grant_followers` tables are also deleted.

### DIFF
--- a/packages/server/migrations/20241118203005_cascade-delete-grant-notes-for-user.js
+++ b/packages/server/migrations/20241118203005_cascade-delete-grant-notes-for-user.js
@@ -1,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.schema.table('grant_notes', (table) => {
+        table.dropForeign('user_id');
+        table.foreign('user_id').references('users.id').onDelete('CASCADE');
+    });
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function (knex) {
+    await knex.schema.table('grant_notes', (table) => {
+        table.dropForeign('user_id');
+        table.foreign('user_id').references('users.id');
+    });
+};
--- a/packages/server/migrations/20241118203302_cascade-delete-followed-grants-for-user.js
+++ b/packages/server/migrations/20241118203302_cascade-delete-followed-grants-for-user.js
@@ -1,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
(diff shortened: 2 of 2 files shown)


packages/server/migrations/20241118203005_cascade-delete-grant-notes-for-user.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+    await knex.schema.table('grant_notes', (table) => {
+        table.dropForeign('user_id');
```
as1729, line 7: do these tables have existing data?

TylerHendrickson: Only the FK *constraint* is being dropped (the actual data remains). In some cases, it may be possible for an invalid `user_id` value to get set on a row in the time between when the old FK is dropped and the new FK is added. However, since Knex wraps migrations in a transaction, the table will be locked until the transaction is committed, so we don't need to worry about that happening here.

For reference, here's the SQL for the `grant_notes` migration – you can see that it only affects the `grant_notes_user_id_foreign` constraint, not the actual `user_id` column:
```sql
BEGIN;
ALTER TABLE grant_notes 
    DROP CONSTRAINT grant_notes_user_id_foreign;
ALTER TABLE grant_notes 
    ADD CONSTRAINT grant_notes_user_id_foreign 
        FOREIGN KEY (user_id) 
        REFERENCES users (id) 
        ON DELETE CASCADE;
COMMIT;

TylerHendrickson: **Edit to add:** In a high-traffic scenario where a lock might be disruptive, it might be better to not use a transaction and do something like this instead:
1. Rename the constraint to `<constraint_name>__old`
    ```sql
    ALTER TABLE grant_notes
       RENAME CONSTRAINT grant_notes_user_id_foreign 
       TO grant_notes_user_id_foreign__old;
    ```
2. Create the new constraint with the original name (`NOT VALID` may not be necessary here, but we can use it to skip initial validation of the new FK constraint, which might take a long time for a big table; we can do this because we know that the old constraint continues to ensure that `user_id` values are correct):
    ```sql
    ALTER TABLE grant_notes 
        ADD CONSTRAINT grant_notes_user_id_foreign 
            NOT VALID
            FOREIGN KEY (user_id) 
            REFERENCES users (id) 
            ON DELETE CASCADE;
     ```
3. Drop the old constraint, which is now redundant:
    ```sql
    ALTER TABLE grant_notes 
      DROP CONSTRAINT grant_notes_user_id_foreign;
    ```

as1729 (approved):

PR #3851, opened 2025-01-03
3802 - Ensure evidence based total spend value is derived from the most recent upload

as1729:
## Description

Some values in the ARPA audit report must be derived from the most recent upload-file rather than an aggregate of of all the upload-files.

In this case, we updated the `Evidence Based Total Spend` value in the KPI tab to reflect the new logic and take the most-recent value.

### After
The latest value of $35,000 is being picked.
image

### Before
The KPI tab is returning the summation of two values $100,000 and $35,000 to return $135,000.
image

## Testing
1. Retrieve the two uploads needed to test from this directory here: https://drive.google.com/drive/folders/1YCnXLi-j3k8m3qktKrt9MRf20B5YsUDD
2. Upload both workbooks in the order `FIRST_UPLOAD` then `SECOND_UPLOAD` into the app through the Submit Workbook Flow. http://localhost:8080/arpa_reporter/new_upload
3. Verify that both uploads are validated.
4. In a separate tab hit this endpoint to synchronously generate the Audit Report: http://localhost:8080/api/audit_report
5. Verify that the audit report reflects $35,000 when generated from the current branch.
6. Switch to the `main` branch and run the same endpoint: http://localhost:8080/api/audit_report - This time verify that the audit report contains the value $135,000.

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [x] Added PR reviewers


packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -478,12 +502,13 @@ async function createKpiDataGroupedByProject(periodId, tenantId, logger = log) {
...
             project: { id: projectId, totalRecords: projectRecords.length },
         });
+        const mostRecentRecord = getMostRecentRecordForProject(projectRecords, projectLogger);
         projectLogger.debug('populating row from records in project');
         const row = {
             'Project ID': projectId,
             'Number of Subawards': 0,
             'Number of Expenditures': 0,
-            'Evidence Based Total Spend': 0,
+            'Evidence Based Total Spend': mostRecentRecord?.content.Spending_Allocated_Toward_Evidence_Based_Interventions || 0,
```
lsr-explore, line 511: is content always guaranteed to be present?  In the test data, the first record did not have a content.  I see the comment that says it should be filtered out - is lack of content obj a filtering criteria?

packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -465,6 +465,30 @@ async function createReportsGroupedBySubAward(periodId, tenantId, dateFormat = R
     return rowsFormatted;
 }
 
+function getMostRecentRecordForProject(records, logger = log) {
+    let mostRecentRecord = {};
+    // Ensures we are only looking at records that are in the EC-tabs rather than the other tabs
+    records = records.filter((record) => Object.keys(EXPENDITURE_CATEGORIES).includes(record.type));
```
lsr-explore, line 471: Should any other fields be checked?  e.g. subcategory? validated_at?   If it has been validated, does it mean that it is valid and there are no errors?

packages/server/src/arpa_reporter/lib/audit-report.js
```diff
@@ -493,7 +518,6 @@ async function createKpiDataGroupedByProject(periodId, tenantId, logger = log) {
             if ((r.content.Current_Period_Expenditures__c || 0) > 0) {
                 row['Number of Expenditures'] += 1;
```
lsr-explore, line 519: Should reviewing how the "Number of expenditures" are calculated part of this PR?

image

packages/server/src/arpa_reporter/lib/audit-report.js
as1729: This check shouldn't be necessary since the `records` here are already validated. Its a few steps removed, but if we follow the stack they should originate from this function which is responsible for doing the initial filtering and making sure that the data is already validated prior to it getting to this point: https://github.com/usdigitalresponse/usdr-gost/blob/main/packages/server/src/arpa_reporter/services/records.js#L207

packages/server/src/arpa_reporter/lib/audit-report.js
as1729: Yes thanks for catching that, I missed it! Will add shortly.

packages/server/src/arpa_reporter/lib/audit-report.js
as1729: This is a good question - records that are filtered by type `ec*` should always have a `content`. Let me verify that this is the case and if not, I'll add something to check for the existence of `content`.

lsr-explore (approved):

lsr-explore:
Thanks for the quick responses @as1729 - I approved.

packages/server/src/arpa_reporter/lib/audit-report.js
as1729: Confirming based on Claire's latest comment on the issue that these fields are working correctly and don't need to be changed.

packages/server/src/arpa_reporter/lib/audit-report.js
as1729: Based on the latest bug- it seems prudent to add a check to make sure these fields are present. Since the data comes directly from the spreadsheets and the users can input anything they want there is no guarantee that all the data is fully available. I added these checks in the latest PR. #3863

Issue #3933, opened 2025-01-28
[Issue]: Update picklist for Sub_Category__c


ClaireValdivia:
### Why is this issue important?

Treasury changes the EC Code name for the 7.3 selection

### Current State

Previous value for 7.3 selection was: "7.3-Costs Associated with Satisfying Certain Legal and Administrative
Requirements of the SLFRF Program After December 31, 2024"

### Expected State

- [ ] Input Template: On the cover tab, detailed expenditure category dropdown should reflect the new value: "7.3-Costs Associated with Satisfying the Administrative and Other Legal Requirements of the SLFRF Program After the Obligation Deadline has Passed" -- Note - non-eng task, Claire will update input template
- [ ] Upload validation - new value should be accepted for the picklist (i.e. validation error should not generate if update value is selected)
- [ ] Output template: On treasury report files, selection of the 7.3 code should reflect the updated name for the field `Sub_Category__c`
- [ ] Audit report: selection of the 7.3 code should reflect the updated name for the field `Sub_Category__c`

### Implementation Plan

The following functions need to be added...

### Relevant Code Snippets

```shell

```

sophaskins:
I've pushed up a PR to address this: https://github.com/usdigitalresponse/usdr-gost/pull/4019

I believe this covers all of the expected items, but I still need to test out actually generating an output report and audit report (vs just relying on the test coverage).

I want to note that the name of this field isn't actually validated when a workbook is uploaded (I think it may care about the '7.3' portion but none of the rest?).

PR #4019, opened 2025-02-19
#3933 update name of category 7.3

sophaskins:
## Description
This pull request updates the Input Template and Output Template to refer to EC 7.3 in a way we believe Treasury will accept:
```diff
-7.3-Costs Associated with Satisfying Certain Legal and Administrative Requirements of the SLFRF Program After December 31, 2024
+7.3-Costs Associated with Satisfying the Administrative and Other Legal Requirements of the SLFRF Program After the Obligation Deadline has Passed,
```
The updated workbook only has a change to the name of that EC and the version number (bumping it to 20250219). The change to the workbook isn't strictly necessary here, as the value from the workbook doesn't get directly passed through to the Output Template. That is handled in `packages/server/src/arpa_reporter/lib/arpa-ec-codes.js`.

## Screenshots / Demo Video
An upload using the new template and the EC 7.3 validates properly on my local checkout:
Successful validate on local

An upload using the _old_/current template has a warning because of the old version:
failed validate on local with old template

### Automated and Unit Tests
- [x] Added Unit tests

### Manual tests for Reviewer
- [ ] ~~Added steps to test feature/functionality manually~~ n/a

## Checklist
- [x] Added PR reviewers


sophaskins:
I alluded to this above, but we do not actually validate (or even _read_) the text of the name of EC 7.3 on the upload side - we just use the value from `packages/server/src/arpa_reporter/lib/arpa-ec-codes.js`. You can see one side-effect of that when trying to use an update workbook with non-updated code (from staging) - it passes validation with no issues even though the category name is different:

image

I also noticed this while updating the tests - the update on line 191 of `packages/server/__tests__/arpa_reporter/server/services/generate-arpa-report.spec.js` isn't actually necessary to make the tests pass.

PR #4041, opened 2025-02-24
Allow ARPA Exporter tasks to ListBucket

TylerHendrickson:
## Description

This PR adds `ListBucket` permissions on the audit report bucket to FFE tasks. It's intended (along with the previous addition of `HeadObject` permissions in #4040) to fix the "Forbidden" error that these tasks are getting when running `download_fileobj()`.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -283,6 +283,16 @@ data "aws_iam_policy_document" "arpa_audit_report_rw_reports_bucket" {
   }
 }
 
+data "aws_iam_policy_document" "arpa_audit_report_list_bucket" {
+  statement {
+    sid = "ListBucketObjects"
+    actions = [
+      "s3:ListBucket",
+    ]
+    resources = [module.api.arpa_audit_reports_bucket_arn]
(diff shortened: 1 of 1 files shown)


terraform/main.tf
```diff
@@ -491,8 +501,9 @@ module "arpa_exporter" {
     access_point_id = module.api.efs_data_volume_access_point_id
   }]
   additional_task_role_json_policies = {
-    rw-audit-reports-bucket = data.aws_iam_policy_document.arpa_audit_report_rw_reports_bucket.json
-    send-emails             = module.api.send_emails_policy_json
+    list-audit-reports-bucket = data.aws_iam_policy_document.arpa_audit_report_list_bucket.json
```
as1729, line 504: Needs to be added to arpa_exporter module instead

TylerHendrickson: @as1729 It actually is (caught me off guard too for a minute) – I think there's something weird in the diff view.

as1729 (approved):

PR #4079, opened 2025-02-28
fix: ensure that .xls and .xlsx files are handled correctly

as1729:
## Description
- This fixes the error where buil_zip is unable to find the source file in EFS.

## Screenshots / Demo Video
N/A

## Testing
Existing unit tests are changed

### Automated and Unit Tests
- [ ] Added Unit tests

### Manual tests for Reviewer

## Checklist
- [ ] Added PR reviewers

### DIFF
--- a/packages/server/src/arpa_reporter/services/full-file-export.js
+++ b/packages/server/src/arpa_reporter/services/full-file-export.js
@@ -9,6 +9,7 @@ const metadataFileKey = (organizationId) => `full-file-export/org_${organization
 const zipFileKey = (organizationId) => `full-file-export/org_${organizationId}/archive.zip`;
 
 async function getUploadsForArchive(organizationId) {
+    /* eslint-disable */
     const uploads = await knex.raw(`
         WITH uploads_for_treasury_export AS (
             SELECT DISTINCT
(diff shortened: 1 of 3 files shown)


arpa-exporter/src/worker.py
```diff
@@ -76,7 +76,8 @@ def build_zip(fh: typing.BinaryIO, source_uploads: typing.Iterator[UploadInfo])
     with zipfile.ZipFile(fh, "a") as archive:
         for upload in source_uploads:
             files_checked += 1
-            source_path = os.path.join(DATA_DIR, f"{upload.upload_id}.xlsm")
+            file_extension = upload.path_in_zip.split(".")[-1]
+            source_path = os.path.join(DATA_DIR, f"{upload.upload_id}.{file_extension}")
```
TylerHendrickson, line 80: This will handle odd edge cases (no file extension, dots in directory names, etc.) slightly better:
```suggestion
            _, file_extension = os.path.splitext(upload.path_in_zip)
            source_path = os.path.join(DATA_DIR, f"{upload.upload_id}{file_extension}")
```

Note: The second value in the 2-tuple returned by `os.path.splitext()` contains the dot (assuming there is one) before the extension (e.g. `os.path.splitext("something.txt") -> ("something", ".txt")`), so important to leave the `.` separator out of the f-string literal.

packages/server/src/arpa_reporter/services/full-file-export.js
```diff
@@ -28,13 +29,13 @@ async function getUploadsForArchive(organizationId) {
...
+                WHEN ue.id IS NOT NULL THEN '/' || rp.name || '/Final Treasury/' || regexp_replace(u1.filename, '\.[^.]+$', '') || '--' || u1.id || '.' || split_part(filename, '.', array_length(string_to_array(filename, '.'), 1))
                 WHEN u1.validated_at IS NOT NULL
-                AND ue.id IS NULL THEN '/' || rp.name || '/Not Final Treasury/Valid files/' || SPLIT_PART(u1.filename, '.xlsm', 1) || '--' || u1.id || '.xlsm'
-                ELSE '/' || rp.name || '/Not Final Treasury/Invalid files/' || SPLIT_PART(u1.filename, '.xlsm', 1) || '--' || u1.id || '.xlsm'
+                AND ue.id IS NULL THEN '/' || rp.name || '/Not Final Treasury/Valid files/' || regexp_replace(u1.filename, '\.[^.]+$', '') || '--' || u1.id || '.' || split_part(filename, '.', array_length(string_to_array(filename, '.'), 1))
+                ELSE '/' || rp.name || '/Not Final Treasury/Invalid files/' || regexp_replace(u1.filename, '\.[^.]+$', '') || '--' || u1.id || '.' || split_part(filename, '.', array_length(string_to_array(filename, '.'), 1))
```
TylerHendrickson, line 38: I feel like this string-formatting/translation is pretty squarely app-level logic that should be handled in JS code. IMO it doesn't feel appropriate/necessary for the SQL `select` statement for a few reasons:
- Harder to read
- It's arguably not related to fetching data
- There doesn't seem to be a performance-related reason or other practical advantage having it that way (e.g. it's not helping us avoid additional DB queries). 
- Adds bloat to the query that isn't particularly helpful in debugging scenarios (especially since most tools will mask the literals in logged DB output).
- Where the SQL implementation (although impressive in its own right) is composed of regex and string split/join operations that might be vulnerable to odd edge cases, Node.js has a built-in [`path` module](https://nodejs.org/api/path.html) that provides functions for these kinds of operations.

IMO it would be better to fetch the necessary values to make this determination and implement this in JavaScript.

TylerHendrickson (changes requested):
@as1729 Couple of suggestions, mostly with a common theme of avoiding custom string manipulations in favor of native functions for working with file names and paths.

packages/server/src/arpa_reporter/services/full-file-export.js
as1729: As discussed via Slack- I updated the code to handle the complex business logic in javascript and left the SQL query to focus on retrieving the data. This is more readable as a result and easier to make changes in the future.

TylerHendrickson (approved):
@as1729 Looks good!

PR #4093, opened 2025-03-05
Add IAM permissions for data migration

TylerHendrickson:
## Description

This PR allows per-stage configurations that conditionally allow API tasks in ECS Fargate to access S3 buckets/objects for the purposes of data migration. 

When at least one bucket name is supplied to the `var.data_migration_destination_bucket_names` input variable, an additional inline IAM policy called `rw-data-migration-buckets` will be attached to the ECS Task (runtime) role for `gost-api` tasks.

This PR includes configuration for both Staging and Production environments.


sophaskins (approved):

terraform/modules/gost_api/task.tf
```diff
@@ -233,12 +233,17 @@ resource "aws_iam_role" "task" {
...
-  }
+  for_each = !var.enabled ? {} : merge(
+    {
+      connect-to-postgres   = module.connect_to_postgres_policy.json
+      ecs-exec              = module.ecs_exec_policy.json
+      send-emails           = module.send_emails_policy.json
+      rw-arpa-audit-reports = module.access_arpa_reports_bucket_policy.json
+    },
+    length(var.data_migration_bucket_names) == 0 ? {} : {
+      rw-data-migration-buckets = module.data_migration_bucket_names[0].json
```
sophaskins, line 244:
```suggestion
      rw-data-migration-buckets = var.data_migration_bucket_names[0].json
```

TylerHendrickson: Looks like I was mistakenly referencing the name of the variable and not the name of the module providing the JSON policy document. Should be fixed now.

sophaskins (approved):

PR #4097, opened 2025-03-06
Fix duplicate entries in FFE zipfiles

TylerHendrickson:
## Description

This PR fixes a bug that is currently resulting in duplicate entries being created in FFE zipfiles produced by `arpa-exporter`. Currently, we attempt to avoid writing duplicate files by checking for the presence of the zip path (from CSV metadata's `path_in_zip` value) in the existing contents of the zipfile. However, Python performs some amount of normalization to these values when writing entries, so files that require normalization (i.e. are not zip-compatible) are considered missing. Since zip files allow for duplicate entries, one current problem is that subsequent runs of `arpa-exporter` will (at minimum) duplicate the size of the zip file.

In our case, this problem is manifesting because we check for absolute paths in the zipfile, but zipfiles normalize absolute paths to relative (to the root of the zipfile) when these entries are written.

The solution provided by this PR is twofold:
1. Pre-normalize zip entry paths before checking whether an entry with that path already exists, so we're comparing apples to apples. 
2. Tweak the Full File Export metadata CSV generation logic so that `path_in_zip` column values are not formatted as absolute paths (that is, we remove the leading `/` slash from these values).

### DIFF
--- a/arpa-exporter/src/lib/logging.py
+++ b/arpa-exporter/src/lib/logging.py
@@ -23,7 +23,7 @@ shared_processors: List[Callable] = [
 ]
 
 processors: List[Callable] = shared_processors + []
-if sys.stderr.isatty():
+if sys.stderr.isatty():  # pragma: nocover
     processors += [
         structlog.dev.ConsoleRenderer(),
     ]
--- a/arpa-exporter/tests/conftest.py
+++ b/arpa-exporter/tests/conftest.py
@@ -7,6 +7,7 @@ import moto.ses
 import moto.ses.models
 import pytest
 
+os.environ.setdefault("DD_TRACE_ENABLED", "false")
 os.environ.setdefault("TASK_QUEUE_URL", "https://example.com/queue")
 os.environ.setdefault("TASK_QUEUE_RECEIVE_TIMEOUT", "1")
 os.environ.setdefault(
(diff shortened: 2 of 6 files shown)


arpa-exporter/src/lib/logging.py
```diff
@@ -23,7 +23,7 @@
 ]
 
 processors: List[Callable] = shared_processors + []
-if sys.stderr.isatty():
+if sys.stderr.isatty():  # pragma: nocover
```
TylerHendrickson, line 26: Unrelated to the bug, I just noticed this was throwing off coverage metrics, but this expression will never evaluate differently in a realistic unit-testing scenario.

arpa-exporter/tests/conftest.py
```diff
@@ -7,6 +7,7 @@
 import moto.ses.models
 import pytest
 
+os.environ.setdefault("DD_TRACE_ENABLED", "false")
```
TylerHendrickson, line 10: Also unrelated to the bug. I just noticed that `ddtrace` was logging output (complaining that it couldn't reach Datadog) during unit test execution. This disables `ddtrace` unless it has been explicitly enabled (which may be the case if Datadog's testing suite is in-use, e.g. in a CI environment).

as1729 (approved):

PR #4145, opened 2025-03-21
Use environment variables to configure allowed email types

TylerHendrickson:
## Description

This PR introduces two API server environment variables, `ENABLED_EMAIL_TYPES` and `DISABLE_ALL_EMAILS`, which allows emails to be selectively enabled or else entirely prevented from being sent. This behavior manifests as a module level constant array, `ENABLED_EMAIL_TYPES`, which will contain some subset of values from the `tags.emailTypes` object in `packages/server/src/lib/email/constants.js`.

- `DISABLE_ALL_EMAILS` env var: When set to a non-empty value, no emails will be sent, regardless of other environment configuration.
- `ENABLED_EMAIL_TYPES` env var: When set, is expected to provide a comma-separated string of zero or more email types (per the types defined in `packages/server/src/lib/email/constants.js`):
  - When the environment variable is unset or an empty string, all email types are enabled, allowing us to maintain existing behavior by default.
  - When the environment variable has a non-empty value, only the named emails (multiple can be provided with a by comma-separating individual email types) are allowed to send. If a value is provided that does not match a known email type, it is ignored and a warning is logged. If the environment variable is non-empty but no emails are enabled (i.e. no provided values match a known email type), another warning is logged.

Since the `ENABLED_EMAIL_TYPES` constant is loaded with the `packages/server/src/lib/email.js` module, log messages related to the configuration process are emitted whenever this module is first loaded (normally on startup of the API server).

Note that this change only introduces the ability to enable emails based on environment configuration – no environment variables are modified for any of our deployment environments in this PR.

### DIFF
--- a/packages/server/src/lib/email.js
+++ b/packages/server/src/lib/email.js
@@ -85,6 +85,12 @@ async function deliverEmail({
     subject,
     emailType,
 }) {
+    if (!ENABLED_EMAIL_TYPES.includes(emailType)) {
+        log.info({ emailType, enabledEmailTypes: ENABLED_EMAIL_TYPES },
+            'this type of email is not enabled so it will not be sent');
+        return null;
+    }
+
     let userTags = [];
     const recipientId = await db.getUserIdForEmail(toAddress);
     const activeContext = tracer.scope().active()?.context();
--- a/packages/server/src/lib/email.js
+++ b/packages/server/src/lib/email.js
@@ -30,6 +30,44 @@ const HELPDESK_EMAIL = 'grants-helpdesk@usdigitalresponse.org';
 const GENERIC_FROM_NAME = 'USDR Grants';
 const GRANT_FINDER_EMAIL_FROM_NAME = 'USDR Federal Grant Finder';
 const ARPA_EMAIL_FROM_NAME = 'USDR ARPA Reporter';
+const ENABLED_EMAIL_TYPES = Object.freeze((() => {
+    let logger = log.child({
+        env: {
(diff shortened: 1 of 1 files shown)


packages/server/src/lib/email.js
```diff
@@ -47,6 +85,12 @@ async function deliverEmail({
     subject,
     emailType,
 }) {
+    if (!ENABLED_EMAIL_TYPES.includes(emailType)) {
+        log.info({ emailType, enabledEmailTypes: ENABLED_EMAIL_TYPES },
+            'this type of email is not enabled so it will not be sent');
+        return null;
+    }
```
TylerHendrickson, line 92: This is the actual check that prevents certain/all emails from sending (assuming all emails sent in production do so via this function, which should be the case).

packages/server/src/lib/email.js
```diff
@@ -533,7 +577,13 @@ function yesterday() {
 }
 
 async function buildAndSendGrantDigestEmails(userId, openDate = yesterday()) {
-    console.log(`Building and sending Grants Digest email for user: ${userId} on ${openDate}`);
+    if (!ENABLED_EMAIL_TYPES.includes(tags.emailTypes.grantDigest)) {
```
TylerHendrickson, line 580: This is an additional short-circuit check that happens earlier than the one above, added simply because this email tends to be more expensive to build and send as a batch digest to many recipients, so we avoid needlessly building emails.

sophaskins (approved):

